PR #10, opened 2019-11-18
Enable code deploy

dustinburson:


deploy/templates/default-azuredeploy-sandbox.json
```diff
@@ -12,7 +12,7 @@
     },
     "RepositoryUrl": {
       "type": "string",
-      "defaultValue": "https://github.com/microsoft/iomt-fhir",
+      "defaultValue": "https://github.com/Microsoft/iomt-fhir",
```
namalu: Is this really case sensitive?

LiquidPT: I wondered about this when I saw it in the code. Should have asked then. Yes, non-IIS servers (i.e. most) are case sensitive past the domain name

dustinburson: I don't believe so but made the change just to be sure.

PR #18, opened 2019-12-10
Personal/liquidpt/code analysis

LiquidPT:
Add FxCopAnalyzers back into projects. Added compiler package to allow the Analyzers to run

### DIFF
--- a/src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementFhirImportService.cs
+++ b/src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementFhirImportService.cs
@@ -24,7 +24,7 @@ namespace Microsoft.Health.Fhir.Ingest.Service
         private readonly FhirImportService _fhirImportService;
 
         public MeasurementFhirImportService(FhirImportService fhirImportService, MeasurementFhirImportOptions options)
-            : base(options, options.ParallelTaskOptions.MaxConcurrency)
+            : base(options, PassThroughNonNull(options).ParallelTaskOptions.MaxConcurrency)
         {
             _fhirImportService = EnsureArg.IsNotNull(fhirImportService, nameof(fhirImportService));
         }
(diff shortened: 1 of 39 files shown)


src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementFhirImportService.cs
```diff
@@ -24,7 +24,7 @@ public class MeasurementFhirImportService : ParallelTaskWorker<MeasurementFhirIm
         private readonly FhirImportService _fhirImportService;
 
         public MeasurementFhirImportService(FhirImportService fhirImportService, MeasurementFhirImportOptions options)
-            : base(options, options.ParallelTaskOptions.MaxConcurrency)
+            : base(options, PassThroughNonNull(options).ParallelTaskOptions.MaxConcurrency)
         {
```
dustinburson: Can probably replace with if null conditional ? and default to 1 if null.

test/Microsoft.Health.Fhir.Ingest.UnitTests/Microsoft.Health.Fhir.Ingest.UnitTests.csproj
```diff
@@ -8,11 +8,19 @@
...
+    <PackageReference Include="Microsoft.Net.Compilers" Version="2.9.0">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+    </PackageReference>
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.4.0" />
     <PackageReference Include="Newtonsoft.Json" Version="11.0.2" />
     <PackageReference Include="xunit" Version="2.4.0" />
     <PackageReference Include="xunit.runner.visualstudio" Version="2.4.0" />
     <PackageReference Include="NSubstitute" Version="4.2.0" />
+    <PackageReference Include="Microsoft.CodeAnalysis.FxCopAnalyzers" Version="2.9.0">
```
dustinburson: It looks like adding this back in caused build errors for default literals in the test project.

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementFhirImportService.cs
dustinburson: Also, Ensure.IsNotNull should return the value so you can just inline that with out needing your own function.

PR #19, opened 2020-01-03
Fixed issue where bundle extension could return null when resource exists

namalu:
A condition exists, where, a search request can return an empty bundle with a continuation token. The ReadOneFromBundle method assumed that if any Resource meeting the search criteria was found, the first bundle returned would contain at least one resource.

These changes include:

- A fix to ensure the continuation token is respected.
- Fixes to unit tests so they can be run on Linux and Mac machines.
- An update to the .gitignore file to ignore Mac OS system files.

### DIFF
--- a/src/lib/Microsoft.Health.Extensions.Fhir.R4/BundleExtensions.cs
+++ b/src/lib/Microsoft.Health.Extensions.Fhir.R4/BundleExtensions.cs
@@ -41,21 +41,28 @@ namespace Microsoft.Health.Extensions.Fhir
             }
         }
 
-        public static TResource ReadOneFromBundle<TResource>(this Bundle bundle, bool throwOnMultipleFound = true)
+        public static async Task<TResource> ReadOneFromBundleWithContinuationAsync<TResource>(this Bundle bundle, IFhirClient fhirClient, bool throwOnMultipleFound = true)
             where TResource : Resource, new()
         {
-            var bundleCount = bundle?.Entry?.Count ?? 0;
-            if (bundleCount == 0)
+            if (bundle == null)
             {
                 return null;
             }
 
-            if (throwOnMultipleFound && bundleCount > 1)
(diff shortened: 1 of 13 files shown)


test/Microsoft.Health.Extensions.Fhir.R4.UnitTests/BundleExtensionsTests.cs
```diff
@@ -0,0 +1,121 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
```
LiquidPT: Check order of using statements.

src/lib/Microsoft.Health.Extensions.Fhir.R4/BundleExtensions.cs
```diff
@@ -41,21 +41,28 @@ public static IEnumerable<TResource> ReadFromBundle<TResource>(this Bundle bundl
...
-            var bundleCount = bundle?.Entry?.Count ?? 0;
-            if (bundleCount == 0)
+            if (bundle == null)
+            {
+                return null;
+            }
+
+            var resources = await bundle?.ReadFromBundleWithContinuationAsync<TResource>(fhirClient, 2);
+
+            var resourceCount = resources.Count();
```
dustinburson: Minor, but best to use .ToArray to avoid multiple enumerations over the same collection (happens with Count & FirstOrDefault).  Impact is minimal since ReadFromBundleWithContinuation is using a materialized collection behind the scenes (list in this case) but if that ever changes we should re-evaluate.  

We also may want to look at using async enumerables going forward as well https://docs.microsoft.com/en-us/archive/msdn-magazine/2019/november/csharp-iterating-with-async-enumerables-in-csharp-8

PR #21, opened 2020-01-06
Fixed issue that caused token audience to be incorrect

namalu:
The Sandbox ARM template would set Audience in the authentication settings to "https://azurehealthcareapis.com" instead of "https://{SERVICE_NAME}.azurehealthcareapis.com". This causes tokens issued to the service client application to be denied access to the FHIR Server.

### DIFF
--- a/deploy/templates/default-azuredeploy-sandbox.json
+++ b/deploy/templates/default-azuredeploy-sandbox.json
@@ -217,8 +217,7 @@
     "iot_central_name": "[parameters('ServiceName')]",
     "fhir_service_name": "[parameters('ServiceName')]",
     "fhir_service_url": "[concat('https://', parameters('ServiceName'),'.azurehealthcareapis.com')]",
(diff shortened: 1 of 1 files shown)


deploy/templates/default-azuredeploy-sandbox.json
```diff
@@ -218,7 +218,7 @@
     "fhir_service_name": "[parameters('ServiceName')]",
     "fhir_service_url": "[concat('https://', parameters('ServiceName'),'.azurehealthcareapis.com')]",
     "fhir_service_cosmos_throughput": 1000,
-    "aad_fhir_server_audience": "https://azurehealthcareapis.com"
+    "aad_fhir_server_audience": "[concat('https://', parameters('ServiceName'),'.azurehealthcareapis.com')]"
```
LiquidPT: Could we just use variables('fhir_service_url')? This would make it explicit what we're doing and only one place to update if needed

PR #25, opened 2020-01-13
Updated NormalizeDeviceData Function to use Managed Identity

namalu:
Updated ARM Templates to leverage Managed Identity for NormalizeDeviceData Function.

### DIFF
--- a/deploy/templates/consumption-azuredeploy.json
+++ b/deploy/templates/consumption-azuredeploy.json
@@ -161,13 +161,17 @@
   "variables": {
     "asa_job_name": "[parameters('ServiceName')]",
     "eventhub_namespace_name": "[parameters('ServiceName')]",
+    "normalizeddata_eventhub_name": "[concat(variables('eventhub_namespace_name'), '/normalizeddata')]",
+    "devicedata_eventhub_name": "[concat(variables('eventhub_namespace_name'), '/devicedata')]",
     "storage_account_name": "[parameters('ServiceName')]",
     "app_plan_name": "[concat(parameters('ServiceName'), 'plan')]",
     "app_service_name": "[parameters('ServiceName')]",
     "app_insights_name": "[parameters('ServiceName')]",
     "key_vault_name": "[parameters('ServiceName')]",
     "app_service_resource_id": "[resourceId('Microsoft.Web/sites', variables('app_service_name'))]",
-    "deploy_source_code": "[and(not(empty(parameters('repositoryUrl'))),not(empty(parameters('repositoryBranch'))))]"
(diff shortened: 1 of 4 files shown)


deploy/templates/consumption-azuredeploy.json
```diff
@@ -351,74 +355,64 @@
...
-        ]
       }
     },
     {
-      "type": "Microsoft.EventHub/namespaces/eventhubs/consumergroups",
-      "apiVersion": "2017-04-01",
-      "name": "[concat(variables('eventhub_namespace_name'), '/devicedata/$Default')]",
-      "location": "[parameters('ResourceLocation')]",
+      "type": "Microsoft.EventHub/namespaces/eventhubs/providers/roleAssignments",
+      "apiVersion": "2018-01-01-preview",
```
namalu: Unfortunately, the best way to limit access to the Event Hub is to use a preview API. There is a way to avoid using a preview API, but the access control is less granular.

deploy/templates/consumption-azuredeploy.json
```diff
@@ -351,74 +355,64 @@
...
+      "name": "[concat(variables('devicedata_eventhub_name'), '/Microsoft.Authorization/', guid(uniqueString(variables('devicedata_eventhub_name'))))]",
```
dustinburson: Confused why we need a guid wrapped over unique string for the authorization rule?  Can't it just be a known name?

namalu: This is just why the documentation instructed - I didn’t try to use anything else once I got everything working.

https://docs.microsoft.com/en-us/azure/role-based-access-control/role-assignments-template#resource

src/func/Microsoft.Health.Fhir.Ingest.Host/Microsoft.Health.Fhir.Ingest.Host.csproj
```diff
@@ -13,6 +13,7 @@
     <WarningsAsErrors />
   </PropertyGroup>
   <ItemGroup>
+    <PackageReference Include="Microsoft.Azure.EventHubs.Processor" Version="4.1.0" />
```
LiquidPT: ><PackageReference Include="Microsoft.Azure.EventHubs.Processor" Version="4.1.0" /> [](start = 4, length = 82)

So you need this new package reference, but didn't actually change any code? Would have thought a code change would go along with this.

dustinburson: The change is in the connection string.  The new package allows the connection string using managed identity to function iirc.

PR #26, opened 2020-01-15
Personal/wiy/support observation category

wi-y:
added support for observation category

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -322,7 +322,7 @@ The CodeValueFhirTemplate is currently the only template supported in FHIR mappi
 | --- | ---
 |**TypeName**| The type of measurement this template should bind to. There should be at least one DeviceContent template that outputs this type.
 |**PeriodInterval**|The period of time the observation created should represent. Supported values are 0 (an instance), 60 (an hour), 1440 (a day).
-|**Category**|Any number of [Codings](http://hl7.org/fhir/datatypes-definitions.html#coding) to classify the type of observation created.
+|**Category**|Any number of [CodeableConcepts](http://hl7.org/fhir/datatypes-definitions.html#codeableconcept) to classify the type of observation created.
 |**Codes**|One or more [Codings](http://hl7.org/fhir/datatypes-definitions.html#coding) to apply to the observation created.
 |**Codes[].Code**|The code for the [Coding](http://hl7.org/fhir/datatypes-definitions.html#coding).
 |**Codes[].System**|The system for the [Coding](http://hl7.org/fhir/datatypes-definitions.html#coding).
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -322,6 +322,7 @@ The CodeValueFhirTemplate is currently the only template supported in FHIR mappi
 | --- | ---
 |**TypeName**| The type of measurement this template should bind to. There should be at least one DeviceContent template that outputs this type.
 |**PeriodInterval**|The period of time the observation created should represent. Supported values are 0 (an instance), 60 (an hour), 1440 (a day).
+|**Category**|Any number of [Codings](http://hl7.org/fhir/datatypes-definitions.html#coding) to classify the type of observation created.
 |**Codes**|One or more [Codings](http://hl7.org/fhir/datatypes-definitions.html#coding) to apply to the observation created.
 |**Codes[].Code**|The code for the [Coding](http://hl7.org/fhir/datatypes-definitions.html#coding).
 |**Codes[].System**|The system for the [Coding](http://hl7.org/fhir/datatypes-definitions.html#coding).
--- a/src/lib/Microsoft.Health.Fhir.Ingest/Template/CodeValueFhirTemplate.cs
+++ b/src/lib/Microsoft.Health.Fhir.Ingest/Template/CodeValueFhirTemplate.cs
@@ -10,6 +10,10 @@ namespace Microsoft.Health.Fhir.Ingest.Template
     public class CodeValueFhirTemplate : FhirTemplate
     {
 #pragma warning disable CA2227
+        public virtual IList<FhirCodeableConcept> Category { get; set; }
+#pragma warning restore CA2227
+
+#pragma warning disable CA2227
         public virtual IList<FhirCode> Codes { get; set; }
 #pragma warning restore CA2227
 
--- a/src/lib/Microsoft.Health.Fhir.R4.Ingest/Template/CodeValueFhirTemplateProcessor.cs
+++ b/src/lib/Microsoft.Health.Fhir.R4.Ingest/Template/CodeValueFhirTemplateProcessor.cs
@@ -159,7 +159,7 @@ namespace Microsoft.Health.Fhir.Ingest.Template
                 new CodeableConcept
                 {
                     Text = category.Text,
-                    Coding = category.Codes.Select(code => new Coding { System = code.System, Code = code.Code, Display = code.Display })
+                    Coding = category?.Codes?.Select(code => new Coding { System = code.System, Code = code.Code, Display = code.Display })
                     .ToList(),
                 }).ToList();
         }
--- a/src/lib/Microsoft.Health.Fhir.R4.Ingest/Template/CodeValueFhirTemplateProcessor.cs
+++ b/src/lib/Microsoft.Health.Fhir.R4.Ingest/Template/CodeValueFhirTemplateProcessor.cs
@@ -46,6 +46,11 @@ namespace Microsoft.Health.Fhir.Ingest.Template
                 },
             };
 
+            if (template?.Category?.Count > 0)
+            {
+                observation.Category = ResolveCategory(template.Category);
+            }
+
             var values = grp.GetValues();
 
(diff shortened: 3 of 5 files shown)


src/lib/Microsoft.Health.Fhir.Ingest/Template/CodeValueFhirTemplate.cs
```diff
@@ -9,6 +9,10 @@ namespace Microsoft.Health.Fhir.Ingest.Template
 {
     public class CodeValueFhirTemplate : FhirTemplate
     {
+#pragma warning disable CA2227
+        public virtual IList<FhirCodeableConcept> Category { get; set; }
```
LiquidPT: Should this be IList<FhirCode> like the "Codes" property below?

LiquidPT: Oh, I see, Codes is a singular CodeableConcept (list of codes), whereas Category is a list of CodeableConcepts: https://www.hl7.org/fhir/observation.html

src/lib/Microsoft.Health.Fhir.R4.Ingest/Template/CodeValueFhirTemplateProcessor.cs
```diff
@@ -81,6 +86,12 @@ protected override Observation MergeObservationImpl(CodeValueFhirTemplate templa
 
             existingObservation.Status = ObservationStatus.Amended;
 
+            existingObservation.Category = null;
```
LiquidPT: Do we want to eliminate the categories on the existing observation? Are we assuming that the only ones there are the ones from our template?

test/Microsoft.Health.Fhir.R4.Ingest.UnitTests/Template/CodeValueFhirTemplateProcessorTests.cs
```diff
@@ -542,5 +542,60 @@ public void GivenTemplateWithComponentAndObservationWithOutComponent_WhenMergObs
                         && v.end == boundary.end
                         && v.values.First().Item2 == "v2"));
         }
+
+        [Fact]
+        public void GivenTemplateWithCategory_ThenCategoryReturned()
```
LiquidPT: Maybe add another test where we merge observations with categories? Would help us solidify what we expect from that and ensure it's working as expected.

test/Microsoft.Health.Fhir.R4.Ingest.UnitTests/Template/CodeValueFhirTemplateProcessorTests.cs
```diff
@@ -542,5 +542,60 @@ public void GivenTemplateWithComponentAndObservationWithOutComponent_WhenMergObs
...
+            var observationGroup = Substitute.For<IObservationGroup>()
+                .Mock(m => m.Boundary.Returns(boundary))
+                .Mock(m => m.Name.Returns("code"))
+                .Mock(m => m.GetValues().Returns(values));
+
+            var observation = processor.CreateObservation(template, observationGroup);
+            Assert.Equal("a", observation.Category[0].Coding[0].Code);
+            Assert.Equal("b", observation.Category[0].Coding[0].Display);
+            Assert.Equal("c", observation.Category[0].Coding[0].System);
+            Assert.Equal("category text 1", observation.Category[0].Text);
```
LiquidPT: Should there not be a second coding (d;e;f) in here to test for?

src/lib/Microsoft.Health.Fhir.Ingest/Template/CodeValueFhirTemplate.cs
dustinburson: Yeap! Confusing at first but it makes sense.  It is something Will and I talked through today.  For codes it is singular codeable concept because it is still one concept but there could be different code systems (LOINC, SNOMED, etc).  Categories is different.  You could have two different concepts say something is a vital sign as well as a iomt.

src/lib/Microsoft.Health.Fhir.R4.Ingest/Template/CodeValueFhirTemplateProcessor.cs
wi-y: It's up for debate, but my preference is to overwrite. 

To me it seems strange to have one system sending in a category for an observation and then have another system send in another category for the same observation. I would assume that both systems would be consistent and send the same category for the same observation. 

I don't mind pulling out this particular line of code though. If no category is specified in the template then this line wipes out the category if it previously existed in the observation, which is a bit controversial. 

If we remove this line then we only overwrite if there is a category specified in the template. However, as a minor side effect, there would be no way to delete a category.

Thoughts?

test/Microsoft.Health.Fhir.R4.Ingest.UnitTests/Template/CodeValueFhirTemplateProcessorTests.cs
wi-y: Good catch. Will update the test

src/lib/Microsoft.Health.Fhir.R4.Ingest/Template/CodeValueFhirTemplateProcessor.cs
```diff
@@ -141,5 +152,16 @@ protected static CodeableConcept ResolveCode(string type, IList<FhirCode> codes)
...
             };
         }
+
+        protected static List<CodeableConcept> ResolveCategory(IList<FhirCodeableConcept> categories)
+        {
+            return categories.Select(category =>
+                new CodeableConcept
+                {
+                    Text = category.Text,
+                    Coding = category.Codes.Select(code => new Coding { System = code.System, Code = code.Code, Display = code.Display })
```
dustinburson: Will want to make sure a text only category but null codings works correctly as well.  I believe this code will error if the category codes are null.

src/lib/Microsoft.Health.Fhir.R4.Ingest/Template/CodeValueFhirTemplateProcessor.cs
dustinburson: I am fine keeping it.  Wiping out is fine since if that is the template change there won't be a category for observations going forward.

sordahl-ga: So just to make sure the only way this changes is template changes, so overwrite going forward is OK with me and we would have to assume some consistency due diligence on the template managers part.  Also if it runs through a profile on the way in that could also help with enforcement/consistency.

dustinburson: @sordahl-ga through the connector/tool that is correct.  Changing the template is the only way to make modifications.  If an organization allowed observations created by the connector to be edited by others there could be an issue but our position is the connector should own the observations it creates and they aren't mutable by other actors.

docs/Configuration.md
```diff
@@ -322,6 +322,7 @@ The CodeValueFhirTemplate is currently the only template supported in FHIR mappi
 | --- | ---
 |**TypeName**| The type of measurement this template should bind to. There should be at least one DeviceContent template that outputs this type.
 |**PeriodInterval**|The period of time the observation created should represent. Supported values are 0 (an instance), 60 (an hour), 1440 (a day).
+|**Category**|Any number of [Codings](http://hl7.org/fhir/datatypes-definitions.html#coding) to classify the type of observation created.
```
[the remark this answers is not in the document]
wi-y: Yeah, I agree

test/Microsoft.Health.Fhir.R4.Ingest.UnitTests/Template/CodeValueFhirTemplateProcessorTests.cs
```diff
@@ -542,5 +542,60 @@ public void GivenTemplateWithComponentAndObservationWithOutComponent_WhenMergObs
                         && v.end == boundary.end
                         && v.values.First().Item2 == "v2"));
         }
+
+        [Fact]
```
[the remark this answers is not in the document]
wi-y: Got it - updated the tests.

src/lib/Microsoft.Health.Fhir.R4.Ingest/Template/CodeValueFhirTemplateProcessor.cs
LiquidPT: @dustinburson I was thinking of a use case (and maybe this is an incorrect usage of category) where a system would analyze observations and flag them for some reason (possibly for further investigation, or it's out of the norm or some other reason). Our system would then wipe that out if another reading in that time window comes in.

dustinburson: @LiquidPT The scenario makes sense but I don't think that is an appropriate use of category according to the FHIR spec.  More appropriate would be using another observation that references the device one with an interpretation or a risk assessment.  If the underlying observation needs to be modified for the purpose you describe using the interpretation property (https://www.hl7.org/fhir/observation-definitions.html#Observation.interpretation) or an extension seems more appropriate.

PR #35, opened 2020-02-18
support docker containers for functions code

wi-y:
This PR may be controversial since it requires some package references to be updated.

"Microsoft.Net.Compilers" was changed to "Microsoft.Net.Compilers.Toolset" since only the Toolset can successfully compile for Linux. 

"Microsoft.Net.Compilers" also appeared to be deprecated, so we should probably migrate off of it anyway. https://github.com/dotnet/roslyn/issues/40129#issuecomment-565661236

The controversial part is that switching to Microsoft.Net.Compilers.Toolset seems to require Microsoft.NET.Sdk.Functions to be upgraded and TargetFramework to be updated to 3.1

### DIFF
--- a/DockerRun.ps1
+++ b/DockerRun.ps1
@@ -1,0 +1,39 @@
+<#
+    This script can be used to build and run the IoMT Azure Function code in a docker container.
+    It can be used for local development to build and run Docker images, and images can be used for cloud deployment.
+    By default console logging is turned on, but can be turned off by removing AzureFunctionsJobHost__Logging__Console__IsEnabled
+    One caveat with local development is that Docker cannot easily communicate with the Azure Storage Emulator, and it is easiest use a cloud storage container instead.
+#>
+
+if(-not (Test-Path .\src\func\Microsoft.Health.Fhir.Ingest.Host\local.settings.json)) {
+    Write-Output 'Could not locate local.settings.json'
+    Exit
+}
+
+$json = Get-Content .\src\func\Microsoft.Health.Fhir.Ingest.Host\local.settings.json | ConvertFrom-Json
+$env = $json.Values
+
+if($env.AzureWebJobsStorage -eq "UseDevelopmentStorage=true") {
+    Write-Output 'Setting "AzureWebJobsStorage":"UseDevelopmentStorage=true" in local.settings.json not supported with Docker, must use connection string'
+    Exit
+}
+
+docker build --tag iomt:v1.0.0 .
+docker run -p 8080:80 `
+    -e AzureFunctionsJobHost__Logging__Console__IsEnabled='true' `
+    -e AzureWebJobsScriptRoot='/home/site/wwwroot' `
+    -e AzureWebJobsStorage="$($env."AzureWebJobsStorage")" `
+    -e AzureWebJobsSecretStorageType="$($env."AzureWebJobsSecretStorageType")" `
+    -e FhirService:Authority="$($env."FhirService:Authority")" `
+    -e FhirService:ClientId="$($env."FhirService:ClientId")" `
+    -e FhirService:ClientSecret="$($env."FhirService:ClientSecret")" `
+    -e FhirService:Resource="$($env."FhirService:Resource")" `
+    -e FhirService:Url="$($env."FhirService:Url")" `
+    -e FUNCTIONS_WORKER_RUNTIME="$($env."FUNCTIONS_WORKER_RUNTIME")" `
+    -e InputEventHub="$($env."InputEventHub")" `
+    -e OutputEventHub="$($env."OutputEventHub")" `
+    -e ResourceIdentity:ResourceIdentityServiceType="$($env."ResourceIdentity:ResourceIdentityServiceType")" `
+    -e ResourceIdentity:DefaultDeviceIdentifierSystem="$($env."ResourceIdentity:DefaultDeviceIdentifierSystem")" `
+    -e Template:DeviceContent="$($env."Template:DeviceContent")" `
+    -e Template:FhirMapping="$($env."Template:FhirMapping")" `
+    -it iomt:v1.0.0
--- a/src/func/Microsoft.Health.Fhir.Ingest.Host/Microsoft.Health.Fhir.Ingest.Host.csproj
+++ b/src/func/Microsoft.Health.Fhir.Ingest.Host/Microsoft.Health.Fhir.Ingest.Host.csproj
@@ -13,7 +13,6 @@
     <WarningsAsErrors />
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.AspNetCore.App" />
     <PackageReference Include="Microsoft.Azure.EventHubs.Processor" Version="4.1.0" />
     <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.11" />
     <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.EventHubs" Version="3.0.5" />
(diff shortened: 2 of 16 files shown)


LiquidPT:
So, we were on .NET Core 3 originally, and that broke the deploy step of our ARM templates (it didn’t support 3.x). We downgraded back down to 2.x and some of the test pipelines broke and long with the deploy step (some of the references for the analyzers didn’t work). The fix, as per the analyzer team, was to explicitly reference the .NET core complier libraries that we wanted to use.

Unless Azure has fixed their source code deployment in the last couple of months, this will break our OSS ARM templates

src/func/Microsoft.Health.Fhir.Ingest.Host/Microsoft.Health.Fhir.Ingest.Host.csproj
```diff
@@ -13,19 +13,18 @@
     <WarningsAsErrors />
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.Azure.EventHubs.Processor" Version="4.1.0" />
```
dustinburson: Do we know why this was removed?  We need this version of EvenHubs.Processor or higher to ensure managed identity works.

src/func/Microsoft.Health.Fhir.Ingest.Host/Microsoft.Health.Fhir.Ingest.Host.csproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.2</TargetFramework>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
```
dustinburson: Test projects will need to be updated as well.  You will probably want to upgrade the function version to v3.

src/func/Microsoft.Health.Fhir.Ingest.Host/Microsoft.Health.Fhir.Ingest.Host.csproj
```diff
@@ -1,6 +1,6 @@
...
     <AzureFunctionsVersion>v2</AzureFunctionsVersion>
```
dustinburson: You will want to upgrade to v3 to match the latest Azure Functions SDK I believe.

DockerRun.ps1
```diff
@@ -0,0 +1,39 @@
+<#
+    This script can be used to build and run the IoMT Azure Function code in a docker container.
```
dustinburson: This is currently in the root?  We probably want this in a directory.  How about a docker dir under deploy?

dustinburson:
To verify we need to do a deployment of the original ARM template from the feature branch to make sure it works as expected.

> So, we were on .NET Core 3 originally, and that broke the deploy step of our ARM templates (it didn’t support 3.x). We downgraded back down to 2.x and some of the test pipelines broke and long with the deploy step (some of the references for the analyzers didn’t work). The fix, as per the analyzer team, was to explicitly reference the .NET core complier libraries that we wanted to use. Unless Azure has fixed their source code deployment in the last couple of months, this will break our OSS ARM templates.

src/func/Microsoft.Health.Fhir.Ingest.Host/Microsoft.Health.Fhir.Ingest.Host.csproj
namalu: We do need this for Managed Identity. It needs to be added back manually because we are declaring a newer version than is required by another dependency.

wi-y: I'll add it back since it is needed for managed identity. While running in docker, this package prevents the function from reading from the event hub. Will dig into it a bit more.

wi-y:
Looks like the .NET Core 3.1 SDK isn't yet supported for Windows app services at the moment, which will prevent our ARM template from being deployed if we specify version 3.

However there are some github comments saying that 3.1.2 should be supported by the end of Febuary 2020, so maybe we can give it a try at the end of the month.

Here's where this is being tracked:
https://github.com/Azure/app-service-announcements-discussions/issues/129
https://github.com/dotnet/core/blob/master/release-notes/3.1/3.1.2/3.1.2.md
https://aspnetcoreon.azurewebsites.net/#

dustinburson:
> Looks like the .NET Core 3.1 SDK isn't yet supported for Windows app services at the moment, which will prevent our ARM template from being deployed if we specify version 3.
> 
> However there are some github comments saying that 3.1.2 should be supported by the end of Febuary 2020, so maybe we can give it a try at the end of the month.
> 
> Here's where this is being tracked:
> [Azure/app-service-announcements-discussions#129](https://github.com/Azure/app-service-announcements-discussions/issues/129)
> https://github.com/dotnet/core/blob/master/release-notes/3.1/3.1.2/3.1.2.md
> https://aspnetcoreon.azurewebsites.net/#

Just a heads up, I believe this is only support for the runtime.  I.E. you can run .NET  Core 3.1 code.  I don't know if this means the project build will work but we can verify when it is available.

src/func/Microsoft.Health.Fhir.Ingest.Host/Microsoft.Health.Fhir.Ingest.Host.csproj
wi-y: Added it back. Root cause was having two functions running at the same time from the same event hub and apparently there is differing behavior between 4.1.0 and lower versions when that occurs.

wi-y:
In summary, we reverted back to the 2.2 SDK and then fixed a couple of dependency issues to fix the builds. Should be good now.

DockerRun.ps1
dustinburson: Thank you for moving the script but can we move the docker file itself to the folder as well?  Thanks!  Otherwise changes look great.

wi-y:
> If we can move the docker file to the same place as the script I think we are good to finish this!

Unfortunately it doesn't seem possible with Docker, since Docker does not allow you to add files outside of it's "context" for security reasons. For example if we put the Dockerfile in ./deploy/docker/ then Docker sets the context as that directory and can no longer access ../../src/*

dustinburson:
> > If we can move the docker file to the same place as the script I think we are good to finish this!
> 
> Unfortunately it doesn't seem possible with Docker, since Docker does not allow you to add files outside of it's "context" for security reasons. For example if we put the Dockerfile in ./deploy/docker/ then Docker sets the context as that directory and can no longer access ../../src/*

Would it work if we placed it in iomt-fhir/src/func/Microsoft.Health.Fhir.Ingest.Host? Trying to avoid having containers in the root and everything needed to run the function app should be in that directory.  It would also align with how containers are packaged in our other repos like https://github.com/microsoft/fhir-server-samples/tree/master/src/FhirImporter

wi-y:
> If we can move the docker file to the same place as the script I think we are good to finish this!
> 
> Unfortunately it doesn't seem possible with Docker, since Docker does not allow you to add files outside of it's "context" for security reasons. For example if we put the Dockerfile in ./deploy/docker/ then Docker sets the context as that directory and can no longer access ../../src/*
> 
> Would it work if we placed it in iomt-fhir/src/func/Microsoft.Health.Fhir.Ingest.Host? Trying to avoid having containers in the root and everything needed to run the function app should be in that directory.  It would also align with how containers are packaged in our other repos like https://github.com/microsoft/fhir-server-samples/tree/master/src/FhirImporter

Unfortunately no. For it to work for Docker we would need to move the libs into that directory as well, but moving the libs also would break the project since multiple .csproj would be found when building (at least that was my experience).

dustinburson:
> moving the libs also would break the project since multiple .csproj would be found when building (at least that was my experience).

We can sync offline but I am confused why the lib projects themselves are needed at that point.  If you build the function all the dlls needed to run the project should be bin placed with in the func app.  I trust you this I just want to understand where my understanding is breaking down :)

PR #39, opened 2020-02-27
Add document for deploying managed identity

LiquidPT:
Adding documentation for installing the connector using managed identity to connect to Azure API for FHIR, including the "easy button"

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,11 +25,11 @@ This open source project is fully backed by the Microsoft Healthcare team, but w
 
 # Getting Started
 To get started, there are a few options:
-1. Deploy the [IoMT FHIR Connector for Azure](./docs/ARMInstallationManagedIdentity.md) by itself for use with Azure API for FHIR and Azure Active directory in the same subscription
-2. Deploy the [IoMT FHIR Connector for Azure](./docs/ARMInstallation.md) by itself for use with another FHIR server or identity provider
+1. Deploy the [IoMT FHIR Connector for Azure](./docs/ARMInstallationManagedIdentity.md) by itself for use with Azure API for FHIR and Azure Active directory in the same subscription. Requests made to the Azure API for FHIR will be authenticated using the Azure Active Directory within this subscription - External identity providers cannot be used if the IoMT FHIR Connector for Azure is deployed with this template.
+2. Deploy the [IoMT FHIR Connector for Azure](./docs/ARMInstallation.md) by itself for use with another FHIR server or identity provider.
 3. Start with a complete [sandbox environment](./docs/Sandbox.md) that includes an instance of [IoT Central](https://azure.microsoft.com/en-us/services/iot-central/) with simulated devices and a deployed instance of the [Azure API for FHIR](https://docs.microsoft.com/en-us/azure/healthcare-apis/).
 
-To send messages to the connector you can [send events](https://docs.microsoft.com/en-us/azure/event-hubs/event-hubs-dotnet-standard-getstarted-send) directly to the `devicedata` EventHub deployed or [send events](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-sdks) to one of the Azure IoT solutions and [export messages](./docs/Iot.md) to the connector. 
+To send messages to the connector you can [send events](https://docs.microsoft.com/en-us/azure/event-hubs/event-hubs-dotnet-standard-getstarted-send) directly to the `devicedata` EventHub deployed as part of IoMT FHIR Connector for Azure or [send events](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-sdks) to one of the Azure IoT solutions and [export messages](./docs/Iot.md) to the connector. 
 
 # Architecture
 
@@ -48,7 +48,8 @@ To send messages to the connector you can [send events](https://docs.microsoft.c
 
 # Documentation
 - [Configuration](./docs/Configuration.md): Documents the different configurations required for the connector.
-- [Open Source Deployment](./docs/ARMInstallation.md): Describes how to deploy the IoMT FHIR Connector for Azure.
+- [Open Source Deployment using Managed Identity](./docs/ARMInstallationManagedIdentity.md): Describes how to deploy the IoMT FHIR Connector for Azure using Azure API for FHIR and Azure Active directory in the same subscription.
+- [Open Source Deployment - Flexible](./docs/ARMInstallation.md): Describes how to deploy the IoMT FHIR Connector for Azure using another FHIR server or identity provider.
 - [Sandbox Deployment](./docs/Sandbox.md): Describes how to deploy an end to end sandbox environment using IoT Central, IoMT FHIR Connector for Azure, and the Azure API for FHIR.
 - [Connecting to Azure IoT](./docs/Iot.md): Describes how to connect the IoMT FHIR Connector for Azure with different Azure IoT solutions like IoT Hub and IoT Central.
 - [Debugging](./docs/Debugging.md): Documents steps for local and cloud debugging.
--- a/README.md
+++ b/README.md
@@ -24,7 +24,10 @@ This open source project is fully backed by the Microsoft Healthcare team, but w
 - A FHIR mapping template uploaded to the template storage container. See [Configuration](./docs/Configuration.md) for more information.
 
 # Getting Started
-To get started you can deploy the [IoMT FHIR Connector for Azure](./docs/ARMInstallation.md) by itself or start with a complete [sandbox environment](./docs/Sandbox.md) that includes an instance of [IoT Central](https://azure.microsoft.com/en-us/services/iot-central/) with simulated devices and a deployed instance of the [Azure API for FHIR](https://docs.microsoft.com/en-us/azure/healthcare-apis/).
+To get started, there are a few options:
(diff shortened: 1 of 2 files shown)


docs/ARMInstallationManagedIdentity.md
```diff
@@ -0,0 +1,59 @@
...
+
+|Type|Behavior
+|---|---
+|**R4DeviceAndPatientLookupIdentityService**|Default setting.  Device identifier from ingested messages is retrieved from the FHIR server. Patient is expected to be linked to the device.
+|**R4DeviceAndPatientCreateIdentityService**|System attempts to retrieve the device identifier and associated patient from the FHIR server. If either isn't found a shell resource with just the identity will be created. Requires a patient identifier be mapped in the device content configuration template.
+|**R4DeviceAndPatientWithEncounterLookupIdentityService**|Like the first setting but allows you to include an encounter identifier with the message to associate with the device/patient.  The encounter is looked up during processing and any observations created are linked to the encounter. The association here is assumed to be one encounter per device.
+
+## Post Deployment
+After the ARM template is successfully deployed, add the Managed Identity ID output by the ARM deployment to the Allowed Object IDs on the Authentication page of your Azure API for FHIR.
+
```
namalu: I know it's redundant, but I feel just a quick blurb here about using the Azure Active Directory as the identity provider for the Azure API for FHIR will reduce the chance of confusion. In fact we should point our that changing the Authority on the Authentication page of your Azure API for FHIR will cause the connector to error with 401s.

README.md
```diff
@@ -24,7 +24,10 @@ This open source project is fully backed by the Microsoft Healthcare team, but w
 - A FHIR mapping template uploaded to the template storage container. See [Configuration](./docs/Configuration.md) for more information.
 
 # Getting Started
-To get started you can deploy the [IoMT FHIR Connector for Azure](./docs/ARMInstallation.md) by itself or start with a complete [sandbox environment](./docs/Sandbox.md) that includes an instance of [IoT Central](https://azure.microsoft.com/en-us/services/iot-central/) with simulated devices and a deployed instance of the [Azure API for FHIR](https://docs.microsoft.com/en-us/azure/healthcare-apis/).
+To get started, there are a few options:
+1. Deploy the [IoMT FHIR Connector for Azure](./docs/ARMInstallationManagedIdentity.md) by itself for use with Azure API for FHIR and Azure Active directory in the same subscription
```
ms-puneet-nagpal: Does it mean managed identity flavor of IoMT connector doesn't work if Azure API for FHIR and IoMT Connector are on different subscriptions? If yes, we may want to cover this use case in #2 line item.

README.md
```diff
@@ -24,7 +24,10 @@ This open source project is fully backed by the Microsoft Healthcare team, but w
...
+To get started, there are a few options:
+1. Deploy the [IoMT FHIR Connector for Azure](./docs/ARMInstallationManagedIdentity.md) by itself for use with Azure API for FHIR and Azure Active directory in the same subscription
+2. Deploy the [IoMT FHIR Connector for Azure](./docs/ARMInstallation.md) by itself for use with another FHIR server or identity provider
+3. Start with a complete [sandbox environment](./docs/Sandbox.md) that includes an instance of [IoT Central](https://azure.microsoft.com/en-us/services/iot-central/) with simulated devices and a deployed instance of the [Azure API for FHIR](https://docs.microsoft.com/en-us/azure/healthcare-apis/).
 
 To send messages to the connector you can [send events](https://docs.microsoft.com/en-us/azure/event-hubs/event-hubs-dotnet-standard-getstarted-send) directly to the `devicedata` EventHub deployed or [send events](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-sdks) to one of the Azure IoT solutions and [export messages](./docs/Iot.md) to the connector.
```
ms-puneet-nagpal: I remember getting confused by "device data EventHub deployed". Could we extend it further like "device data EvenHub deployed as part of IoMT FHIR Connector for Azure"?

ms-puneet-nagpal: It would be good to add Managed Identity deployment option under Documentation section below. We may also need some renaming for existing "Open Source Deployment" option.

docs/ARMInstallationManagedIdentity.md
```diff
@@ -0,0 +1,59 @@
...
+* App Insights - Used to record telemetry.
+* Managed Identity - an Azure Active Directory service identity for the IoMT FHIR Connector for Azure to use to connect to the Azure API for FHIR
+
+### Prerequisites
+To run this ARM template the following additional items must be set up before execution:
+
+* FHIR Server - An Azure API for FHIR instance using FHIR version R4
+
+### Parameters
+The following parameters are provided by the ARM template:
```
ms-puneet-nagpal: There are two sections of parameters in the template: BASICS and SETTINGS. Below parameters are covering all parameters for SETTINGS section, but parameters for BASICS section are missing. Won't we need to document those?

docs/ARMInstallationManagedIdentity.md
```diff
@@ -0,0 +1,59 @@
...
+
+|Type|Behavior
+|---|---
+|**R4DeviceAndPatientLookupIdentityService**|Default setting.  Device identifier from ingested messages is retrieved from the FHIR server. Patient is expected to be linked to the device.
+|**R4DeviceAndPatientCreateIdentityService**|System attempts to retrieve the device identifier and associated patient from the FHIR server. If either isn't found a shell resource with just the identity will be created. Requires a patient identifier be mapped in the device content configuration template.
+|**R4DeviceAndPatientWithEncounterLookupIdentityService**|Like the first setting but allows you to include an encounter identifier with the message to associate with the device/patient.  The encounter is looked up during processing and any observations created are linked to the encounter. The association here is assumed to be one encounter per device.
+
+## Post Deployment
+After the ARM template is successfully deployed, add the Managed Identity ID output by the ARM deployment to the Allowed Object IDs on the Authentication page of your Azure API for FHIR.
```
ms-puneet-nagpal: It might be helpful to specify how ARM deployment output can be obtained from the Azure portal.

README.md
```diff
@@ -24,7 +24,10 @@ This open source project is fully backed by the Microsoft Healthcare team, but w
...
 
 # Getting Started
-To get started you can deploy the [IoMT FHIR Connector for Azure](./docs/ARMInstallation.md) by itself or start with a complete [sandbox environment](./docs/Sandbox.md) that includes an instance of [IoT Central](https://azure.microsoft.com/en-us/services/iot-central/) with simulated devices and a deployed instance of the [Azure API for FHIR](https://docs.microsoft.com/en-us/azure/healthcare-apis/).
+To get started, there are a few options:
+1. Deploy the [IoMT FHIR Connector for Azure](./docs/ARMInstallationManagedIdentity.md) by itself for use with Azure API for FHIR and Azure Active directory in the same subscription
+2. Deploy the [IoMT FHIR Connector for Azure](./docs/ARMInstallation.md) by itself for use with another FHIR server or identity provider
+3. Start with a complete [sandbox environment](./docs/Sandbox.md) that includes an instance of [IoT Central](https://azure.microsoft.com/en-us/services/iot-central/) with simulated devices and a deployed instance of the [Azure API for FHIR](https://docs.microsoft.com/en-us/azure/healthcare-apis/).
```
dustinburson: Nit, the list items aren't consistent with usage of period at the end of the item.

docs/ARMInstallationManagedIdentity.md
```diff
@@ -0,0 +1,59 @@
...
+
+### Prerequisites
+To run this ARM template the following additional items must be set up before execution:
+
+* FHIR Server - An Azure API for FHIR instance using FHIR version R4
+
+### Parameters
+The following parameters are provided by the ARM template:
+
+|Parameter|Use
```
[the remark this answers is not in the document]
LiquidPT: I understand that concern. My thought process was to have them choose which installation they were using early, then have complete documentation for that. Otherwise I think we'll have a document with a number of "if... then..." areas in it, and 2 "easy buttons", which might make hitting the wrong one easy. Essentially, I thought 1 big decision up front would be better than a bunch of micro decisions for the user. I agree it's harder to maintain. Thoughts?

README.md
LiquidPT: I believe so, but @namalu or @dustinburson can confirm

docs/ARMInstallationManagedIdentity.md
LiquidPT: The BASICS aren't part of our template or under our control. Those are things that are added by Azure Portal as part of the custom template deployment screen.

docs/ARMInstallationManagedIdentity.md
LiquidPT: It's on the Outputs page of the deployment. Do I need to add that? 

I think we're trying to stay away from being TOO step-by-step since 1) Azure portal can change and 2) there are multiple ways to deploy ARM templates, not just through the portal.

README.md
dustinburson: I think it is more correct to say it won't work if they are on different tenants but in principle yes.

docs/ARMInstallationManagedIdentity.md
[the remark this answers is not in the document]
dustinburson: We can keep the separate page for now.  I thinking of one page with scenario 1 then it would detail scenario 2 and just highlight the differences.  It will probably take some iterations to get it right though so lets proceed like this and consider this something for the backlog,

PR #40, opened 2020-02-28
Include TemplateNotFound in the handled exceptions

dustinburson:
When a FhirMappingTemplate isn't found it currently causes an internal server error and prevents the stream analytics job from proceeding.  Change behavior to capture exception and report as telemetry.


test/Microsoft.Health.Fhir.Ingest.UnitTests/Telemetry/ExceptionTelemetryProcessorTests.cs
```diff
@@ -20,6 +21,7 @@ public class ExceptionTelemetryProcessorTests
         [InlineData(typeof(NotSupportedException))]
         [InlineData(typeof(FhirResourceNotFoundException))]
         [InlineData(typeof(ResourceIdentityNotDefinedException))]
+        [InlineData(typeof(TemplateNotFoundException))]
```
namalu: Should we change this to MappingFileNotFound?

dustinburson: We have a lot of refactoring to do on that front :)

PR #41, opened 2020-03-23
add methods for health checks

wi-y:
- added methods for health checks
- a developer can add CheckHealth() to Functions.cs wrapped in a try/catch to check health of fhir server

### DIFF
--- a/src/lib/Microsoft.Health.Fhir.Ingest/FhirHealthCheckStatus.cs
+++ b/src/lib/Microsoft.Health.Fhir.Ingest/FhirHealthCheckStatus.cs
@@ -1,0 +1,20 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Fhir.Ingest
+{
+    public class FhirHealthCheckStatus
+    {
+        public FhirHealthCheckStatus(string message, int statusCode)
+        {
+            Message = message;
+            StatusCode = statusCode;
+        }
+
+        public string Message { get; }
+
+        public int StatusCode { get; }
+    }
+}
--- a/src/lib/Microsoft.Health.Fhir.Ingest/Service/FhirHealthCheckService.cs
+++ b/src/lib/Microsoft.Health.Fhir.Ingest/Service/FhirHealthCheckService.cs
@@ -1,0 +1,25 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System.Threading.Tasks;
+using EnsureThat;
+
+namespace Microsoft.Health.Fhir.Ingest.Service
+{
+    public class FhirHealthCheckService
+    {
+        private readonly FhirHealthService _fhirHealthService;
+
+        public FhirHealthCheckService(FhirHealthService fhirHealthService)
+        {
+            _fhirHealthService = EnsureArg.IsNotNull(fhirHealthService, nameof(fhirHealthService));
+        }
+
+        public async Task<FhirHealthCheckStatus> CheckHealth()
+        {
+            return await _fhirHealthService.CheckHealth();
+        }
+    }
+}
--- a/src/lib/Microsoft.Health.Fhir.Ingest/Service/FhirHealthCheckService.cs
+++ b/src/lib/Microsoft.Health.Fhir.Ingest/Service/FhirHealthCheckService.cs
@@ -3,8 +3,10 @@
 // Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // -------------------------------------------------------------------------------------------------
 
+using System.Threading;
 using System.Threading.Tasks;
 using EnsureThat;
+using Microsoft.Health.Fhir.Ingest.Data;
 
 namespace Microsoft.Health.Fhir.Ingest.Service
 {
@@ -17,9 +19,9 @@ namespace Microsoft.Health.Fhir.Ingest.Service
             _fhirHealthService = EnsureArg.IsNotNull(fhirHealthService, nameof(fhirHealthService));
         }
 
-        public async Task<FhirHealthCheckStatus> CheckHealth()
+        public async Task<FhirHealthCheckStatus> CheckHealth(CancellationToken token = default)
         {
-            return await _fhirHealthService.CheckHealth();
+            return await _fhirHealthService.CheckHealth(token);
         }
     }
 }
--- a/src/lib/Microsoft.Health.Fhir.Ingest/Service/FhirHealthService.cs
+++ b/src/lib/Microsoft.Health.Fhir.Ingest/Service/FhirHealthService.cs
@@ -1,0 +1,14 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System.Threading.Tasks;
+
+namespace Microsoft.Health.Fhir.Ingest.Service
+{
+    public abstract class FhirHealthService
+    {
+        public abstract Task<FhirHealthCheckStatus> CheckHealth();
+    }
+}
--- a/src/lib/Microsoft.Health.Fhir.Ingest/Service/FhirHealthService.cs
+++ b/src/lib/Microsoft.Health.Fhir.Ingest/Service/FhirHealthService.cs
@@ -3,12 +3,14 @@
 // Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // -------------------------------------------------------------------------------------------------
 
+using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.Health.Fhir.Ingest.Data;
 
 namespace Microsoft.Health.Fhir.Ingest.Service
 {
     public abstract class FhirHealthService
     {
-        public abstract Task<FhirHealthCheckStatus> CheckHealth();
+        public abstract Task<FhirHealthCheckStatus> CheckHealth(CancellationToken token = default);
     }
 }
--- a/src/lib/Microsoft.Health.Fhir.Ingest/Service/FhirImportService.cs
+++ b/src/lib/Microsoft.Health.Fhir.Ingest/Service/FhirImportService.cs
@@ -31,5 +31,7 @@ namespace Microsoft.Health.Fhir.Ingest.Service
         }
 
         public abstract Task ProcessAsync(ILookupTemplate<IFhirTemplate> config, IMeasurementGroup data, Func<Exception, IMeasurementGroup, Task<bool>> errorConsumer = null);
+
+        public abstract Task<string> CheckHealth();
     }
 }
--- a/src/lib/Microsoft.Health.Fhir.Ingest/Service/FhirImportService.cs
+++ b/src/lib/Microsoft.Health.Fhir.Ingest/Service/FhirImportService.cs
@@ -31,7 +31,5 @@ namespace Microsoft.Health.Fhir.Ingest.Service
         }
 
         public abstract Task ProcessAsync(ILookupTemplate<IFhirTemplate> config, IMeasurementGroup data, Func<Exception, IMeasurementGroup, Task<bool>> errorConsumer = null);
-
-        public abstract Task<string> CheckHealth();
     }
 }
--- a/src/lib/Microsoft.Health.Fhir.R4.Ingest/Host/MeasurementFhirImportExtensions.cs
+++ b/src/lib/Microsoft.Health.Fhir.R4.Ingest/Host/MeasurementFhirImportExtensions.cs
@@ -10,6 +10,7 @@ using Microsoft.Azure.WebJobs;
 using Microsoft.Extensions.Caching.Memory;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection.Extensions;
 using Microsoft.Extensions.Options;
 using Microsoft.Health.Common;
 using Microsoft.Health.Extensions.Fhir;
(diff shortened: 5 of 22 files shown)


src/lib/Microsoft.Health.Fhir.Ingest/Service/FhirImportService.cs
```diff
@@ -31,5 +31,7 @@ protected static (string Identifer, string System) GenerateObservationId(IObserv
         }
 
         public abstract Task ProcessAsync(ILookupTemplate<IFhirTemplate> config, IMeasurementGroup data, Func<Exception, IMeasurementGroup, Task<bool>> errorConsumer = null);
+
+        public abstract Task<string> CheckHealth();
```
dustinburson: Why part of the FhirImportService?  Would ideally like this to be it's own service

src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4FhirImportService.cs
```diff
@@ -98,6 +98,23 @@ public virtual async Task<string> SaveObservationAsync(ILookupTemplate<IFhirTemp
...
+        public override Task<string> CheckHealth()
+        {
+            SearchParams q = new SearchParams()
+                .ForSubject<Model.Patient>("IoMTHealthCheck")
+                .SetCount(1);
+
+            var result = _client.Search(q);
+            if (result != null && result.GetType() == typeof(Model.Bundle))
+            {
+                return Task.FromResult("Healthy");
```
dustinburson: We should record the result as a metric and include the timestamp as a dimension.  It would be even better to record the http status code as well so we can differentiate between unauthorized access and the service is not available.

src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4FhirImportService.cs
```diff
@@ -98,6 +98,23 @@ public virtual async Task<string> SaveObservationAsync(ILookupTemplate<IFhirTemp
             return result.Id;
         }
 
+        public override Task<string> CheckHealth()
+        {
+            SearchParams q = new SearchParams()
+                .ForSubject<Model.Patient>("IoMTHealthCheck")
```
dustinburson: Would be better to do a get against a foundational resource that won't have any PII like StructureDefinition or SearchParameter.  I understand the desire is to have no search result hence an id that shouldn't exist but a call against a non-PII resource would be preferred.  A get or search against a structural component will have less overhead on the underlying service too.  This will still execute a complete search against the patients.

dustinburson: There is a /health/check endpoint you can hit for the FHIR server as well but it doesn't require authentication.  We could consider adding a special one that behaves the same way but requires auth.  This wouldn't be in the open source though since it isn't a FHIR standard endpoint, rather an implementation detail on our end.

src/lib/Microsoft.Health.Fhir.Ingest/Service/FhirImportService.cs
dustinburson: How will this get called?  Ideally it would be another service that would be wrapped in a timer based function.  https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-timer?tabs=csharp

wi-y: Yes the plan would be to wrap this in a timer based function from a non OSS repo. The reason why I put it in FhirImportService is that it would be nice to just pass in MeasurementFhirImportService measurementImportService to our timed based Health Check function and perform a health check that would mimic the same behavior/config as the real MeasurementCollectionToFhir function. 

I had originally started writing it as a separate service but found it involved copying nearly identical code in many places. I don't have a preference though.

src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4FhirImportService.cs
wi-y: Yeah, we should. I was thinking of doing the logging as part of a try/catch in Functions.cs in a non OSS repo though. I didn't do the logging here since the metrics logging probably isn't needed for OSS.

The http status code is revealed in the exception when _client.search fails, so if we wrap CheckMetrics in a try/catch in Functions.cs then we should have access to the status code (It won't be discrete but it will be in the exception text).

src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4FhirImportService.cs
wi-y: That makes sense, I can change it to something without PII, for now.

namalu: It would be fairly easy to add a controller to the FhirApplication that could do an "authentication health check". Might not be a bad idea so we can catch issues early on if they occur.

src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4FhirImportService.cs
namalu: I agree with Dustin. If the result is binary, we are putting the DRI in a bit of a pickle - All they see is that something is broken.

src/lib/Microsoft.Health.Fhir.Ingest/Service/FhirImportService.cs
dustinburson: I think a separate service would prefer since the intended function is different.  Having a base IFhirService that takes the client that both inherit from makes sense though to reduce duplication.

src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4FhirImportService.cs
dustinburson: Regarding the merits of recording the metrics in OSS, if it is done around the ILogger interface we should be in a good position to inject additional logging channels when appropriate.

I am also fine with the recording being further up the stack but the result from the function needs to return the detail necessary still.

src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4FhirImportService.cs
dustinburson: I think the decision we need to make here is if the health check should be something that works against any FHIR server or if we want something specific when connecting to the Azure API for FHIR.

src/lib/Microsoft.Health.Fhir.Ingest/FhirHealthCheckStatus.cs
```diff
@@ -0,0 +1,20 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Fhir.Ingest
```
dustinburson, line 6: If this is being used to record the outcome the health check it probably makes sense to be in the Data namespace & folder.

src/lib/Microsoft.Health.Fhir.Ingest/Service/FhirHealthService.cs
```diff
@@ -0,0 +1,14 @@
...
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System.Threading.Tasks;
+
+namespace Microsoft.Health.Fhir.Ingest.Service
+{
+    public abstract class FhirHealthService
+    {
+        public abstract Task<FhirHealthCheckStatus> CheckHealth();
```
dustinburson, line 12: This should probably be updated to take a cancellation token since this will be asynchronous.  You can set it to default if you would like but it should be an option.

src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4FhirHealthService.cs
```diff
@@ -0,0 +1,45 @@
...
+        FhirHealthService
+    {
+        private readonly IFhirClient _client;
+
+        public R4FhirHealthService(IFhirClient fhirClient)
+        {
+            _client = EnsureArg.IsNotNull(fhirClient, nameof(fhirClient));
+        }
+
+        public override Task<FhirHealthCheckStatus> CheckHealth()
```
dustinburson, line 23: should be async

src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4FhirHealthService.cs
```diff
@@ -0,0 +1,45 @@
...
+        {
+            try
+            {
+                Hl7.Fhir.Model.Bundle result = _client.Search<Hl7.Fhir.Model.StructureDefinition>();
```
dustinburson, line 27: is there a SearchAsync we can use?  Alternately you can just use a /get since you aren't specifying any search parameters.  I would either switch to a get for efficiency or add a _count = 1 on the StructureDefinition search.

src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4FhirHealthService.cs
```diff
@@ -0,0 +1,45 @@
...
+                return Task.FromResult(new FhirHealthCheckStatus(ex.Message, (int)ex.Status));
+            }
+            catch (IdentityModel.Clients.ActiveDirectory.AdalServiceException ex)
+            {
+                return Task.FromResult(new FhirHealthCheckStatus(ex.Message, ex.StatusCode));
+            }
+            catch (Exception ex)
+            {
+                return Task.FromResult(new FhirHealthCheckStatus(ex.Message, 500));
+                throw;
```
dustinburson, line 41: Do we want to throw here?  I would think we want to record the issue and be done.  Was there a specific case you wanted to throw for?

src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4FhirHealthService.cs
```diff
@@ -0,0 +1,45 @@
...
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using System.Threading.Tasks;
+using EnsureThat;
+using Hl7.Fhir.Rest;
+
+namespace Microsoft.Health.Fhir.Ingest.Service
+{
+    public class R4FhirHealthService :
```
dustinburson, line 13: Can we have unit tests with mocks of the exceptions to test the various outputs?

src/lib/Microsoft.Health.Fhir.Ingest/Service/FhirHealthService.cs
```diff
@@ -0,0 +1,14 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System.Threading.Tasks;
+
+namespace Microsoft.Health.Fhir.Ingest.Service
+{
+    public abstract class FhirHealthService
```
dustinburson, line 10: Thoughts on the public method just returning Task and taking an ILogger for telemetry recording?  In this instance you would still have version of CheckHealth returning the FhirHealthCheckStatus but it would be a protected abstract member that the implementer needs to specify.  Then the public method of Task Name(ILogger, CancelationToken) would record the result in a standard way in the logger so the upstream implementer doesn't need to.  Only thing you need in the function definition is the instance of the Health Check & the ILogger which you can get with DI on the function signature.  This also makes the use of the abstract class more appropriate.  Right now the abstract class  doesn't have any implementation outside a contract and could just be an interface.

src/lib/Microsoft.Health.Fhir.R4.Ingest/Host/MeasurementFhirImportExtensions.cs
```diff
@@ -36,18 +37,18 @@ public static IWebJobsBuilder AddMeasurementFhirImport(this IWebJobsBuilder buil
...
+            builder.Services.TryAddSingleton<IResourceIdentityService>(
                 sp =>
                 {
                     var fhirClient = sp.GetRequiredService<IFhirClient>();
                     var resourceIdentityOptions = sp.GetRequiredService<IOptions<ResourceIdentityOptions>>();
                     return ResourceIdentityServiceFactory.Instance.Create(resourceIdentityOptions.Value, fhirClient);
                 });
-            builder.Services.AddSingleton<IMemoryCache>(sp => new MemoryCache(Options.Create<MemoryCacheOptions>(new MemoryCacheOptions { SizeLimit = 5000 })));
-            builder.Services.AddSingleton<FhirImportService, R4FhirImportService>();
+            builder.Services.TryAddSingleton<IMemoryCache>(sp => new MemoryCache(Options.Create<MemoryCacheOptions>(new MemoryCacheOptions { SizeLimit = 5000 })));
```
dustinburson, line 50: Looks good, thanks for making this change!

src/lib/Microsoft.Health.Fhir.Ingest/FhirHealthCheckStatus.cs
wi-y: Makes sense. Moved

src/lib/Microsoft.Health.Fhir.Ingest/Service/FhirHealthService.cs
wi-y: Updated

src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4FhirHealthService.cs
wi-y: Went with SearchAsync

src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4FhirHealthService.cs
wi-y: Updated

src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4FhirHealthService.cs
wi-y: Not really, I only added it originally because of the linter. I went ahead and removed the throw

src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4FhirHealthService.cs
wi-y: Added tests

src/lib/Microsoft.Health.Fhir.Ingest/Service/FhirHealthService.cs
wi-y: That makes sense, but I'm not sure I would like to address this at this moment

dustinburson: Okay, it can be part of the next step.

src/lib/Microsoft.Health.Fhir.Ingest/Service/FhirHealthCheckService.cs
```diff
@@ -0,0 +1,27 @@
...
+    public class FhirHealthCheckService
+    {
+        private readonly FhirHealthService _fhirHealthService;
+
+        public FhirHealthCheckService(FhirHealthService fhirHealthService)
+        {
+            _fhirHealthService = EnsureArg.IsNotNull(fhirHealthService, nameof(fhirHealthService));
+        }
+
+        public async Task<FhirHealthCheckStatus> CheckHealth(CancellationToken token = default)
```
dustinburson, line 22: standard convention is to have an Async suffix on Async methods.

PR #47, opened 2020-06-11
Add correlation id as option for grouping measurements

dustinburson:
Add an optional correlation id for grouping measurements to create observations in FHIR.
When a template is set to the new ObservationPeriodInterval = CorrelationId all measurements for a given device, type, and correlation id will share the same observation in FHIR.  This overrides the traditional time based grouping like hourly or daily.

In order to use the feature the device messages need to supply a correlation id that is then extracted with the via the CorrelationIdExpression property added to the Json & Iot templates.


test/Microsoft.Health.Fhir.Ingest.UnitTests/Data/CorrelationMeasurementObservationGroupFactoryTests.cs
```diff
@@ -0,0 +1,102 @@
+﻿using System;
```
LiquidPT, line 1: Build failing because of missing file header.

dustinburson: Thanks, this is now fixed.

src/lib/Microsoft.Health.Fhir.Ingest/Template/JsonPathContentTemplate.cs
```diff
@@ -87,19 +100,27 @@ private Measurement CreateMeasurementFromToken(JToken token)
...
+                correlationId = GetCorrelationId(token);
+                EnsureArg.IsNotNull(correlationId, nameof(correlationId));
+            }
+
             var measurement = new Measurement
             {
                 DeviceId = deviceId,
                 OccurrenceTimeUtc = timestamp.Value,
                 Type = TypeName,
                 PatientId = GetPatientId(token),
```
namalu, line 121: Should we treat PatientId and EncounterId the same as CorrelationId? Where if there is an expression defined we require the value in the message?

src/lib/Microsoft.Health.Fhir.Ingest/Data/TimePeriodMeasurementObservationGroup.cs
```diff
@@ -0,0 +1,31 @@
...
+namespace Microsoft.Health.Fhir.Ingest.Data
+{
+    public class TimePeriodMeasurementObservationGroup : SortedObservationGroup
+    {
+        public const string DateFormat = "yyyyMMddHHmmssZ";
+        private readonly (DateTime Start, DateTime End) _bondary;
+
+        public TimePeriodMeasurementObservationGroup((DateTime Start, DateTime End) boundary)
+        {
+            _bondary = boundary;
```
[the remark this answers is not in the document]
dustinburson: No, it is value type, can never be null.  The implicit tuple syntax i.e. ( , ) is compiled as a struct.

src/lib/Microsoft.Health.Fhir.Ingest/Template/JsonPathContentTemplate.cs
dustinburson: This was something I considered but it would change existing behavior.  Right now the values will be normalized and then generate specific errors in the FHIR conversion side that will be logged in App Insights.

dustinburson:
> Please update the Configuration document with information on correlationId.

Added documentation.

PR #49, opened 2020-06-26
Add FHIR versioning handling strategy to take its selection in template from customer, as well as update the instruction documents.

ms-teli:
Add FHIR versioning handling strategy to take its selection in template from customer, as well as update the instruction documents.

Test Result

| Support | Test Result |
|-|-|
| Managed identity deployment | Succeeded |
| Un-managed identity deployment | Succeeded |
| Sandbox deployment | Succeeded |
| Docker | Succeeded |
| Local VS | Succeeded |

### DIFF
--- a/src/func/Microsoft.Health.Fhir.Ingest.Host/Microsoft.Health.Fhir.Ingest.Host.csproj
+++ b/src/func/Microsoft.Health.Fhir.Ingest.Host/Microsoft.Health.Fhir.Ingest.Host.csproj
@@ -12,6 +12,9 @@
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <WarningsAsErrors />
   </PropertyGroup>
+  <PropertyGroup>
+    <FhirVersion Condition="'$(FhirVersion)' == ''">R4</FhirVersion>
+  </PropertyGroup>  
   <ItemGroup>
     <PackageReference Include="Microsoft.Azure.EventHubs.Processor" Version="4.1.0" />
     <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.14" />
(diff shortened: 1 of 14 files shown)


src/func/Microsoft.Health.Fhir.Ingest.Host/Microsoft.Health.Fhir.Ingest.Host.csproj
```diff
@@ -12,6 +12,9 @@
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <WarningsAsErrors />
   </PropertyGroup>
+  <PropertyGroup>
+    <FhirVersion Condition="'$(FhirVersion)' == ''">R4</FhirVersion>
```
dustinburson, line 16: This is interesting but not sure I like this.  Would prefer having a separate project for R4 and R5 that way both solutions could be open at the same time.  This also in the PR and CI builds will only compile against the R4 version.  Once we add R5 support we will want to make sure both versions are compile.  It can stay as is but I think will need to be adjusted when we do add R5 support.

ms-teli: Thanks for reviewing. One tradeoff of creating a separate project for different FHIR versions is that we will likely have to duplicate all the functions' code in the new projects. In my tests, I found that the Azure Function only scans the function entry-point attributes in the start up Assembly, not in all its dependencies. I've not gotten a clear answer yet of whether it's no or it's possible to register the assembly to let the framework scan in some way. Let's review this topic when the R5 work comes up again. Meantime, I'll reach out to broader Azure community to find something.

PR #50, opened 2020-06-30
Update the resource identity service type with Assembly attribute and short names

ms-teli:
- Add the Assembly attribute to register the implementation classes.
- Use structured enum data type for ResourceIdentityServiceType.
- Update Unit Tests with new enum type.

### DIFF
--- a/deploy/templates/consumption-azuredeploy.json
+++ b/deploy/templates/consumption-azuredeploy.json
@@ -180,7 +180,8 @@
     "app_service_resource_id": "[resourceId('Microsoft.Web/sites', variables('app_service_name'))]",
     "deploy_source_code": "[and(not(empty(parameters('repositoryUrl'))),not(empty(parameters('repositoryBranch'))))]",
     "sender_role": "[concat('/subscriptions/', subscription().subscriptionId, '/providers/Microsoft.Authorization/roleDefinitions/', '2b629674-e913-4c01-ae53-ef4638d8f975')]",
-    "receiver_role": "[concat('/subscriptions/', subscription().subscriptionId, '/providers/Microsoft.Authorization/roleDefinitions/', 'a638d3c7-ab3a-418d-83e6-5f17a39d4fde')]"
+    "receiver_role": "[concat('/subscriptions/', subscription().subscriptionId, '/providers/Microsoft.Authorization/roleDefinitions/', 'a638d3c7-ab3a-418d-83e6-5f17a39d4fde')]",
+    "iomt_fhir_connector_name": "[concat(parameters('FhirVersion'), ':', 'ResourceIdentity', parameters('ResourceIdentityServiceType'))]"
   },
   "resources": [
     {
@@ -196,7 +197,7 @@
         "[resourceId('Microsoft.EventHub/namespaces/eventhubs/consumergroups', variables('eventhub_namespace_name'), 'normalizeddata', '$Default')]"
       ],
       "tags": {
-        "IomtFhirConnector": "[parameters('ResourceIdentityServiceType')]"
+        "IomtFhirConnector": "[variables('iomt_fhir_connector_name')]"
       },
       "identity": {
         "type": "SystemAssigned"
@@ -274,7 +275,7 @@
       "name": "[variables('eventhub_namespace_name')]",
       "location": "[parameters('ResourceLocation')]",
       "tags": {
-        "IomtFhirConnector": "[parameters('ResourceIdentityServiceType')]"
+        "IomtFhirConnector": "[variables('iomt_fhir_connector_name')]"
       },
       "sku": {
         "name": "Standard",
@@ -430,7 +431,7 @@
       "name": "[variables('storage_account_name')]",
       "location": "[parameters('ResourceLocation')]",
       "tags": {
-        "IomtFhirConnector": "[parameters('ResourceIdentityServiceType')]"
+        "IomtFhirConnector": "[variables('iomt_fhir_connector_name')]"
       },
       "sku": {
         "name": "Standard_RAGRS",
@@ -496,7 +497,7 @@
       "name": "[variables('app_plan_name')]",
       "location": "[parameters('ResourceLocation')]",
       "tags": {
-        "IomtFhirConnector": "[parameters('ResourceIdentityServiceType')]"
+        "IomtFhirConnector": "[variables('iomt_fhir_connector_name')]"
       },
       "sku": {
         "name": "Y1",
@@ -523,7 +524,7 @@
       "name": "[variables('app_service_name')]",
       "location": "[parameters('ResourceLocation')]",
       "tags": {
-        "IomtFhirConnector": "[parameters('ResourceIdentityServiceType')]"
+        "IomtFhirConnector": "[variables('iomt_fhir_connector_name')]"
       },
       "dependsOn": [
         "[resourceId('Microsoft.Web/serverfarms', variables('app_plan_name'))]",
@@ -707,7 +708,7 @@
       "name": "[variables('app_insights_name')]",
       "location": "[parameters('ResourceLocation')]",
       "tags": {
-        "IomtFhirConnector": "[parameters('ResourceIdentityServiceType')]"
+        "IomtFhirConnector": "[variables('iomt_fhir_connector_name')]"
       },
       "kind": "web",
       "properties": {
@@ -722,7 +723,7 @@
       "name": "[variables('key_vault_name')]",
       "location": "[parameters('ResourceLocation')]",
       "tags": {
-        "IomtFhirConnector": "[parameters('ResourceIdentityServiceType')]"
+        "IomtFhirConnector": "[variables('iomt_fhir_connector_name')]"
       },
       "dependsOn": [
         "[variables('app_service_resource_id')]"
--- a/deploy/templates/consumption-azuredeploy.json
+++ b/deploy/templates/consumption-azuredeploy.json
@@ -150,11 +150,11 @@
     "ResourceIdentityServiceType": {
       "type": "string",
       "allowedValues": [
-        "R4DeviceAndPatientLookupIdentityService",
-        "R4DeviceAndPatientCreateIdentityService",
-        "R4DeviceAndPatientWithEncounterLookupIdentityService"
+        "Lookup",
+        "Create",
+        "LookupWithEncounter"
       ],
-      "defaultValue": "R4DeviceAndPatientLookupIdentityService",
+      "defaultValue": "Lookup",
       "metadata": {
         "description": "Configures how patient, device, and other FHIR resource identities are resolved from the ingested data stream."
       }
--- a/deploy/templates/consumption-azuredeploy.json
+++ b/deploy/templates/consumption-azuredeploy.json
@@ -181,7 +181,7 @@
     "deploy_source_code": "[and(not(empty(parameters('repositoryUrl'))),not(empty(parameters('repositoryBranch'))))]",
     "sender_role": "[concat('/subscriptions/', subscription().subscriptionId, '/providers/Microsoft.Authorization/roleDefinitions/', '2b629674-e913-4c01-ae53-ef4638d8f975')]",
     "receiver_role": "[concat('/subscriptions/', subscription().subscriptionId, '/providers/Microsoft.Authorization/roleDefinitions/', 'a638d3c7-ab3a-418d-83e6-5f17a39d4fde')]",
-    "iomt_fhir_connector_name": "[concat(parameters('FhirVersion'), ':', 'ResourceIdentity', parameters('ResourceIdentityServiceType'))]"
+    "iomt_fhir_connector_name": "[concat('ResourceIdentity', ':', parameters('ResourceIdentityServiceType'))]"
   },
   "resources": [
     {
@@ -197,7 +197,8 @@
         "[resourceId('Microsoft.EventHub/namespaces/eventhubs/consumergroups', variables('eventhub_namespace_name'), 'normalizeddata', '$Default')]"
       ],
       "tags": {
-        "IomtFhirConnector": "[variables('iomt_fhir_connector_name')]"
+        "IomtFhirConnector": "[variables('iomt_fhir_connector_name')]",
+        "IomtFhirVersion": "[parameters('FhirVersion')]"
       },
       "identity": {
         "type": "SystemAssigned"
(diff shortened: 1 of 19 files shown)


src/lib/Microsoft.Health.Fhir.Ingest/Config/ResourceIdentityServiceType.cs
```diff
@@ -0,0 +1,28 @@
...
+
+namespace Microsoft.Health.Fhir.Ingest.Config
+{
+    /// <summary>
+    /// Resource Identity Service Type.
+    /// </summary>
+    public enum ResourceIdentityServiceType
+    {
+        /// <summary>
+        /// Create the identity resources if not existed.
```
wi-y, line 14: nit: do you mind changing this to "Create the identity resources if they do not already exist." or something similar? I am not sure the phrase "not existed" is correct although I did generally understand what it meant

wi-y:
In ./docs/Sandbox.md, do you think we should update the reference of "R4DeviceAndPatientCreateIdentityService" to "Create" and add a clarifying sentence? You can also choose to leave it as is if you would like since that is still how things work within in the code, but it might be confusing to users configuring the template since to them the only options are "Lookup", "LookupWithEncounter", and "Create"

src/lib/Microsoft.Health.Fhir.Ingest/Config/ResourceIdentityServiceType.cs
ms-teli: Thank you for pointing it out, I'll correct it.

ms-teli:
> In ./docs/Sandbox.md, do you think we should update the reference of "R4DeviceAndPatientCreateIdentityService" to "Create" and add a clarifying sentence? You can also choose to leave it as is if you would like since that is still how things work within in the code, but it might be confusing to users configuring the template since to them the only options are "Lookup", "LookupWithEncounter", and "Create"

I also think the docs should use consistent terms. I'll fix it. Thanks for reviewing.

src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4DeviceAndPatientCreateIdentityService.cs
```diff
@@ -9,9 +9,14 @@
 using Hl7.Fhir.Rest;
 using Microsoft.Health.Extensions.Fhir;
 using Microsoft.Health.Extensions.Fhir.Service;
+using Microsoft.Health.Fhir.Ingest.Config;
 using Microsoft.Health.Fhir.Ingest.Data;
+using Microsoft.Health.Fhir.Ingest.Host;
+using Microsoft.Health.Fhir.Ingest.Service;
 using Model = Hl7.Fhir.Model;
 
+[assembly: ResourceIdentityService(ResourceIdentityServiceType.Create, typeof(R4DeviceAndPatientCreateIdentityService))]
```
dustinburson, line 18: You want to have the attribute on the class not the assembly. If you do that you don't need the type attribute since it is implicit since the attribute is already associated with the desired class.

src/lib/Microsoft.Health.Fhir.Ingest/Config/ResourceIdentityServiceType.cs
```diff
@@ -0,0 +1,28 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Fhir.Ingest.Config
+{
+    /// <summary>
+    /// Resource Identity Service Type.
+    /// </summary>
+    public enum ResourceIdentityServiceType
```
dustinburson, line 11: Please make sure this resolves properly when changed to an enumeration.  Specifically you want to make sure builder.Services.Configure<ResourceIdentityOptions>(config.GetSection("ResourceIdentity")); in MeasurementFhirImportExtensions is properly resolving the type.

src/lib/Microsoft.Health.Fhir.Ingest/Service/ResourceIdentityServiceAttribute.cs
```diff
@@ -0,0 +1,24 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using Microsoft.Health.Fhir.Ingest.Config;
+
+namespace Microsoft.Health.Fhir.Ingest.Host
+{
+    [AttributeUsage(AttributeTargets.Assembly, AllowMultiple = true, Inherited = false)]
```
namalu, line 11: I'm curious if there is a specific reason you are using an assembly attribute rather than a class attribute. If a class attribute was used, you could remove the ClassType property here, and when you create the identity service registry in the ResourceIdentityServiceFactory, you can call GetTypes() on the assemblies and .GetCustomAttributes<ResourceIdentityServiceAttribute>() to find classes that have been attributed.

I'm 100% sure this works as is, I'm just curious if there are other reasons.

dustinburson: This is my same concern as well.

src/lib/Microsoft.Health.Fhir.Ingest/Config/ResourceIdentityServiceType.cs
ms-teli: Yes. I've confirmed the string value can be resolved to this enum correctly with all our templates.

src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4DeviceAndPatientCreateIdentityService.cs
ms-teli: Thanks for asking, I compared two approaches and decide to use the assembly one.

I did some research of how to get all types given an attribute type, and I didn't find anything for that, probably the type doesn't get reverse indexed with the attributes. So the only solution would be given a type, check whether it has some attributes. In this case, in the class attribute solution, I've to iterate all types under all assemblies to examine whether the attribute has been attached. That would cost like lots of waste, including a number of nuget trivia and infra packages. 

With the assembly attribute solution, I can filter in only one loop at assembly level (it's my view, not sure it's also the .NET's view : ) ) and retrieve the types from the assembly attributes list. I think this way is a little more scalable and extensible. The developer can also choose to manage the attribute all together in AssemblyInfo.cs or some other file served as a registry when the types grow more.

I borrowed some idea from https://github.com/Azure/azure-webjobs-sdk/blob/e5417775bcb8c8d3d53698932ca8e4e265eac66d/src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsStartupAttribute.cs Though the purpose not exactly same, but it also manage the types injection during bootstrapping time.

Hopefully, this explained but I'm opened to discuss more on this if there were anything not covered.

src/lib/Microsoft.Health.Fhir.Ingest/Service/ResourceIdentityServiceAttribute.cs
ms-teli: Good question and thanks for asking also. I did spend some time on evaluating this. So I've replied Dustin's comment with more details. The basic idea is that the class attribute seems requiring me to iterate all types in all assemblies, thinking of a tree, that's all leaf nodes. Rather than that, I'd like to prune some branch out at assembly level. And with the types just added in the attribute, I don't have to go through all types but looking them up.  Please let me know if there is anything missed though.

ms-teli:
I've updated another revision with the wording fixes and new value format for resource tags. One example is: ```IomtFhirConnector : R4:ResourceIdentityLookup```

src/lib/Microsoft.Health.Fhir.Ingest/Service/ResourceIdentityServiceAttribute.cs
namalu: Thanks for the detailed write up. I have 2 opinions when it comes to this:

1. If we decide to use assembly attributes, I think they should be added to AssemblyInfo.cs rather than in each class. To me, it looks out of place to have the assembly attributes in class files.
2. While I'm okay with either solution, my personal preference would be to attribute the class. I realize this is not as efficient, but I think it makes it easier to read and forces the attribute to be written in the class file.

ms-teli: I think it makes sense to add them all in the AssemblyInfo.cs, so we can manage them altogether as a registry, which is easier to browse and maintain from another perspective. I'll update once this option is finalized.

deploy/templates/consumption-azuredeploy.json
```diff
@@ -150,11 +150,11 @@
     "ResourceIdentityServiceType": {
```
dustinburson, line 150: Can we rename this to ResourceIdentityResolutionType?  Will reflect current naming else where.

src/lib/Microsoft.Health.Fhir.Ingest/Config/ResourceIdentityOptions.cs
```diff
@@ -7,7 +7,7 @@ namespace Microsoft.Health.Fhir.Ingest.Config
 {
     public class ResourceIdentityOptions
     {
-        public string ResourceIdentityServiceType { get; set; } = "R4DeviceAndPatientLookupIdentityService";
+        public string ResourceIdentityServiceType { get; set; } = "Lookup";
```
dustinburson, line 10: I would update this to be your enumeration, i.e. ResourceIdentityServiceType.Lookup.ToString().

src/lib/Microsoft.Health.Fhir.Ingest/Service/ResourceIdentityServiceAttribute.cs
```diff
@@ -0,0 +1,20 @@
...
+// -------------------------------------------------------------------------------------------------
+
+using System;
+
+namespace Microsoft.Health.Fhir.Ingest.Host
+{
+    [AttributeUsage(AttributeTargets.Class, AllowMultiple = true, Inherited = false)]
+    public sealed class ResourceIdentityServiceAttribute : Attribute
+    {
+        public ResourceIdentityServiceAttribute(string type)
```
dustinburson, line 13: You can have an alternative ctor that takes your enumeration. i.e. public ResourceIdentityServiceAttribute(ResourceIdentityServiceType type) : this(type.ToString())

src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4DeviceAndPatientCreateIdentityService.cs
```diff
@@ -10,10 +10,13 @@
 using Microsoft.Health.Extensions.Fhir;
 using Microsoft.Health.Extensions.Fhir.Service;
 using Microsoft.Health.Fhir.Ingest.Data;
+using Microsoft.Health.Fhir.Ingest.Host;
 using Model = Hl7.Fhir.Model;
 
 namespace Microsoft.Health.Fhir.Ingest.Service
 {
+    [ResourceIdentityService("Create")]
```
dustinburson, line 18: If you add optional enumeration ctor then you can update to [ResourceIdentityService(ResourceIdentityServiceType.Create)]

src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4DeviceAndPatientCreateIdentityService.cs
```diff
@@ -10,10 +10,13 @@
 using Microsoft.Health.Extensions.Fhir;
 using Microsoft.Health.Extensions.Fhir.Service;
 using Microsoft.Health.Fhir.Ingest.Data;
+using Microsoft.Health.Fhir.Ingest.Host;
 using Model = Hl7.Fhir.Model;
 
 namespace Microsoft.Health.Fhir.Ingest.Service
 {
+    [ResourceIdentityService("Create")]
+    [ResourceIdentityService("R4DeviceAndPatientCreateIdentityService")]
```
dustinburson, line 19: Better to have this as [ResourceIdentityService(nameof(R4DeviceAndPatientCreateIdentityService))] to avoid typos in the string.

ms-teli: Updated.

src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4DeviceAndPatientCreateIdentityService.cs
ms-teli: Updated.

src/lib/Microsoft.Health.Fhir.Ingest/Service/ResourceIdentityServiceAttribute.cs
ms-teli: Thanks for this suggestion. Updated.

src/lib/Microsoft.Health.Fhir.Ingest/Config/ResourceIdentityOptions.cs
ms-teli: Updated.

deploy/templates/consumption-azuredeploy.json
```diff
@@ -180,7 +180,8 @@
     "app_service_resource_id": "[resourceId('Microsoft.Web/sites', variables('app_service_name'))]",
     "deploy_source_code": "[and(not(empty(parameters('repositoryUrl'))),not(empty(parameters('repositoryBranch'))))]",
     "sender_role": "[concat('/subscriptions/', subscription().subscriptionId, '/providers/Microsoft.Authorization/roleDefinitions/', '2b629674-e913-4c01-ae53-ef4638d8f975')]",
-    "receiver_role": "[concat('/subscriptions/', subscription().subscriptionId, '/providers/Microsoft.Authorization/roleDefinitions/', 'a638d3c7-ab3a-418d-83e6-5f17a39d4fde')]"
+    "receiver_role": "[concat('/subscriptions/', subscription().subscriptionId, '/providers/Microsoft.Authorization/roleDefinitions/', 'a638d3c7-ab3a-418d-83e6-5f17a39d4fde')]",
+    "iomt_fhir_connector_name": "[concat(parameters('FhirVersion'), ':', 'ResourceIdentity', parameters('ResourceIdentityServiceType'))]"
```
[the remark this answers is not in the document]
ms-teli: Make sense, thanks for the suggestion. I'll update it.

deploy/templates/consumption-azuredeploy.json
ms-teli: Synced up offline. Would like to discuss the scope of this naming convention update, either only for the names on the templates or elsewhere for all other keys, variables etc. For either case, I'll do a separate PR.

PR #54, opened 2020-07-15
Support shallow validations on template schemas

ms-teli:
Changes:
1. Pulled the templates and factories out into a standalone libraries with basic validations for the schema requirements.
2. Added UTs to cover valid and invalid scenarios.
3. Make API changes to use the .NET STANDARD 2.0.

Limitations:
1. As for now, it's hard to get the rid of factories in validations. The chained factory wrappers are used to handle template variations in the generalized workflow. Another option would be to tweak the JSON.NET library to come up with another framework, which could be very similar but time consuming. On the other side, by having the validation in the factory logic, it also does one stop shop for deserialization and validation together.
2. This PR hasn't added the deep relational validation yet, for example, the schema requirements sit on multiple objects, like mutual exclusive, or collection of objects, like non-empty etc. That can be done in next iteration or a new revision when the current way gets consented by team.
3. The whole templating workflow and strategy will be revisited from a higher level with holistic view as well as some new roadmap or requirements. So the goal of this PR is to manage the affecting areas, introducing basic validation logics running for the current templates before the new solution implemented. It also forms shared dependency for other platforms and OSS function runtimes, which will be built with more sharing logic gradually.


src/lib/Microsoft.Health.Fhir.Ingest.Common/Data/IMeasurement.cs
```diff
@@ -1,29 +1,29 @@
-﻿// -------------------------------------------------------------------------------------------------
-// Copyright (c) Microsoft Corporation. All rights reserved.
-// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
-// -------------------------------------------------------------------------------------------------
-
-using System;
-using System.Collections.Generic;
-
-namespace Microsoft.Health.Fhir.Ingest.Data
```
namalu, line 9: Should we update the namespace in all of these files? 
Microsoft.Health.Fhir.Ingest.Common.Data

src/lib/Microsoft.Health.Fhir.Ingest.Common/Template/FhirCodeableConcept.cs
```diff
@@ -4,14 +4,17 @@
 // -------------------------------------------------------------------------------------------------
 
 using System.Collections.Generic;
+using Newtonsoft.Json;
 
 namespace Microsoft.Health.Fhir.Ingest.Template
 {
     public class FhirCodeableConcept
     {
+        [JsonProperty(Required = Required.Always)]
```
namalu, line 13: It would actually still be a valid Observation if Text was null. Does the Function use this property for something that would make this required?

test/Microsoft.Health.Fhir.Ingest.UnitTests/Microsoft.Health.Fhir.Ingest.UnitTests.csproj
```diff
@@ -39,42 +39,9 @@
     <ProjectReference Include="..\Microsoft.Health.Tests.Common\Microsoft.Health.Tests.Common.csproj" />
   </ItemGroup>
   <ItemGroup>
-    <None Update="TestInput\data_CodeValueFhirTemplate_Components.json">
```
namalu, line 42: Was this change intentional? If so, should the other .json files not be copied to the output directory?

src/lib/Microsoft.Health.Fhir.Ingest.Common/Data/IMeasurement.cs
dustinburson: I don't think so.  We don't need to align libraries with the namespaces.  This is a refactor to split out the libraries.  I would suggest the library should be call Microsoft.Health.Fhir.Ingest.Schema or Microsoft.Health.Fhir.Ingest.Contracts since that is what is being stored in the library.

Microsoft.Health.Fhir.Ingest.sln
```diff
@@ -71,6 +71,10 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "healthkitOnFhir", "healthki
 		sample\templates\healthkitOnFhir\fhirmapping.json = sample\templates\healthkitOnFhir\fhirmapping.json
 	EndProjectSection
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.Health.Fhir.Ingest.Common", "src\lib\Microsoft.Health.Fhir.Ingest.Common\Microsoft.Health.Fhir.Ingest.Common.csproj", "{37F9E4E3-851E-4B8A-BECA-FD4388679E22}"
```
dustinburson, line 74: Mentioned in other comment but I believe Microsoft.Health.Fhir.Ingest.Schema is better fit.  We already have several common libraries and another could be confusing.  Plus I don't think we should have anything in this library except schemas going forward.

src/lib/Microsoft.Health.Fhir.Ingest.Common/Microsoft.Health.Fhir.Ingest.Common.csproj
```diff
@@ -0,0 +1,36 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
```
dustinburson, line 3: You will want to set the default namespace to  Microsoft.Health.Fhir.Ingest.Data.

src/lib/Microsoft.Health.Fhir.Ingest.Common/Template/CodeValueFhirTemplateFactory.cs
```diff
@@ -1,32 +1,50 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System.Collections.Generic;
+using System.Linq;
+using EnsureThat;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Health.Fhir.Ingest.Template
```
dustinburson, line 11: I thought the templates were going in a separate library/assembly than the IMeasurement related classes?

src/lib/Microsoft.Health.Fhir.Ingest.Common/Template/FhirCodeableConcept.cs
dustinburson: Yes this isn't required.

src/lib/Microsoft.Health.Fhir.Ingest.Common/Template/FhirCodeableConcept.cs
```diff
@@ -4,14 +4,17 @@
...
         public string Text { get; set; }
 
 #pragma warning disable CA2227
+        [JsonProperty(Required = Required.Always)]
```
dustinburson, line 17: Not required

src/lib/Microsoft.Health.Fhir.Ingest.Common/Template/FhirTemplate.cs
```diff
@@ -1,14 +1,18 @@
...
+using Newtonsoft.Json;
+
+namespace Microsoft.Health.Fhir.Ingest.Template
+{
+    public abstract class FhirTemplate : IFhirTemplate
+    {
+        [JsonProperty(Required = Required.Always)]
+        public virtual string TypeName { get; set; }
+
+        [JsonProperty(Required = Required.Always)]
```
dustinburson, line 15: Not required.  Default will be used.

src/lib/Microsoft.Health.Fhir.Ingest.Common/Template/FhirTemplate.cs
```diff
@@ -1,14 +1,18 @@
...
+// -------------------------------------------------------------------------------------------------
+
+using Newtonsoft.Json;
+
+namespace Microsoft.Health.Fhir.Ingest.Template
+{
+    public abstract class FhirTemplate : IFhirTemplate
+    {
+        [JsonProperty(Required = Required.Always)]
+        public virtual string TypeName { get; set; }
```
dustinburson, line 13: Type is the only required field since this is where determine the binding the class representing the template.

test/Microsoft.Health.Fhir.Ingest.UnitTests/Microsoft.Health.Fhir.Ingest.UnitTests.csproj
ms-teli: These tests are moved along with the source code to a new assembly.

src/lib/Microsoft.Health.Fhir.Ingest.Common/Template/FhirCodeableConcept.cs
ms-teli: Thank you both for pointing out. I'll remove this requirement accordingly.

src/lib/Microsoft.Health.Fhir.Ingest.Common/Data/IMeasurement.cs
ms-teli: The namespace will remain same in this refactoring work to avoid other changes on the consumer classes.

For the assembly name, I think "Schema" suggested by Dustin would sound better. I'll update it.

Microsoft.Health.Fhir.Ingest.sln
ms-teli: Cool, thanks for the suggestion. Agree that the Schema would be better, I will update.

src/lib/Microsoft.Health.Fhir.Ingest.Common/Template/FhirCodeableConcept.cs
ms-teli: Will remove.

src/lib/Microsoft.Health.Fhir.Ingest.Common/Template/FhirTemplate.cs
ms-teli: Will remove.

src/lib/Microsoft.Health.Fhir.Ingest.Common/Template/FhirTemplate.cs
ms-teli: Will remove. Thanks for double checking on these requirements.

src/lib/Microsoft.Health.Common/Handler/IResponsibilityHandler.cs
```diff
@@ -13,6 +13,38 @@ namespace Microsoft.Health.Common.Handler
     public interface IResponsibilityHandler<in TRequest, out TResult>
         where TResult : class
     {
+        /// <summary>
+        /// Evalue the request with chained responsibility handlers.
```
ms-teli, line 17: A typo. Will fix before merge.

src/lib/Microsoft.Health.Common/Handler/IResponsibilityHandler.cs
```diff
@@ -13,6 +13,38 @@ namespace Microsoft.Health.Common.Handler
...
+
+        /// <summary>
+        /// Evalue the request with chained responsibility handlers with a sharable
+        /// context container.
+        /// </summary>
+        /// <typeparam name="TContext">The type of a context container shared by chained handlers.</typeparam>
+        /// <param name="request">The request to be evaluated.</param>
+        /// <param name="context">The context container.</param>
+        /// <returns>The evaluation result.</returns>
+        TResult Evaluate(TRequest request, out TContext context);
```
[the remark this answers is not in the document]
ms-teli: Discussed with Dustin offline. Making the context inside of the req/res would make more sense. So far, we only need to make  that in the response. 

Will Update.

src/lib/Microsoft.Health.Fhir.Ingest.Template/HandlerProxyTemplateFactory.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        /// <summary>
+        /// Create mapping template for the container of serialzied JSON template.
+        /// Return the deserialized mapping object with validation errors.
+        /// </summary>
+        /// <typeparam name="TContext">The type of context container</typeparam>
+        /// <param name="jsonTemplate">JSON template container </param>
+        /// <param name="errors">Template validation errors</param>
+        /// <returns>Mapping template object.</returns>
+        public abstract TOutput Create(TInput jsonTemplate, out IList<string> errors);
```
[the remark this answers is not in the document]
ms-teli: Will update.

src/lib/Microsoft.Health.Fhir.Ingest.Template/HandlerProxyTemplateFactory.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        TOutput IResponsibilityHandler<TInput, TOutput, IList<string>>.Evaluate(TInput request) => Create(request);
+
+        TOutput IResponsibilityHandler<TInput, TOutput, IList<string>>.Evaluate(TInput request, out IList<string> errors) => Create(request, out errors);
+
+        protected JsonSerializer GetJsonSerializer()
+        {
+            return JsonSerializer.Create(new JsonSerializerSettings
+            {
+                MissingMemberHandling = MissingMemberHandling.Error,
```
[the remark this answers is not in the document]
ms-teli: I think this is the other way around. It means that if JSON has a property that doesn't have a match in the Class schema, it would fail. https://www.newtonsoft.com/json/help/html/DeserializeMissingMemberHandling.htm

Since this caused confusion to the reading, I'll leave a short comment on it.

src/lib/Microsoft.Health.Fhir.Ingest.Common/Microsoft.Health.Fhir.Ingest.Common.csproj
ms-teli: Discussed offline, I'll make the change.

src/lib/Microsoft.Health.Fhir.Ingest.Common/Template/CodeValueFhirTemplateFactory.cs
ms-teli: Discussed offline, I'll make split the the Measurement schema and make another PR since this one is pretty bulky.

ms-teli:
Replace this pull request with https://github.com/microsoft/iomt-fhir/pull/56

PR #55, opened 2020-07-21
Create new projects for template files and schema files, renaming the exiting files to their own modules.

ms-teli:
Changes include:
1. Move template to a sharable assembly that is using .NET Standard 2.0.
2. Move the data model of the Measurement and the MeasurementGroup to a sharable assembly.
3. Move and update the unit tests.


test/Microsoft.Health.Fhir.Ingest.Template.UnitTests/IotJsonPathContentTemplateTests.cs
```diff
@@ -38,17 +39,15 @@ public class IotJsonPathContentTemplateTests
         [FileData(@"TestInput/data_IotHubPayloadExample.json")]
         public void GivenTemplateAndSingleValidToken_WhenGetMeasurements_ThenSingleMeasurementReturned_Test(string eventJson)
         {
-            var evt = EventDataTestHelper.BuildEventFromJson(eventJson);
-            var token = new EventDataWithJsonBodyToJTokenConverter().Convert(evt);
-
+            var token = JsonConvert.DeserializeObject<JToken>(eventJson);
```
dustinburson, line 42: Why the change from the EventDataTestHelper Event to a JToken?

ms-teli: Thanks for reviewing.

Because otherwise, the EventHub dependencies would still be required in this assembly which is not quite necessary for the measurement models. These tests were primarily for the template so this should not reduce the testing coverage by skipping the converters. And in the original package, we still have UTs to cover the conversions.

dustinburson: Makes sense. Thanks for clarifying.

PR #56, opened 2020-07-22
Add validation logics and error handling.

ms-teli:
- Add validation logics and error handling.
- Add and update validation unit tests.
- Examples:
```
                var templateContext = CollectionContentTemplateFactory.Default.Create(templateDefinitions);
                if (!template.IsValid(out string error))
                {
                    throw new InvalidTemplateException($"The current Device template definition is not valid: {error}");
                }
``` 

This is spitted from https://github.com/microsoft/iomt-fhir/pull/54 to demo part for validation changes.

### DIFF
--- a/src/func/Microsoft.Health.Fhir.Ingest.Host/IomtConnectorFunctions.cs
+++ b/src/func/Microsoft.Health.Fhir.Ingest.Host/IomtConnectorFunctions.cs
@@ -58,6 +58,10 @@ namespace Microsoft.Health.Fhir.Ingest.Service
                 EnsureArg.IsNotNull(events, nameof(events));
 
                 var template = CollectionContentTemplateFactory.Default.Create(templateDefinitions);
+                if (!template.IsValid(out string error))
+                {
+                    throw new InvalidTemplateException($"The current Device template definition is not valid: {error}");
+                }
 
                 log.LogMetric(Metrics.DeviceEvent, events.Length);
                 IDataNormalizationService<EventData, IMeasurement> dataNormalizationService = new MeasurementEventNormalizationService(log, template);
--- a/src/func/Microsoft.Health.Fhir.Ingest.Host/IomtConnectorFunctions.cs
+++ b/src/func/Microsoft.Health.Fhir.Ingest.Host/IomtConnectorFunctions.cs
@@ -57,11 +57,9 @@ namespace Microsoft.Health.Fhir.Ingest.Service
                 EnsureArg.IsNotNull(templateDefinitions, nameof(templateDefinitions));
                 EnsureArg.IsNotNull(events, nameof(events));
 
-                var template = CollectionContentTemplateFactory.Default.Create(templateDefinitions);
-                if (!template.IsValid(out string error))
-                {
-                    throw new InvalidTemplateException($"The current Device template definition is not valid: {error}");
-                }
+                var templateContext = CollectionContentTemplateFactory.Default.Create(templateDefinitions);
+                templateContext.EnsureValid();
+                var template = templateContext.Template;
 
                 log.LogMetric(Metrics.DeviceEvent, events.Length);
                 IDataNormalizationService<EventData, IMeasurement> dataNormalizationService = new MeasurementEventNormalizationService(log, template);
--- a/src/lib/Microsoft.Health.Fhir.Ingest.Template/FhirLookupTemplate.cs
+++ b/src/lib/Microsoft.Health.Fhir.Ingest.Template/FhirLookupTemplate.cs
@@ -13,11 +13,11 @@ namespace Microsoft.Health.Fhir.Ingest.Template
 {
     public class FhirLookupTemplate : ILookupTemplate<IFhirTemplate>
     {
-        private readonly IList<string> _serializationErrors = new List<string>();
+        private readonly IList<string> _templateErrors = new List<string>();
 
         private readonly IDictionary<string, IFhirTemplate> _templates = new Dictionary<string, IFhirTemplate>(StringComparer.InvariantCultureIgnoreCase);
 
-        public IList<string> SerializationErrors => _serializationErrors;
+        public IList<string> TemplateErrors => _templateErrors;
 
         public FhirLookupTemplate RegisterTemplate(IFhirTemplate fhirTemplate)
         {
@@ -33,18 +33,18 @@ namespace Microsoft.Health.Fhir.Ingest.Template
                 }
                 else
                 {
-                    _serializationErrors.Add($"Duplicate template defined for type name: '{fhirTemplate.TypeName}'");
+                    _templateErrors.Add($"Duplicate template defined for type name: '{fhirTemplate.TypeName}'");
                 }
             }
             else if (fhirTemplate.TypeName?.Trim().Length == 0)
             {
-                _serializationErrors.Add($"Empty type name is not allowed.");
+                _templateErrors.Add($"Empty type name is not allowed.");
             }
 
             if (!hasTemplateRegistered)
             {
-                fhirTemplate.SerializationErrors.ToList()
-                    .ForEach(e => _serializationErrors.Add(e));
+                fhirTemplate.TemplateErrors.ToList()
+                    .ForEach(e => _templateErrors.Add(e));
             }
 
             return this;
@@ -64,7 +64,7 @@ namespace Microsoft.Health.Fhir.Ingest.Template
         public IEnumerable<ValidationResult> Validate(ValidationContext validationContext)
         {
             List<ValidationResult> aggregatedResult = new List<ValidationResult>();
-            _serializationErrors.ToList()
+            _templateErrors.ToList()
                 .ForEach(e => aggregatedResult.Add(new ValidationResult(e)));
 
             _templates.Values.ToList()
--- a/src/lib/Microsoft.Health.Fhir.Ingest.Template/FhirLookupTemplate.cs
+++ b/src/lib/Microsoft.Health.Fhir.Ingest.Template/FhirLookupTemplate.cs
@@ -23,18 +23,25 @@ namespace Microsoft.Health.Fhir.Ingest.Template
         {
             EnsureArg.IsNotNull(fhirTemplate, nameof(fhirTemplate));
 
-            if (fhirTemplate?.TypeName != null)
+            bool hasTemplateRegistered = false;
+            if (!string.IsNullOrWhiteSpace(fhirTemplate.TypeName))
             {
                 if (!_templates.ContainsKey(fhirTemplate.TypeName))
                 {
                     _templates.Add(fhirTemplate.TypeName, fhirTemplate);
+                    hasTemplateRegistered = true;
                 }
                 else
                 {
                     _serializationErrors.Add($"Duplicate template defined for type name: '{fhirTemplate.TypeName}'");
                 }
             }
-            else
+            else if (fhirTemplate.TypeName?.Trim().Length == 0)
+            {
+                _serializationErrors.Add($"Empty type name is not allowed.");
+            }
+
+            if (!hasTemplateRegistered)
             {
                 fhirTemplate.SerializationErrors.ToList()
                     .ForEach(e => _serializationErrors.Add(e));
--- a/src/lib/Microsoft.Health.Fhir.Ingest.Template/FhirLookupTemplate.cs
+++ b/src/lib/Microsoft.Health.Fhir.Ingest.Template/FhirLookupTemplate.cs
@@ -5,23 +5,40 @@
 
 using System;
 using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+using System.Linq;
 using EnsureThat;
 
 namespace Microsoft.Health.Fhir.Ingest.Template
 {
     public class FhirLookupTemplate : ILookupTemplate<IFhirTemplate>
     {
+        private readonly IList<string> _serializationErrors = new List<string>();
+
         private readonly IDictionary<string, IFhirTemplate> _templates = new Dictionary<string, IFhirTemplate>(StringComparer.InvariantCultureIgnoreCase);
+
+        public IList<string> SerializationErrors => _serializationErrors;
 
         public FhirLookupTemplate RegisterTemplate(IFhirTemplate fhirTemplate)
         {
             EnsureArg.IsNotNull(fhirTemplate, nameof(fhirTemplate));
-            if (_templates.ContainsKey(fhirTemplate.TypeName))
-            {
-                throw new InvalidOperationException($"Duplicate template defined for {fhirTemplate.TypeName}");
-            }
 
-            _templates.Add(fhirTemplate.TypeName, fhirTemplate);
+            if (fhirTemplate?.TypeName != null)
+            {
+                if (!_templates.ContainsKey(fhirTemplate.TypeName))
+                {
+                    _templates.Add(fhirTemplate.TypeName, fhirTemplate);
+                }
+                else
+                {
+                    _serializationErrors.Add($"Duplicate template defined for type name: '{fhirTemplate.TypeName}'");
+                }
+            }
+            else
+            {
+                fhirTemplate.SerializationErrors.ToList()
+                    .ForEach(e => _serializationErrors.Add(e));
+            }
 
             return this;
         }
@@ -35,6 +52,18 @@ namespace Microsoft.Health.Fhir.Ingest.Template
             }
 
             return template;
+        }
+
+        public IEnumerable<ValidationResult> Validate(ValidationContext validationContext)
+        {
+            List<ValidationResult> aggregatedResult = new List<ValidationResult>();
+            _serializationErrors.ToList()
+                .ForEach(e => aggregatedResult.Add(new ValidationResult(e)));
+
+            _templates.Values.ToList()
+                .ForEach(t => aggregatedResult.AddRange(t.Validate(validationContext)));
+
+            return aggregatedResult;
         }
     }
 }
--- a/src/lib/Microsoft.Health.Fhir.Ingest.Template/FhirLookupTemplate.cs
+++ b/src/lib/Microsoft.Health.Fhir.Ingest.Template/FhirLookupTemplate.cs
@@ -5,46 +5,32 @@
 
 using System;
 using System.Collections.Generic;
-using System.ComponentModel.DataAnnotations;
-using System.Linq;
 using EnsureThat;
 
 namespace Microsoft.Health.Fhir.Ingest.Template
 {
     public class FhirLookupTemplate : ILookupTemplate<IFhirTemplate>
     {
-        private readonly IList<string> _templateErrors = new List<string>();
-
         private readonly IDictionary<string, IFhirTemplate> _templates = new Dictionary<string, IFhirTemplate>(StringComparer.InvariantCultureIgnoreCase);
-
-        public IList<string> TemplateErrors => _templateErrors;
 
         public FhirLookupTemplate RegisterTemplate(IFhirTemplate fhirTemplate)
         {
             EnsureArg.IsNotNull(fhirTemplate, nameof(fhirTemplate));
 
-            bool hasTemplateRegistered = false;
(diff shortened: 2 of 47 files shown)


src/lib/Microsoft.Health.Fhir.Ingest.Template/HandlerProxyTemplateFactory.cs
```diff
@@ -12,8 +14,29 @@ public abstract class HandlerProxyTemplateFactory<TInput, TOutput> :
...
+
+        protected JsonSerializer GetJsonSerializer()
+        {
+            _serializationErrors.Clear();
+            return JsonSerializer.Create(new JsonSerializerSettings
+            {
+                MissingMemberHandling = MissingMemberHandling.Error,
+                Error = (sender, args) =>
+                {
+                    _serializationErrors.Add(args.ErrorContext.Error.Message);
```
dustinburson, line 36: Handling the error like doesn't seem like the best.  Have you tried catching serialization errors from the caller instead of internalizing this?  My fear is we will have a complete but no functional template as part of the collection.  It would be better to let this template error, catch and record then continue with others.

src/func/Microsoft.Health.Fhir.Ingest.Host/IomtConnectorFunctions.cs
```diff
@@ -58,6 +58,10 @@ public static class IomtConnectorFunctions
                 EnsureArg.IsNotNull(events, nameof(events));
 
                 var template = CollectionContentTemplateFactory.Default.Create(templateDefinitions);
+                if (!template.IsValid(out string error))
```
dustinburson, line 61: This could also be done as an extension method.

test/Microsoft.Health.Fhir.Ingest.Template.UnitTests/CodeValueFhirTemplateFactoryTests.cs
```diff
@@ -280,6 +280,20 @@ public void GivenValidTemplateJsonWithMixedMainAndComponentValues_WhenFactoryCre
...
+        [FileData(@"TestInput/data_CodeValueFhirTemplateInvalid_MissingFields.json")]
+        public void GivenInvalidTemplateJsonMissingFields_WhenFactoryCreate_ThenTemplateCreated_Test(string json)
+        {
+            var templateContainer = JsonConvert.DeserializeObject<TemplateContainer>(json);
+
+            var factory = new CodeValueFhirTemplateFactory();
+
+            var template = factory.Create(templateContainer);
+            Assert.NotNull(template);
+            template.IsValid(out string errors);
```
dustinburson, line 293: should test the return value here.

test/Microsoft.Health.Fhir.Ingest.Template.UnitTests/CodeValueFhirTemplateFactoryTests.cs
```diff
@@ -280,6 +280,20 @@ public void GivenValidTemplateJsonWithMixedMainAndComponentValues_WhenFactoryCre
...
+            Assert.Contains("TypeName", errors);
```
dustinburson, line 294: Should test the template is functional.  A separate test that has a good template and bad template and verifies the bad template is skipped and the good template works would be a good addtition.

dustinburson, line 294: Should test the template is functional.  A separate test that has a good template and bad template and verifies the bad template is skipped and the good template works would be a good addition.

test/Microsoft.Health.Fhir.Ingest.Template.UnitTests/CollectionContentTemplateFactoryTests.cs
```diff
@@ -24,6 +26,11 @@ public void GivenEmptyTemplateCollection_WhenCreate_ThenTemplateReturned_Test(st
         {
             var template = CollectionContentTemplateFactory.Default.Create(json);
             Assert.NotNull(template);
+
```
dustinburson, line 29: Should a test to cover a bad template and assert the failure.  Also, as with the comment in the CodeValueFhirTemplateFactoryTests should verify if we have a CollectionContentTemplate with bad and good templates. the good templates still work.

src/lib/Microsoft.Health.Fhir.Ingest.Template/IContentTemplate.cs
```diff
@@ -9,7 +9,7 @@
 
 namespace Microsoft.Health.Fhir.Ingest.Template
 {
-    public interface IContentTemplate
+    public interface IContentTemplate : ITemplate
```
dustinburson, line 12: As with my comments in ITemplate, I think this is what needs to be a IValidatableObject.  NOt sure there is much merit extending that to the underlying templates.

src/lib/Microsoft.Health.Fhir.Ingest.Template/HandlerProxyTemplateFactory.cs
ms-teli: The Newtonsoft APIs give the option to validate JToken against an schema predefined https://www.newtonsoft.com/json/help/html/JTokenIsValid.htm. Or, it can use given serializer with customized error handler registered to stop it from throwing exception directly, in which we can get a full list of errors.  

This is not logically straightforward since the invocation happened to run two logics together: deserialization and validation. That's against the single responsibility.

In next revision, if we're going to homemade this by us, I'd like to create a framework that has the validation interface to retrieve the data schema requirements and other business validation requirements and run them on top of a de-serialized POCO.

src/lib/Microsoft.Health.Fhir.Ingest.Template/IContentTemplate.cs
ms-teli: I updated it to have the wrapper wrapping the template and other contextual information.

src/lib/Microsoft.Health.Fhir.Ingest.Template/ITemplate.cs
```diff
@@ -0,0 +1,15 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+
+namespace Microsoft.Health.Fhir.Ingest.Template
+{
```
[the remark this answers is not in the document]
ms-teli: I updated it to have the wrapper wrapping the template and other contextual information.

src/lib/Microsoft.Health.Fhir.Ingest.Template/ITemplate.cs
```diff
@@ -0,0 +1,15 @@
...
+    public interface ITemplate : IValidatableObject
```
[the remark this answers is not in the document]
ms-teli: Updated based off our discussion.

src/lib/Microsoft.Health.Fhir.Ingest.Template/IValidatableObjectExtensions.cs
```diff
@@ -0,0 +1,38 @@
...
+// -------------------------------------------------------------------------------------------------
+
+using System.ComponentModel.DataAnnotations;
+using System.Linq;
+using System.Text;
+using EnsureThat;
+
+namespace Microsoft.Health.Fhir.Ingest.Template
+{
+    public static class IValidatableObjectExtensions
```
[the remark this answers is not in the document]
ms-teli: This was a miss. Thanks for pointing out!

Updated based off our discussion.

test/Microsoft.Health.Fhir.Ingest.Template.UnitTests/CodeValueFhirTemplateFactoryTests.cs
ms-teli: Updated and based on our discussion, I modified and let the template-type-level factory to throw invalid template exception.

test/Microsoft.Health.Fhir.Ingest.Template.UnitTests/CodeValueFhirTemplateFactoryTests.cs
ms-teli: Thanks for pointing out. I thought the validity of the template is binary, and was not aware of this case.

I'll update.

test/Microsoft.Health.Fhir.Ingest.Template.UnitTests/CollectionContentTemplateFactoryTests.cs
ms-teli: I'll update.

src/func/Microsoft.Health.Fhir.Ingest.Host/IomtConnectorFunctions.cs
ms-teli: Updated.

src/lib/Microsoft.Health.Fhir.Ingest.Template/HandlerProxyTemplateFactory.cs
```diff
@@ -12,8 +14,29 @@ public abstract class HandlerProxyTemplateFactory<TInput, TOutput> :
...
+        protected IList<string> SerializationErrors
+        {
+            get { return _serializationErrors; }
+        }
+
         public abstract TOutput Create(TInput jsonTemplate);
 
         TOutput IResponsibilityHandler<TInput, TOutput>.Evaluate(TInput request) => Create(request);
+
+        protected JsonSerializer GetJsonSerializer()
```
dustinburson, line 28: What are the reasons for keeping the custom serializer settings? Are the seriailzation errors thrown not enough?

I had written a prior comment but looks like I forgot to publish it.  I am concerned with using a delegate approach for the error handler.  This can lead to memory leaks.  A handler like this is an object reference that can keep the object reference alive and prevents garbage collection.

src/lib/Microsoft.Health.Fhir.Ingest.Template/HandlerProxyTemplateFactory.cs
```diff
@@ -12,8 +14,29 @@ public abstract class HandlerProxyTemplateFactory<TInput, TOutput> :
...
+        {
+            _serializationErrors.Clear();
```
dustinburson, line 30: Coupling the clearing of the errors with getting the serialization doesn't seem correct to me.  It would be better to have errors part of the TOutput return object so error life cycle naturally

src/lib/Microsoft.Health.Fhir.Ingest.Template/TemplateContext.cs
```diff
@@ -0,0 +1,44 @@
...
+        {
+            _template = template;
+        }
+
+        public TTemplate Template
+        {
+            get { return _template; }
+        }
+
+        public ICollection<TemplateError> Errors
```
dustinburson, line 29: Reason for capturing just the message instead of the full exception?

src/lib/Microsoft.Health.Fhir.Ingest.Template/TemplateContext.cs
```diff
@@ -0,0 +1,44 @@
...
+        {
+            get { return _errors; }
+        }
+
+        public IEnumerable<ValidationResult> Validate(ValidationContext validationContext)
+        {
+            List<ValidationResult> aggregatedResult = new List<ValidationResult>();
+            _errors
```
dustinburson, line 37: This can be simplified using Linq, return _errors.Select(e => new ValidationResult(e.Message));

dustinburson: Alternatively you can just iterate through and use yield return.  No need to create an intermediate collection using additional memory.

src/lib/Microsoft.Health.Fhir.Ingest.Template/TemplateError.cs
```diff
@@ -0,0 +1,22 @@
...
+    public class TemplateError
+    {
+        private readonly string _message;
+
+        public TemplateError(string message)
+        {
+            _message = message;
+        }
+
+        public string Message
```
namalu, line 17: nit: You can actually remove the _message field and just use:

public string Message { get; }

You can set the Message property in the constructor. I saw this in a few other places, feel free to update if you like.

src/lib/Microsoft.Health.Fhir.Ingest.Template/IotJsonPathContentTemplateFactory.cs
```diff
@@ -26,7 +26,14 @@ public override IContentTemplate Create(TemplateContainer jsonTemplate)
                 throw new InvalidTemplateException($"Expected an object for the template property value for template type {targetTypeName}.");
             }
 
-            return jsonTemplate.Template.ToObject<IotJsonPathContentTemplate>();
+            var iotJsonPathContentTemplate = jsonTemplate.Template.ToObject<IotJsonPathContentTemplate>(GetJsonSerializer());
+            if (SerializationErrors?.Count > 0)
```
dustinburson, line 30: I did a test.  It looks like the default serializer will stop on the first error so the benefit of this is we get all serialization errors.  I would prefer to modify this so we get a Serialization context that will have a the errors and a reference to the JsonSerializer.  It would also allow you to wrap this error checking and raising logic as part of the context so we have less repeated code.  You also no longer have to bind the errors to the factory now and reset them if there are multiple invocations.  My concern about memory leaks is also gone since we are creating a delegate to link the serializer to a factory with a longer lifetime.

test/Microsoft.Health.Fhir.Ingest.Template.UnitTests/CodeValueFhirTemplateFactoryTests.cs
ms-teli: Updated.

src/lib/Microsoft.Health.Fhir.Ingest.Template/TemplateContext.cs
ms-teli: in my understand, the exception conceptually is for wrapping/ending up abnormal situations in one class or one component. However, to the consumer of that class or component, those could be translated into other legal states. For example, the validation logic expects errors to occur. 

I'd be a bit more cautious to use the exceptions for conveying the information, so I'd modeled the message in a meaningful way to the upper level and that model can somehow be shared as well without a dependency on the exception class.  If we need more info, this model can be extended to contain those as well. Hopefully that would make sense.

src/lib/Microsoft.Health.Fhir.Ingest.Template/TemplateContext.cs
ms-teli: Thanks for the tips. I'll update this.

src/lib/Microsoft.Health.Fhir.Ingest.Template/FhirLookupTemplate.cs
```diff
@@ -16,12 +16,22 @@ public class FhirLookupTemplate : ILookupTemplate<IFhirTemplate>
...
+                if (!_templates.ContainsKey(fhirTemplate.TypeName))
+                {
+                    _templates.Add(fhirTemplate.TypeName, fhirTemplate);
+                }
+                else
+                {
+                    throw new InvalidTemplateException($"Duplicate template defined for type name: '{fhirTemplate.TypeName}'");
+                }
+            }
+            else if (fhirTemplate.TypeName?.Trim().Length == 0)
```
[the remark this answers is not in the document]
ms-teli: The first "if" covered the case when the typeName is not null and not empty. We check the map of templates under this case. For the "else" case, I wanted to alert on the "empty" case but slip on the null case, since the error message of the null case already be covered presumably by the serializer. 

This actually could be improved by a more sophisticated schema validation framework.  The newtonsoft uses Json schema to define more validations. However I don't want to create two copies of schema boilerplates. So one way I think we can do in the future is to utilize the attributes with a validation rules' engine for easily execute these logics.

src/lib/Microsoft.Health.Fhir.Ingest.Template/TemplateError.cs
ms-teli: Thanks for the suggestion. I think that could be cleaner. I'll update

src/lib/Microsoft.Health.Fhir.Ingest.Template/HandlerProxyTemplateFactory.cs
ms-teli: Updated with the extension approach suggested by Dustin. Much clean and better now.

src/lib/Microsoft.Health.Fhir.Ingest.Template/TemplateContext.cs
dustinburson: Thanks for the detailed answer.  The part that is confusing to me is the TemplateError isn't for the user of the template.  It is an intermediate that is used to track state that is then in turned used to generate the validation result.  I would be more in favor if it was an interface ITemplateError { string Message {get;} }.  In that case we could have an implementation that wraps an Exception for these exception based errors and later we can add different types when other validations are tracked.

src/lib/Microsoft.Health.Fhir.Ingest.Template/FhirLookupTemplate.cs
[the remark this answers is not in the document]
dustinburson: I believe Nate is referring to the string.IsNullOrWhiteSpace(fhirTemplate.TypeName) and else if (fhirTemplate.TypeName?.Trim().Length == 0).  IsNullOrWhiteSpace also checks if the string is empty.  The else if can be just an else.  *If* the else if is need then we would other logical conditions that aren't accounted for in the current code which also isn't correct.  Reiterating Nate's suggestion of replacing the else if with an else.

src/func/Microsoft.Health.Fhir.Ingest.Host/IomtConnectorFunctions.cs
```diff
@@ -57,7 +57,9 @@ public static class IomtConnectorFunctions
                 EnsureArg.IsNotNull(templateDefinitions, nameof(templateDefinitions));
                 EnsureArg.IsNotNull(events, nameof(events));
 
-                var template = CollectionContentTemplateFactory.Default.Create(templateDefinitions);
+                var templateContext = CollectionContentTemplateFactory.Default.Create(templateDefinitions);
+                templateContext.EnsureValid();
```
dustinburson, line 61: Suggestion.  Can shorten these two lines if the EnsureValid returns the template, i.e. var template = templateContext.Template.EnsureValid();  Seems a little cleaner.

src/lib/Microsoft.Health.Fhir.Ingest.Template/FhirLookupTemplate.cs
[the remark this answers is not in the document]
ms-teli: Updated the code to remove that unnecessary check.

src/func/Microsoft.Health.Fhir.Ingest.Host/IomtConnectorFunctions.cs
ms-teli: Discussed offline, I'll check on this in the next revision.

src/lib/Microsoft.Health.Fhir.Ingest.Template/TemplateContext.cs
ms-teli: Discussed offline, I'll check on this in the next revision.

src/lib/Microsoft.Health.Fhir.Ingest.Template/HandlerProxyTemplateFactory.cs
ms-teli: Discussed offline. This has been addressed.

PR #58, opened 2020-08-07
Support additional loggers, add metrics dimensions

wi-y:


src/lib/Microsoft.Health.Fhir.Ingest/Telemetry/Metrics/IomtMetrics.cs
```diff
@@ -0,0 +1,152 @@
...
+using System.Collections.Generic;
+using EnsureThat;
+using Microsoft.Health.Fhir.Ingest.Telemetry.Metrics;
+
+namespace Microsoft.Health.Fhir.Ingest.Telemetry
+{
+    /// <summary>
+    /// Defines known metrics and metric dimensions for use in Application Insights
+    /// </summary>
+    public static class IomtMetrics
```
namalu, line 16: Feel free to disregard this idea, but I'm wondering if it makes sense to create an IMetric interface that defines the members of a Metric object, and instead of initializing each one of these below, create a separate class for each one of these that implements the IMetric interface. They can all be derived from a common base class and could contain a static method to create a new instance.

src/lib/Microsoft.Health.Fhir.Ingest/Telemetry/IomtTelemetryLogger.cs
```diff
@@ -0,0 +1,273 @@
...
+                return;
+            }
+
+            string[] dimNames = new string[dimensions.Count];
+            dimensions.Keys.CopyTo(dimNames, 0);
+
+            string[] dimValues = new string[dimensions.Count];
+            dimensions.Values.CopyTo(dimValues, 0);
+
+            switch (dimensionNumber)
```
namalu, line 61: I wonder if, for the sake of encapsulating ugly code, we could create an extension method that hides the conversion implementation.

public static Metric ToMetric(TelemetryClient telemetryClient);

src/func/Microsoft.Health.Fhir.Ingest.Host/Startup.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        public override void Configure(IFunctionsHostBuilder builder)
+        {
+            EnsureArg.IsNotNull(builder, nameof(builder));
+
+            builder.Services.AddSingleton<ITelemetryLogger>(sp =>
+            {
+                var telemetryConfiguration = new TelemetryConfiguration();
+                telemetryConfiguration.TelemetryInitializers.Add(new OperationCorrelationTelemetryInitializer());
+                telemetryConfiguration.InstrumentationKey = Environment.GetEnvironmentVariable("APPINSIGHTS_INSTRUMENTATIONKEY") ?? string.Empty;
```
ms-teli, line 29: Should we make it fail on missing that configuration?

src/func/Microsoft.Health.Fhir.Ingest.Host/IomtConnectorFunctions.cs
```diff
@@ -12,45 +12,50 @@
...
+            _logger = logger;
+        }
+
         [FunctionName("MeasurementCollectionToFhir")]
-        public static async Task<IActionResult> MeasurementCollectionToFhir(
+        public async Task<IActionResult> MeasurementCollectionToFhir(
             [HttpTrigger(AuthorizationLevel.Function, "post", Route = null)] HttpRequest req,
             [Blob("template/%Template:FhirMapping%", FileAccess.Read)] string templateDefinition,
-            [MeasurementFhirImport] MeasurementFhirImportService measurementImportService,
-            ILogger log)
```
ms-teli, line 31: Just curious why we can't ingest the logger instance via method parameter anymore and have to remove the static modifier?

src/lib/Microsoft.Health.Fhir.Ingest/Telemetry/IomtTelemetryLogger.cs
```diff
@@ -0,0 +1,273 @@
...
+                _telemetryClient.TrackException(
+                    new Exception($"Metric {metricName} exceeds the amount of allowed dimensions"));
+                return;
+            }
+
+            string[] dimNames = new string[dimensions.Count];
+            dimensions.Keys.CopyTo(dimNames, 0);
+
+            string[] dimValues = new string[dimensions.Count];
+            dimensions.Values.CopyTo(dimValues, 0);
```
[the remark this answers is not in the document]
wi-y: I don't think so unless we added to dictionary between when we copy the values into the arrays. But just to be safe I will update this to get the values for each key.

src/lib/Microsoft.Health.Fhir.Ingest/Telemetry/IomtTelemetryLogger.cs
wi-y: Will look into this

src/lib/Microsoft.Health.Fhir.Ingest/Telemetry/Metrics/IomtMetrics.cs
wi-y: This is a personal preference, but I like having all of the metrics in one class for now.

src/func/Microsoft.Health.Fhir.Ingest.Host/Startup.cs
wi-y: Yes, that makes sense. I was thinking we did not want it to fail when people are running the function locally and potentially not providing an app insights instrumentation key. But I'll update this to specifically check AZURE_FUNCTIONS_ENVIRONMENT = development, to not have the function fail when running locally.

src/func/Microsoft.Health.Fhir.Ingest.Host/IomtConnectorFunctions.cs
wi-y: Either should work. 

I was reading some Azure Functions examples that weren't using the parameter and instead doing DI in the constructor, so I updated it to match the example.

https://docs.microsoft.com/en-us/azure/azure-monitor/app/azure-functions-supported-features
https://docs.microsoft.com/en-us/azure/azure-functions/functions-monitoring?tabs=cmd

src/lib/Microsoft.Health.Fhir.Ingest/Telemetry/ExceptionTelemetryProcessor.cs
```diff
@@ -31,33 +30,71 @@ public ExceptionTelemetryProcessor()
...
+                        metricValue: 1);
                 }
 
                 return true;
             }
 
             return false;
         }
+
+        public static Metric ConvertExceptionToMetric(System.Type lookupType, Type exType)
```
dustinburson, line 69: If we keep this it would be better written as a switch statement with pattern matching;
https://docs.microsoft.com/en-us/dotnet/csharp/pattern-matching

src/lib/Microsoft.Health.Fhir.Ingest/Telemetry/IomtTelemetryLogger.cs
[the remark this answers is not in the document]
wi-y: Updated

src/lib/Microsoft.Health.Fhir.Ingest/Telemetry/IomtTelemetryLogger.cs
wi-y: updated to use extension method. thank you for the suggestion

src/lib/Microsoft.Health.Fhir.Ingest/Telemetry/ExceptionTelemetryProcessor.cs
```diff
@@ -31,33 +30,71 @@ public ExceptionTelemetryProcessor()
...
 
-        public ExceptionTelemetryProcessor(params Type[] handledExceptionTypes)
+        public ExceptionTelemetryProcessor(params System.Type[] handledExceptionTypes)
         {
-            _handledExceptions = new HashSet<Type>(handledExceptionTypes);
+            _handledExceptions = new HashSet<System.Type>(handledExceptionTypes);
         }
 
-        public virtual bool HandleException(Exception ex, ILogger log)
+        public virtual bool HandleException(Exception ex, ITelemetryLogger log, string connectorStage)
```
dustinburson, line 38: This signature takes an ILogger because it needed to be passed into the function method.  This can probably be refactored to be injected in the ctor with DI and have a cleaner signature.  I can't see a reason currently to keep it as part of the method signature.  Our ITelemetryLogger is locked in during run time currently.

src/func/Microsoft.Health.Fhir.Ingest.Host/Startup.cs
wi-y: Updated to only configure app insights when the setting is not null, but there were some challenges with logging errors during startup. In particular there wouldn't be a place to easily view the error/exception. I have chosen to let the functions startup when the setting isn't populated, because that seemed marginally better than throwing an exception and having no place to view the exception

src/lib/Microsoft.Health.Fhir.Ingest/Telemetry/ExceptionTelemetryProcessor.cs
```diff
@@ -31,33 +30,71 @@ public ExceptionTelemetryProcessor()
...
             EnsureArg.IsNotNull(ex, nameof(ex));
+            EnsureArg.IsNotNull(log, nameof(log));
+
             var exType = ex.GetType();
 
             var lookupType = exType.IsGenericType ? exType.GetGenericTypeDefinition() : exType;
 
             if (_handledExceptions.Contains(lookupType))
             {
-                if (ex is ITelemetryEvent evt)
```
dustinburson, line 48: This is no longer being used and should be removed.

src/lib/Microsoft.Health.Fhir.Ingest/Telemetry/Metrics/IomtMetrics.cs
```diff
@@ -0,0 +1,278 @@
...
+
+        /// <summary>
+        /// An exception recorded when grouping correlation id but the correlation id is null or not found.
+        /// </summary>
+        public static Metric CorrelationIdNotDefinedException()
+        {
+            return _correlationIdNotDefinedException;
+        }
+
+        public static Metric FhirResourceNotFoundException(ResourceType resourceType)
```
dustinburson, line 222: Don't love this here.  Feel like it should be part of the exception definition and resolved through the interface you added or handled via an ExtensionMethod

src/lib/Microsoft.Health.Fhir.Ingest/Telemetry/ExceptionTelemetryProcessor.cs
```diff
@@ -31,33 +30,71 @@ public ExceptionTelemetryProcessor()
...
 
             return false;
         }
+
+        public static Metric ConvertExceptionToMetric(System.Type lookupType, Type exType)
+        {
+            EnsureArg.IsNotNull(lookupType);
+            EnsureArg.IsNotNull(exType);
+
+            if (lookupType == typeof(MultipleResourceFoundException<>))
```
dustinburson, line 74: This breaks the current implementation.  MultipleResourceFoundException today implements ITelemetryEvent and will be resolved a resource specific event name today, i.e. Observation, Patient, Device, etc.

src/lib/Microsoft.Health.Fhir.Ingest/Telemetry/ExceptionTelemetryProcessor.cs
```diff
@@ -31,33 +30,71 @@ public ExceptionTelemetryProcessor()
...
+            {
+                return IomtMetrics.MultipleResourceFoundException();
+            }
+            else if (lookupType == typeof(PatientDeviceMismatchException))
+            {
+                return IomtMetrics.PatientDeviceMismatchException();
```
dustinburson, line 80: Would prefer that these just implement the interface you added, ITelemetryMetric.  Then you would only need the last one.

src/lib/Microsoft.Health.Fhir.Ingest/Telemetry/ExceptionTelemetryProcessor.cs
```diff
@@ -31,33 +30,71 @@ public ExceptionTelemetryProcessor()
...
+            {
+                return IomtMetrics.TemplateNotFoundException();
+            }
+            else if (lookupType == typeof(CorrelationIdNotDefinedException))
+            {
+                return IomtMetrics.CorrelationIdNotDefinedException();
+            }
+            else
+            {
+                return IomtMetrics.HandledException(exType.Name, ConnectorStage.Unknown);
```
dustinburson, line 96: The stage isn't unknown correct?  The stage is passed in through the HandleException call and should be propagated here.

src/lib/Microsoft.Health.Fhir.Ingest/Telemetry/Metrics/IomtMetrics.cs
```diff
@@ -0,0 +1,278 @@
...
+                { _categoryDimension, Category.Errors },
+                { _errorTypeDimension, ErrorType.DeviceMessageError },
+                { _errorSeverityDimension, ErrorSeverity.Critical },
+                { _stageDimension, ConnectorStage.Normalization },
+           });
+
+        /// <summary>
+        /// The latency between event ingestion and output to FHIR processor.
+        /// </summary>
+        public static Metric MeasurementIngestionLatency()
```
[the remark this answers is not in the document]
wi-y: I was thinking we may wanted to add to the Metric later for PaaS if we wanted to add more dimensions in PaaS that aren't applicable in OSS.

src/lib/Microsoft.Health.Fhir.Ingest/Telemetry/ExceptionTelemetryProcessor.cs
wi-y: Updated

src/func/Microsoft.Health.Fhir.Ingest.Host/Startup.cs
```diff
@@ -0,0 +1,37 @@
...
+{
+    public class Startup : FunctionsStartup
+    {
+        public override void Configure(IFunctionsHostBuilder builder)
+        {
+            EnsureArg.IsNotNull(builder, nameof(builder));
+
+            builder.Services.AddSingleton<ITelemetryLogger>(sp =>
+            {
+                var telemetryConfiguration = new TelemetryConfiguration();
```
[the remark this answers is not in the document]
wi-y: It did not, but have now updated the code so that it will.

src/lib/Microsoft.Health.Fhir.Ingest/Telemetry/ExceptionTelemetryProcessor.cs
wi-y: Ended up removing this by going with the ITelemetryMetric approach

src/lib/Microsoft.Health.Fhir.Ingest/Telemetry/ExceptionTelemetryProcessor.cs
wi-y: Thanks for calling this out. Updated to include the type

src/lib/Microsoft.Health.Fhir.Ingest/Telemetry/Metrics/IomtMetrics.cs
wi-y: Updated

src/lib/Microsoft.Health.Fhir.Ingest/Telemetry/ExceptionTelemetryProcessor.cs
wi-y: Updated

src/lib/Microsoft.Health.Fhir.Ingest/Telemetry/ExceptionTelemetryProcessor.cs
wi-y: Yeah, the stage is actually known. Updated.

src/func/Microsoft.Health.Fhir.Ingest.Host/Startup.cs
```diff
@@ -0,0 +1,62 @@
...
+using System;
+using System.Linq;
+using EnsureThat;
+using Microsoft.ApplicationInsights;
+using Microsoft.ApplicationInsights.Extensibility;
+using Microsoft.Azure.Functions.Extensions.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Health.Fhir.Ingest.Telemetry;
+
+[assembly: FunctionsStartup(typeof(Microsoft.Health.Fhir.Ingest.Service.Startup))]
```
namalu, line 15: nit: You can leave this here, but I have seen other projects manage assembly attributes by creating a separate AssemblyInfo.cs file. I personally think it is cleaner if assembly attributes are curated in a single location.

src/lib/Microsoft.Health.Common/Telemetry/ITelemetryMetric.cs
```diff
@@ -1,12 +1,12 @@
...
-    }
-}
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Common.Telemetry
+{
+    public interface ITelemetryMetric
```
namalu, line 8: nit: I think the naming here is a little misleading. ITelemetryMetric seems like it would be the metric itself, but the functionality it provides is a metric getter. Looking at the implementation it appears that classes that implement this interface can be converted to a metric.

If it makes sense I'm wondering if the interface should be called 'ITelemetryFormattable', and a method 'ToMetric()'.

src/lib/Microsoft.Health.Fhir.Ingest/Telemetry/Metrics/IomtMetrics.cs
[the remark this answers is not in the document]
dustinburson: But the method still returns a Metric object.  It's static so it can't be overridden.  Not following how the extension would be different if this is property instead of a method.  Could you clarify please?

src/func/Microsoft.Health.Fhir.Ingest.Host/Startup.cs
```diff
@@ -22,16 +22,41 @@ public override void Configure(IFunctionsHostBuilder builder)
...
+                }
+
+                var implFactory = configDescriptor.ImplementationFactory;
+
+                builder.Services.Remove(configDescriptor);
+                builder.Services.AddSingleton<ITelemetryLogger>(provider =>
+                {
+                    if (!(implFactory.Invoke(provider) is TelemetryConfiguration config))
+                    {
+                        return null;
```
dustinburson, line 42: What happens in the result of the singleton factory is null? It seems like we will just have no logger then.  It seems like we should either throw an exception or proceed with the default configuration like you do when the instrumentation key is null.

src/func/Microsoft.Health.Fhir.Ingest.Host/Startup.cs
```diff
@@ -0,0 +1,62 @@
...
+            EnsureArg.IsNotNull(builder, nameof(builder));
+
+            var instrumentationKey = Environment.GetEnvironmentVariable("APPINSIGHTS_INSTRUMENTATIONKEY");
+
+            if (instrumentationKey != null)
+            {
+                var configDescriptor = builder.Services.SingleOrDefault(tc => tc.ServiceType == typeof(TelemetryConfiguration));
+                if (configDescriptor?.ImplementationFactory == null)
+                {
+                    return;
```
dustinburson, line 32: I think this might be cleaner if you do something like this:

builder.Services.BuildServiceProvider();
TelemetryConfiguration config = serviceProvider.GetService<TelemetryConfiguration>();

We do something similar in MeasurementFhirImportExtensions.AddMeasurementFhirImport

src/lib/Microsoft.Health.Fhir.Ingest/Telemetry/Metrics/IomtMetrics.cs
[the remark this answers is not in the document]
wi-y: I was hoping to add some dimensions to the Metric object before we log the metric. 

So for example if we had a DeviceEventProcessingLatency metric, we currently have 3 dimensions (name, category, stage), but let's say we want to add a 4th dimension for internal use which isn't applicable to OSS. Right before logging the metric I was hoping to modify the metric object dimensions by setting a 4th dimension key/value if it did not already exist.

I'm not positive on this, but thought that making the metric readonly would prevent us from being able to modify the metric object.

src/func/Microsoft.Health.Fhir.Ingest.Host/Startup.cs
wi-y: Updated to throw exception

src/func/Microsoft.Health.Fhir.Ingest.Host/Startup.cs
wi-y: I attempted to do this, and timeboxed the work at around 2 hours, but wasn't able to get this working by getting the service by type using serviceProvider.GetService<TelemetryConfiguration>() or typeof(TelemetryConfiguration) so am going to leave this as is for now

src/lib/Microsoft.Health.Common/Telemetry/ITelemetryMetric.cs
wi-y: I agree it was a little bit ambiguous/confusing, thank you for the feedback. Updated the interface name and method name

src/func/Microsoft.Health.Fhir.Ingest.Host/Startup.cs
wi-y: updated

src/lib/Microsoft.Health.Fhir.Ingest/Telemetry/ExceptionTelemetryProcessor.cs
wi-y: I attempted to refactor this, but wasn't able to get it to work with DI, and need to spend some more time understanding DI. Do you mind if we do not address this in this PR?

src/func/Microsoft.Health.Fhir.Ingest.Host/Microsoft.Health.Fhir.Ingest.Host.csproj
```diff
@@ -16,12 +16,12 @@
     <FhirVersion Condition="'$(FhirVersion)' == ''">R4</FhirVersion>
   </PropertyGroup>  
   <ItemGroup>
-    <PackageReference Include="Microsoft.ApplicationInsights" Version="2.14.0" />
     <PackageReference Include="Microsoft.Azure.EventHubs.Processor" Version="4.1.0" />
     <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.14" />
     <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.EventHubs" Version="3.0.6" />
     <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.Http" Version="3.0.2" />
     <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.Storage" Version="3.0.8" />
+    <PackageReference Include="Microsoft.Azure.WebJobs.Logging.ApplicationInsights" Version="3.0.14" />
```
dustinburson, line 24: Curious, what is the different between these two packages?

src/func/Microsoft.Health.Fhir.Ingest.Host/Startup.cs
```diff
@@ -0,0 +1,60 @@
...
+    {
+        public override void Configure(IFunctionsHostBuilder builder)
+        {
+            EnsureArg.IsNotNull(builder, nameof(builder));
+
+            var instrumentationKey = Environment.GetEnvironmentVariable("APPINSIGHTS_INSTRUMENTATIONKEY");
+
+            if (instrumentationKey != null)
+            {
+                var configDescriptor = builder.Services.SingleOrDefault(tc => tc.ServiceType == typeof(TelemetryConfiguration));
```
dustinburson, line 27: Would you be able to detail the issue you were hitting with prior suggested line of code?  I am find continuing as is but I think I might want to take a look.  Thanks!

src/func/Microsoft.Health.Fhir.Ingest.Host/Microsoft.Health.Fhir.Ingest.Host.csproj
wi-y: My understanding based on reading the code in the [webjobs git repo](https://github.com/Azure/azure-webjobs-sdk/tree/dev/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights) is that the package Microsoft.Azure.WebJobs.Logging.ApplicationInsights is specific to Azure Functions and basically an extension of Microsoft.ApplicationInsights. The Microsoft.Azure.WebJobs.Logging.ApplicationInsights package contains a few additional things such as getting the Role, Host, etc of the function app (which we don't really need) but also includes some of the things we do need such as the adaptive sampling.

src/func/Microsoft.Health.Fhir.Ingest.Host/Startup.cs
wi-y: I am thinking that the serviceProvider doesn't get built correctly or I just don't understand what it is doing. 

I hit this error: System.InvalidOperationException: 'Unable to resolve service for type 'Microsoft.Azure.WebJobs.Script.IEnvironment' while attempting to activate 'Microsoft.Azure.WebJobs.Script.Configuration.ApplicationInsightsLoggerOptionsSetup'.'

with:

var serviceProvider = builder.Services.BuildServiceProvider();
var tconfig = serviceProvider.GetService(typeof(TelemetryConfiguration));
 
 or

var serviceProvider = builder.Services.BuildServiceProvider();
var tconfig = serviceProvider.GetService<TelemetryConfiguration>();

PR #62, opened 2020-08-31
Use milliseconds for metrics, support prefix for metric name, rename stage to operation

wi-y:


src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementEventNormalizationService.cs
```diff
@@ -80,7 +80,7 @@ private async Task StartConsumer(ISourceBlock<EventData> producer, IAsyncCollect
                     {
                         _log.LogMetric(
                             IomtMetrics.DeviceEventProcessingLatency(),
```
namalu, line 82: Do we need to worry about backwards compatibility here? If someone updates an existing IoMT Connector and they are monitoring latency, this will cause an issue?

Should we just create a new metric? DeviceEventProcessingLatencyMs?

wi-y: Creating a new metric would definitely be the safest option from a maintaining backwards compatibility standpoint.

I am worried that creating a new metric would create roughly 2 times as many metrics as before, though. We could also multiple the current metric value by 1000 (internally) for these latency metrics, although that isn't a great solution.

All things considered, creating a new metric is probably the best approach, so will update with that.

wi-y: Updated

dustinburson (approved):

PR #70, opened 2020-09-22
Create a new template factory for IoT Central

rogordon01:
IoT Central has an updated [telemetry format](https://github.com/microsoft/iomt-fhir/pull/new/personal/rogordon01/addIotCentralTemplateFactory) for exporting device data. This PR creates a new template factory in order to normalize events of this type.

### DIFF
--- a/test/Microsoft.Health.Fhir.Ingest.Template.UnitTests/Microsoft.Health.Fhir.Ingest.Template.UnitTests.csproj
+++ b/test/Microsoft.Health.Fhir.Ingest.Template.UnitTests/Microsoft.Health.Fhir.Ingest.Template.UnitTests.csproj
@@ -43,106 +43,10 @@
   </ItemGroup>
 
   <ItemGroup>
-    <None Update="TestInput\data_CodeValueFhirTemplate_CodeableConceptAndStringComponent.json">
+    <None Update="TestInput\*.json">
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
     </None>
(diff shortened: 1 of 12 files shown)


test/Microsoft.Health.Fhir.Ingest.Template.UnitTests/Microsoft.Health.Fhir.Ingest.Template.UnitTests.csproj
```diff
@@ -43,106 +43,10 @@
   </ItemGroup>
 
   <ItemGroup>
-    <None Update="TestInput\data_CodeValueFhirTemplate_CodeableConceptAndStringComponent.json">
+    <None Update="TestInput\*.json">
```
rogordon01, line 46: Instead of hardcoding each file I used a regex. Please let me know if this isn't desired.

dustinburson (approved):
Changes look good.  Only suggestion on the tests is maybe add one that has you pulling properties from both the telemetry section and the enrichments.  It should just fine but be good to have an example in the tests.

We will also need to update our documentation in Configuration.md but that can be handled in a separate PR if you desire.

rogordon01:
> Changes look good. Only suggestion on the tests is maybe add one that has you pulling properties from both the telemetry section and the enrichments. It should just fine but be good to have an example in the tests.
> 
> We will also need to update our documentation in Configuration.md but that can be handled in a separate PR if you desire.

Will do. I'll add that in the PR where I update the documentation

PR #73, opened 2020-10-08
add timed based interval flushing

wi-y:
This is a very early initial proof of concept for flushing events within a given time interval. These changes are just one possibility and wanted to get feedback on this approach - mostly wanted to talk through this at a high level. 

I also plan on addressing:
- multiple partitions
- watermarking/checkpointing
- KEDA
- using the Azure Function Event Hub extension vs. Azure Event Hub SDK

### DIFF
--- a/src/lib/Microsoft.Health.Common/EventHub/Listeners/EventHubListener.cs
+++ b/src/lib/Microsoft.Health.Common/EventHub/Listeners/EventHubListener.cs
@@ -123,11 +123,14 @@ namespace Microsoft.Health.Common.EventHubs
             private readonly ICheckpointer _checkpointer;
             private readonly int _batchCheckpointFrequency;
             private readonly int _eventQueueSize;
-            private readonly Task _timedFlush;
             private readonly TimeSpan _flushTimeSpan;
             private readonly ConcurrentDictionary<string, PartitionEventData> _pendingEventData;
             private int _batchCounter = 0;
             private bool _disposed = false;
+            private DateTime _firstEnqueuedDateTime;
+            private DateTime _currentIntervalStart;
+            private DateTime _currentIntervalEnd;
+            private readonly Task _timeIntervalFlush;
 
             public EventProcessor(EventHubOptions options, ITriggeredFunctionExecutor executor, ILogger logger, bool singleDispatch, ICheckpointer checkpointer = null)
             {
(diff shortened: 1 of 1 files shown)


src/lib/Microsoft.Health.Common/EventHub/Listeners/EventHubListener.cs
```diff
@@ -230,6 +236,129 @@ private async Task TimedFlush()
...
+            private async void TimeIntervalFlush()
+            {
+                while (!_cts.IsCancellationRequested)
+                {
+                    if (_currentIntervalEnd == DateTime.MinValue)
+                    {
+                        _currentIntervalEnd = CalculateIntervalEnd(_currentIntervalStart, _flushTimeSpan);
+                    }
+
+                    if (_currentIntervalEnd == DateTime.MinValue)
```
dustinburson, line 248: Better to use a a reset event or semaphore to coordinate rather than a task delay.

src/lib/Microsoft.Health.Common/EventHub/Listeners/EventHubListener.cs
```diff
@@ -230,6 +236,129 @@ private async Task TimedFlush()
...
+                    {
+                        continue;
+                    }
+
+                    var batch = new List<EventData>();
+                    while (pendingData.Data.TryPeek(out var eventData))
+                    {
+                        eventData.SystemProperties.TryGetValue("x-opt-enqueued-time", out var enqueuedTime);
+                        var enqueuedDateTime = Convert.ToDateTime(enqueuedTime);
+                        eventData.Dispose();
```
dustinburson, line 330: Can't dispose, event is still in the queue.

src/lib/Microsoft.Health.Common/EventHub/Listeners/EventHubListener.cs
```diff
@@ -230,6 +236,129 @@ private async Task TimedFlush()
...
+                {
+                    if (pendingData.Data.IsEmpty)
+                    {
+                        continue;
+                    }
+
+                    var batch = new List<EventData>();
+                    while (pendingData.Data.TryPeek(out var eventData))
+                    {
+                        eventData.SystemProperties.TryGetValue("x-opt-enqueued-time", out var enqueuedTime);
```
dustinburson, line 328: You don't really want to enforce the enqueue time here.  We should be checking before we even buffer the value.

dustinburson: Specifically in the ProcessEventsAsync(PartitionContext context, IEnumerable<EventData> messages) method.

PR #78, opened 2020-11-21
Console app prototype for stream analytics replacement

wi-y:
This is the initial prototype for an oss stream analytics replacement console app.

Before running the app, you will also need fill out appsettings.json with the appropriate event hub and storage account information, and well as instruct the app to read from the appropriate event hub (determined by Console:EventHub). 

To run the app, navigate to the console folder of the project and set Microsoft.Health.Fhir.Ingest.Console as the startup project, and then build/run. Some logging info will appear in the console for number of events processed for a given window.

You can adjust the window in appsettings.json via EventBatching:FlushTimespan and EventBatching:MaxEvents.

There are some todos remaining which include:
- removing hard-coded template files
- logging
- managed identity (with event hubs, storage accounts, fhir server)
- testing min and max batch window interval sizes and queue sizes


src/console/MeasurementCollectionToFhir/Processor.cs
```diff
@@ -0,0 +1,60 @@
+﻿// -------------------------------------------------------------------------------------------------
```
wi-y, line 1: instead of writing these Processor/ProcessorStartup classes, it may be easier to take the existing Functions code in IomtConnectorFunctions and then just remove the EventHub triggers for our use case here. That way we can reuse as much code as possible and would probably not need these separate ProcessorStartup DI classes.

The reason I didn't do this originally is that I didn't want to take a dependency on Webjobs. But later I found that keeping webjobs would allow us to keep the blob triggers which are pretty useful and also allow us to reuse the existing oss code in a more straightforward way, so I will be adding the webjobs dependency back and adding the blob triggers back (and will remove the hard coded mapping templates)

src/console/devicecontent.json
```diff
@@ -0,0 +1,115 @@
+{
```
wi-y, line 1: will remove these mapping files once I add the blob triggers back in.

src/lib/Microsoft.Health.Events/EventConsumers/Service/Infrastructure/EventQueueWindow.cs
```diff
@@ -0,0 +1,34 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+
+namespace Microsoft.Health.Events.EventConsumers.Service.Infrastructure
+{
+    public class EventQueueWindow
```
dustinburson, line 10: Not sure of the value of this class.  Seems like this can be easily tracked on the EventQueue class.  You only ever have one instance of this per queue.

src/lib/Microsoft.Health.Events/EventConsumers/Service/Infrastructure/EventQueueWindow.cs
```diff
@@ -0,0 +1,34 @@
...
+        {
+            _windowEnd = initDateTime.Add(flushTimespan);
+            _flushTimespan = flushTimespan;
+        }
+
+        public void IncrementWindow(DateTime currentEnqueudTime)
+        {
+            while (currentEnqueudTime >= _windowEnd)
+            {
+                _windowEnd = _windowEnd.Add(_flushTimespan);
```
dustinburson, line 25: You should be able to do this with math.  The difference between the times will result in a timespan.  You should be able to determine how many times bigger the time is from the window end.  You can then scale the window end by X amount of the timespan.

src/lib/Microsoft.Health.Events/EventConsumers/Service/Infrastructure/EventQueue.cs
```diff
@@ -0,0 +1,92 @@
...
+
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Microsoft.Health.Events.Model;
+
+namespace Microsoft.Health.Events.EventConsumers.Service.Infrastructure
+{
+    public class EventQueue
```
dustinburson, line 14: nit:EventPartition or something similar seems more appropriate since this is encapsulating a queue and related partition data.

src/lib/Microsoft.Health.Events/EventConsumers/Service/EventBatchingService.cs
```diff
@@ -0,0 +1,139 @@
...
+            UpdateCheckpoint(events);
+        }
+
+        private async void ThresholdWaitReached(string queueId, DateTime windowEnd)
+        {
+            if (windowEnd < DateTime.UtcNow.AddSeconds(_timeBuffer))
+            {
+                Console.WriteLine($"Threshold wait reached. Flushing {_eventQueues[queueId].GetQueueCount()} events up to: {windowEnd}");
+                var events = await GetQueue(queueId).Flush(windowEnd);
+                await _eventConsumerService.ConsumeEvents(events);
```
dustinburson, line 120: We need to increment the window time here as well correct?  If we don't any new event that arrives will immediately trigger a new batch.

dustinburson (changes requested):

src/lib/Microsoft.Health.Events/EventConsumers/IEventConsumer.cs
```diff
@@ -0,0 +1,17 @@
...
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Health.Events.Model;
+
+namespace Microsoft.Health.Events.EventConsumers
+{
+    public interface IEventConsumer
+    {
+        Task<IActionResult> ConsumeAsync(IEnumerable<Event> events);
```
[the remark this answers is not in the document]
wi-y: It isn't  really needed. Will remove and just use void

src/lib/Microsoft.Health.Events/EventConsumers/Service/EventBatchingService.cs
```diff
@@ -0,0 +1,139 @@
...
+            await _eventConsumerService.ConsumeEvents(events);
+            queue.IncrementQueueWindow(eventArg.EnqueuedTime.UtcDateTime);
+            UpdateCheckpoint(events);
+        }
+
+        private async void ThresholdWaitReached(string queueId, DateTime windowEnd)
+        {
+            if (windowEnd < DateTime.UtcNow.AddSeconds(_timeBuffer))
+            {
+                Console.WriteLine($"Threshold wait reached. Flushing {_eventQueues[queueId].GetQueueCount()} events up to: {windowEnd}");
```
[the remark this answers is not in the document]
wi-y: Yep makes sense.

src/lib/Microsoft.Health.Events/EventConsumers/Service/EventBatchingService.cs
```diff
@@ -0,0 +1,139 @@
...
+using Microsoft.Health.Events.EventConsumers.Service.Infrastructure;
+using Microsoft.Health.Events.Model;
+
+namespace Microsoft.Health.Events.EventConsumers.Service
+{
+    public class EventBatchingService : IEventConsumerService
+    {
+        private ConcurrentDictionary<string, EventQueue> _eventQueues;
+        private int _maxEvents = 100;
+        private TimeSpan _flushTimespan = TimeSpan.FromSeconds(300);
```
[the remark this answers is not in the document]
wi-y: These were some default values in the case where the EventBatchingOptions weren't set. I will remove them and just check that the options were set.

src/lib/Microsoft.Health.Events/EventConsumers/Service/EventBatchingService.cs
wi-y: I don't think we need to increment here. This method is only called when we haven't received an event in a while, and will only flush if the window end is sufficiently in the past.

If we get a new event then it has to be outside the current window end, we will flush the previous window which has 0 events and advance the window to contain this new event.

The main reason why I didn't want to increment here is to avoid a race condition with the other flushing mechanisms (ThresholdTimeReached and ThresholdCountReached). Perhaps this should be better designed though

src/lib/Microsoft.Health.Events/Model/Event.cs
```diff
@@ -0,0 +1,47 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Health.Events.Model
+{
+    public class Event
```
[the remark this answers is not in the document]
wi-y: Yeah the main reason for creating this "Event" object was not to be tied to event hub dependencies elsewhere in our code.

It isn't really an improvement (other than removing the dependency) since we are just using a similar object instead of the event hub one. I will try and improve this via an interface or something else

src/lib/Microsoft.Health.Events/EventHubProcessor/EventProcessor.cs
```diff
@@ -0,0 +1,113 @@
...
+            }
+
+            Task ProcessErrorHandler(ProcessErrorEventArgs eventArgs)
+            {
+                // todo: add an error processor
+                Console.WriteLine(eventArgs.Exception.Message);
+                return Task.CompletedTask;
+            }
+
+            async Task ProcessInitializingHandler(PartitionInitializingEventArgs initArgs)
```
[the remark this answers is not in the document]
wi-y: It is partly due to not passing eventArgs around.

But I also did not like eventArgs.UpdateCheckpointAsync() because we don't really want to checkpoint at the end of each flush.

So if we are not checkpointing at the end of each flush, we would need some way of keeping track of some previous eventArgs objects and then execute eventArgs.UpdateCheckpointAsync() when we feel it is appropriate to checkpoint. 

The other piece is using eventArgs.UpdateCheckpointAsync is that we don't have control over what keys/values get added to the checkpoint. Right now it adds the [offset and sequence number](https://github.com/Azure/azure-sdk-for-net/blob/82d737c996d086504d3ce78a604b0dfefd134273/sdk/eventhub/Azure.Messaging.EventHubs.Shared/src/BlobCheckpointStore/BlobsCheckpointStore.cs#L369) but not an enqueuedDateTime which we also thought would be helpful for reporting the status/progress.

src/lib/Microsoft.Health.Events/EventCheckpointing/StorageCheckpointClient.cs
```diff
@@ -0,0 +1,183 @@
...
+
+        private void SetPublisherTimer()
+        {
+            _publisherTimer = new System.Timers.Timer(_publishTimerInterval);
+            _publisherTimer.Elapsed += OnTimedEvent;
+            _publisherTimer.AutoReset = true;
+            _publisherTimer.Enabled = true;
+        }
+
+        private async void OnTimedEvent(object source, ElapsedEventArgs e)
```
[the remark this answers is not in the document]
wi-y: I was thinking after each batch / successful flush we should update the checkpoint internally in memory, and then periodically (on a timer or count) publish the checkpoints that are in memory to the blob container.

It looks like I commented out the publish piece by mistake while looking into something else and forgot to uncomment.

src/console/devicecontent.json
wi-y: removed

src/lib/Microsoft.Health.Events/EventConsumers/Service/EventConsumerService.cs
```diff
@@ -0,0 +1,37 @@
...
+        public Task ConsumeEvent(Event eventArg)
+        {
+            throw new System.NotImplementedException();
+        }
+
+        public Task ConsumeEvents(IEnumerable<Event> events)
+        {
+            foreach (IEventConsumer eventConsumer in eventConsumers)
+            {
+                eventConsumer.ConsumeAsync(events);
```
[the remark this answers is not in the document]
wi-y: It wasn't intentional. Thanks for calling this out. Updated with retries / exponential backoff. Will add some tests later

src/lib/Microsoft.Health.Events/Storage/StorageOptions.cs
```diff
@@ -0,0 +1,18 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Events.Storage
```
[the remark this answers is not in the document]
wi-y: updated

src/lib/Microsoft.Health.Events/Model/Event.cs
```diff
@@ -0,0 +1,47 @@
...
+
+        public ReadOnlyMemory<byte> Body { get; }
+
+        public long SequenceNumber { get; }
+
+        public long Offset { get; }
+
+        public DateTimeOffset EnqueuedTime { get; }
+
+        public Dictionary<string, object> SystemProperties { get; }
```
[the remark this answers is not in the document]
wi-y: yeah, that makes sense. updated

src/lib/Microsoft.Health.Events/Model/Event.cs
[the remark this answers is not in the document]
wi-y: updated to EventMessage. Not a huge fan but couldn't think of a better name that wasn't already taken

src/lib/Microsoft.Health.Events/EventHubProcessor/EventProcessor.cs
```diff
@@ -0,0 +1,113 @@
...
+                            eventArgs.Data.SequenceNumber,
+                            eventArgs.Data.EnqueuedTime.UtcDateTime,
+                            eventArgs.Data.SystemProperties);
+
+                        _eventConsumerService.ConsumeEvent(evt);
+                    }
+                    else
+                    {
+                        var evt = new MaximumWaitEvent(eventArgs.Partition.PartitionId, DateTime.UtcNow);
+                        _eventConsumerService.ConsumeEvent(evt);
```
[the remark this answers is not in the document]
wi-y: thanks. updated

src/lib/Microsoft.Health.Events/EventHubProcessor/EventProcessor.cs
```diff
@@ -0,0 +1,113 @@
...
+
+        public async Task RunAsync(EventProcessorClient processor, CancellationToken ct)
+        {
+            EnsureArg.IsNotNull(processor);
+
+            Task ProcessEventHandler(ProcessEventArgs eventArgs)
+            {
+                try
+                {
+                    if (eventArgs.HasEvent)
```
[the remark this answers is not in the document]
wi-y: added a comment

src/console/MeasurementCollectionToFhir/Processor.cs
wi-y: I ended up not doing this due to time, and not sure it is worth while given that the blob inputs aren't triggers. Ended up just fetching the blob items from storage each event batch to replicate current functions behavior.

src/lib/Microsoft.Health.Events/EventConsumers/Service/EventBatchingService.cs
dustinburson: Is there a race condition they should be processed on the same thread? We had a race condition in my version since we had a timer triggering the processing.

src/lib/Microsoft.Health.Events/EventCheckpointing/StorageCheckpointClient.cs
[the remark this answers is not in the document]
dustinburson: I am still not sure we need a timer.  We can update in memory, that is fine but I don't think saving on a timer add much and introduces complexity.  Would prefer saving every X calls where X could be 1 (we save every time) or could be value greater to reduce the number of times we save.  This is assuming we keep a custom checkpoint story.

---
In reply to: [535658990](https://github.com/microsoft/iomt-fhir/pull/78#discussion_r535658990) [](ancestors = 535658990)

src/lib/Microsoft.Health.Events/EventHubProcessor/EventProcessor.cs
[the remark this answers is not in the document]
dustinburson: Thank you for the explanation.

Personally I think keeping the last eventArgs you processed is fine.  It seems like a small cost to pay .  This is information we have to track in one way or the other.  The point about status/tracking process is a good one but I don't think we need to checkpoint .  This really should be a metric and we shouldn't need to rely on it being saved as a datetime for watermark purposes in order to get our metrics.  And we should be able to do with the event time to metric.

Would be interested in what others think here.

---
In reply to: [535501756](https://github.com/microsoft/iomt-fhir/pull/78#discussion_r535501756) [](ancestors = 535501756)

src/lib/Microsoft.Health.Logger/Telemetry/ITelemetryLogger.cs
```diff
@@ -3,10 +3,10 @@
 // Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // -------------------------------------------------------------------------------------------------
 
-using System;
 using Microsoft.Health.Common.Telemetry;
+using System;
 
-namespace Microsoft.Health.Fhir.Ingest.Telemetry
+namespace Microsoft.Health.Logger.Telemetry
```
dustinburson, line 9: >Logger [](start = 27, length = 6)

Nit: I believe the common pattern here is Logging

src/console/Normalize/ProcessorStartup.cs
```diff
@@ -0,0 +1,36 @@
...
+        {
+            Configuration = configuration;
+        }
+
+        public IConfiguration Configuration { get; }
+
+        public void ConfigureServices(IServiceCollection services)
+        {
+            var outputEventHubConnection = Configuration.GetSection("OutputEventHub").Value;
+            var outputEventHubName = outputEventHubConnection.Substring(outputEventHubConnection.LastIndexOf('=') + 1);
```
dustinburson, line 25: >var outputEventHubName = outputEventHubConnection.Substring(outputEventHubConnection.LastIndexOf('=') + 1); [](start = 12, length = 107)

Should be able to use the event hub connection string builder here.

dustinburson (commented):
@wi-y I re-reviewed.  I think you addressed most of the must fix items.  I would like to get at least one other person's review here though and get another point of view.  It also looks like the build isn't triggering for this PR.  Is that something you could look into?  

Overall I think we may want to iterate on this some more but rather than do it here I would like to get into master and do follow up with some additional PRs.  Make sense?

wi-y:
> @wi-y I re-reviewed. I think you addressed most of the must fix items. I would like to get at least one other person's review here though and get another point of view. It also looks like the build isn't triggering for this PR. Is that something you could look into?
> 
> Overall I think we may want to iterate on this some more but rather than do it here I would like to get into master and do follow up with some additional PRs. Make sense?

@dustinburson thank you for reviewing. I agree with getting this into master and iterating. I'll look into the builds not triggering.

src/console/MeasurementCollectionToFhir/Processor.cs
```diff
@@ -0,0 +1,56 @@
...
+            try
+            {
+                EnsureArg.IsNotNull(_templateDefinition);
+                var templateContent = _templateManager.GetTemplateAsString(_templateDefinition);
+
+                await _measurementImportService.ProcessEventsAsync(events, templateContent, _logger).ConfigureAwait(false);
+            }
+            catch
+            {
+                throw;
```
rogordon01, line 52: Do we need this `try/catch` block if all we do is re-throw the exception?

src/lib/Microsoft.Health.Events/EventCheckpointing/StorageCheckpointClient.cs
```diff
@@ -0,0 +1,240 @@
...
+            {
+                try
+                {
+                    await blobClient.SetMetadataAsync(metadata);
+                }
+                catch (RequestFailedException ex) when ((ex.ErrorCode == BlobErrorCode.BlobNotFound) || (ex.ErrorCode == BlobErrorCode.ContainerNotFound))
+                {
+                    using (var blobContent = new MemoryStream(Array.Empty<byte>()))
+                    {
+                        await blobClient.UploadAsync(blobContent, metadata: metadata).ConfigureAwait(false);
```
rogordon01, line 70: Will this command also create the container if its not present?

src/lib/Microsoft.Health.Events/EventHubProcessor/EventProcessor.cs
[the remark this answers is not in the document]
rogordon01: Simplifying the codebase by using the built in checkpointing sounds like a resonable suggestion. If we find that the extra values being checkpointed (i..e enqueuedDateTime) don't provide value then we could explore a refactor.

src/lib/Microsoft.Health.Events/Repository/StorageManager.cs
```diff
@@ -0,0 +1,38 @@
...
+    public class StorageManager : IRepositoryManager
+    {
+        private BlobContainerClient _blobContainer;
+
+        public StorageManager(string connectionString, string blobContainerName)
+        {
+            EnsureArg.IsNotNull(connectionString);
+            EnsureArg.IsNotNull(blobContainerName);
+
+            _blobContainer = new BlobContainerClient(connectionString, blobContainerName);
```
rogordon01, line 21: We use `BlobContainerClient` in several places. Can we pass in a shared instance via the constructor versus creating new ones? Would also help with testing as we could simply pass in a mocked instance. And your validation logic for the params could be centralized in one place.

src/lib/Microsoft.Health.Events/EventCheckpointing/StorageCheckpointClient.cs
wi-y: It will not create a container. I was thinking we should validate that the container exists on startup or something like that, like you suggested in another comment. 

We will have an arm template create the container, so in theory the container should exist unless someone goes and deletes it.

src/lib/Microsoft.Health.Events/Repository/StorageManager.cs
wi-y: It is a good idea. One issue is that I believe BlobContainerClient is specific to just one blob BlobContainer. Currently we have one blob container for templates called "template" and I didn't want to put my checkpoints in there. So for checkpoints I created another container and concluded that I needed another BlobContainerClient. I was really wanting to consolidate but I think there may be limitations. I will take another looks to see what can be done in case I missed something.

src/lib/Microsoft.Health.Events/EventCheckpointing/StorageCheckpointClient.cs
```diff
@@ -0,0 +1,240 @@
...
+                    }
+                }
+            }
+            catch (RequestFailedException ex) when (ex.ErrorCode == BlobErrorCode.ContainerNotFound)
```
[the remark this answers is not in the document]
wi-y: updated, not sure why I did that.

src/lib/Microsoft.Health.Events/EventCheckpointing/StorageCheckpointClient.cs
```diff
@@ -0,0 +1,240 @@
...
+            }
+
+            try
+            {
+                // todo: consider retries
+                return GetCheckpointsAsync();
+            }
+            catch (RequestFailedException ex) when (ex.ErrorCode == BlobErrorCode.ContainerNotFound)
+            {
+                // todo: log errors
```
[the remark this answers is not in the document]
wi-y: updated

src/console/MeasurementCollectionToFhir/Processor.cs
wi-y: updated

src/lib/Microsoft.Health.Events/Model/EventMessage.cs
```diff
@@ -0,0 +1,51 @@
...
+
+        public EventMessage(
+            string partitionId,
+            ReadOnlyMemory<byte> body,
+            long sequenceNumber,
+            long offset,
+            DateTimeOffset enqueuedTime,
+            IDictionary<string, object> properties,
+            IReadOnlyDictionary<string, object> systemProperties)
+        {
```
namalu, line 27: I'm wondering if it makes sense to do null checks here. It seems like some of these properties are "required", so enforcing this on the initializer might make sense. That or maybe in the factory.

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementFhirImportService.cs
```diff
@@ -87,6 +108,31 @@ private static async Task<IEnumerable<IMeasurementGroup>> ParseAsync(Stream data
             return measurementGroups;
         }
 
+        private static IEnumerable<IMeasurementGroup> ParseEventData(IEnumerable<IEventMessage> data)
+        {
+            // Deserialize events into measurements and then group according to the device, type, and other factors
+            var body = data.First().Body.ToArray();
+            var text = System.Text.Encoding.Default.GetString(body);
+            var measurement = JsonConvert.DeserializeObject<Measurement>(text);
```
namalu, line 116: Maybe I'm missing something, but it looks like measurement isn't used? what is the purpose of deserializing the first message?

wi-y: It should be removed. I think I may have been logging the event for debugging but didn't remove these lines when I was done.

src/console/Normalize/ProcessorStartup.cs
dustinburson: Yes, once this is in I am planning on making stories for supporting MI.  There are few things I would like to iterate on overall in this code but I would like to this initial build in as a base committed to master so we have the history.

---
In reply to: [544605846](https://github.com/microsoft/iomt-fhir/pull/78#discussion_r544605846) [](ancestors = 544605846)

dustinburson (approved):
:shipit:

PR #80, opened 2020-12-03
Updating PatientID usage depending on resolution mode

msjasteppe:


docs/Configuration.md
```diff
@@ -59,7 +59,7 @@ The JsonPathContentTemplate allows matching on and extracting values from an Eve
 |**TypeMatchExpression**|The JSON Path expression that is evaluated against the EventData payload. If a matching JToken is found the template is considered a match. All subsequent expressions are evaluated against the extracted JToken matched here.|`$..[?(@heartRate)]`
 |**TimestampExpression**|The JSON Path expression to extract the timestamp value for the measurement's OccurenceTimeUtc.|`$.endDate`
 |**DeviceIdExpression**|The JSON Path expression to extract the device identifier.|`$.deviceId`
-|**PatientIdExpression**|*Optional*: The JSON Path expression to extract the patient identifier.|`$.patientId`
+|**PatientIdExpression**|*Required* when IdentityResolution is in **Create** mode and *Optional* when IdentityResolution is in **Lookup** mode. The JSON Path expression to extract the patient identifier.|`$.patientId`
```
ms-teli, line 62: Nit: would that be possible to put the link to where we explain the difference between "Create" and "Lookup", it should be convenient for customer to jump to right place to learn or review. It took me a while to understand the difference and prerequisites for "Lookup" in navigating the pages so I thought it would be also helpful to the customer as well.

ms-teli (approved):

ms-teli (approved):

docs/Configuration.md
```diff
@@ -57,9 +57,9 @@ The JsonPathContentTemplate allows matching on and extracting values from an Eve
 | --- | --- | ---
 |**TypeName**|The type to associate with measurements that match the template.|`heartrate`
 |**TypeMatchExpression**|The JSON Path expression that is evaluated against the EventData payload. If a matching JToken is found the template is considered a match. All subsequent expressions are evaluated against the extracted JToken matched here.|`$..[?(@heartRate)]`
-|**TimestampExpression**|The JSON Path expression to extract the timestamp value for the measurement's OccurenceTimeUtc.|`$.endDate`
+|**TimestampExpression**|The JSON Path expression to extract the timestamp value for the measurement's OccurrenceTimeUtc.|`$.endDate`
 |**DeviceIdExpression**|The JSON Path expression to extract the device identifier.|`$.deviceId`
-|**PatientIdExpression**|*Optional*: The JSON Path expression to extract the patient identifier.|`$.patientId`
+|**PatientIdExpression**|*Required* when IdentityResolution is in [Create](ARMInstallation.md#Resource-Identity-Resolution-Type) mode and *Optional* when IdentityResolution is in [Lookup](ARMInstallation.md#Resource-Identity-Resolution-Type) mode. The JSON Path expression to extract the patient identifier.|`$.patientId`
```
dustinburson, line 62: Not sure how best to update the documentation but the PatientIdExpression result won't be used at all when the IdentityResolutionType is set to lookup. You can technically map it still but it won't be used.

dustinburson (approved):

PR #83, opened 2020-12-16
Personal/wiyochum/arm deployment event batching

wi-y:
This PR contains a script and an ARM template for deploying the IoMT Connector as Azure WebJobs.

Will point this PR to master, or merge this PR into my existing SA replacement branch depending on sign off timing.

### DIFF
--- a/deploy/scripts/Create-IomtWebJobsEnvironment.ps1
+++ b/deploy/scripts/Create-IomtWebJobsEnvironment.ps1
@@ -1,0 +1,109 @@
+<#
+.SYNOPSIS
+Creates a new IoMT FHIR Connector for Azure without using Stream Analytics
+.DESCRIPTION
+#>
+param
+(
+    [Parameter(Mandatory = $true)]
+    [ValidateNotNullOrEmpty()]
+    [ValidateLength(5,12)]
+	[ValidateScript({
+        if ("$_" -cmatch "(^([a-z]|\d)+$)") {
+            return $true
+        }
+        else {
+			throw "Environment name must be lowercase and numbers"
+            return $false
+        }
+    })]
+    [string]$EnvironmentName,
+
+    [Parameter(Mandatory = $false)]
+    [ValidateSet('Australia East','East US','East US 2', 'West US', 'West US 2','North Central US','South Central US','Southeast Asia','North Europe','West Europe','UK West','UK South')]
+    [string]$EnvironmentLocation = "North Central US",
+
+    [Parameter(Mandatory = $false)]
+    [ValidateSet('R4')]
+    [string]$FhirVersion = "R4",
+
+    [Parameter(Mandatory = $false)]
+    [string]$SourceRepository = "https://github.com/microsoft/iomt-fhir",
+  
+    [Parameter(Mandatory = $false)]
+    [string]$SourceRevision = "master",
+
+    [Parameter(Mandatory = $true)]
+    [string]$FhirServiceUrl,
+
+    [Parameter(Mandatory = $true)]
+    [string]$FhirServiceAudience,
+
+    [Parameter(Mandatory = $false)]
+    [string]$EnvironmentDeploy = $true
+)
+
+Function BuildPackage() {
+    try {
+        Push-Location $currentPath
+        cd ../../src/console/
+        dotnet restore
+        dotnet build --output $buildPath /p:DeployOnBuild=true /p:DeployTarget=Package
+    } catch {
+        throw
+    } finally {
+        Pop-Location
+    }
+}
+
+Function Deploy-WebJobs($DeviceDataWebJobName, $NormalizedDataWebJobName) {
+    try {
+        Clean-Path -WebJobName $DeviceDataWebJobName
+        Clean-Path -WebJobName $NormalizedDataWebJobName
+
+        $resourceGroupName = $EnvironmentName
+        $webAppName = $EnvironmentName
+        $webJobType = "Continuous"
+        $buildPath = "$currentPath\OSS_Deployment"
+        $tempPath = "$currentPath\Temp"
+
+        $DeviceWebJobPath = "$tempPath\App_Data\jobs\$webJobType\$DeviceDataWebJobName"
+        $NormalizedWebJobPath = "$tempPath\App_Data\jobs\$webJobType\$NormalizedDataWebJobName"
+        Copy-Item "$buildPath\*" -Destination $DeviceWebJobPath -Recurse
+        Copy-Item "$buildPath\*" -Destination $NormalizedWebJobPath -Recurse
+
+        Compress-Archive -Path "$tempPath\*" -DestinationPath "$currentPath\iomtwebjobs.zip" -Force
+
+        Publish-AzWebApp -ArchivePath "$currentPath\iomtwebjobs.zip" -ResourceGroupName $resourceGroupName -Name $webAppName
+    } catch {
+        throw
+    } finally {
+        Pop-Location
+    }
+}
+
+Function Clean-Path($WebJobName) {
+    $WebJobPath = "$tempPath\App_Data\jobs\$webJobType\$WebJobName"
+    Get-ChildItem -Path $WebJobPath -Recurse | Remove-Item -Force -Recurse
+    if( -Not (Test-Path -Path $WebJobPath ) )
+    {
+        New-Item $WebJobPath -ItemType Directory
+    }
+}
+
+Set-StrictMode -Version Latest
+
+# deploy event hubs, app service, key vaults, storage
+if ($EnvironmentDeploy -eq $true) {
+    Write-Host "Deploying environment resources..."
+    $webjobTemplate = "..\templates\default-azuredeploy-webjobs.json"
+    New-AzResourceGroupDeployment -TemplateFile $webjobTemplate -ResourceGroupName $EnvironmentName -ServiceName $EnvironmentName -FhirServiceUrl $fhirServerUrl -FhirServiceAudience $fhirServerUrl -RepositoryUrl $SourceRepository -RepositoryBranch $SourceRevision -ResourceLocation $EnvironmentLocation
+}
+
+# deploy the stream analytics replacement webjobs
+Write-Host "Deploying WebJobs..."
+
+$currentPath = (Get-Location).Path
+BuildPackage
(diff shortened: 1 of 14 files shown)


deploy/templates/default-azuredeploy-webjobs.json
```diff
@@ -0,0 +1,604 @@
...
+      }
+    },
+    "RepositoryBranch": {
+      "type": "string",
+      "defaultValue": "master",
+      "metadata": {
+        "description": "Source code branch to deploy."
+      }
+    },
+    "JobWindowUnit": {
```
wi-y, line 27: this is for stream analytics, but I am considering also using this value to set batching windows in the stream analytics replacement. haven't complete that work yet, but will in some future commits.

deploy/templates/default-azuredeploy-webjobs.json
```diff
@@ -0,0 +1,604 @@
...
+            "preloadEnabled": true
+          }
+        ],
+        "loadBalancing": "LeastRequests",
+        "experiments": {
+          "rampUpRules": []
+        },
+        "autoHealEnabled": false,
+        "localMySqlEnabled": false,
+        "ipSecurityRestrictions": [
```
wi-y, line 488: need to lock this down.

deploy/templates/default-azuredeploy-webjobs.json
```diff
@@ -0,0 +1,604 @@
...
+          "Default.html",
+          "Default.asp",
+          "index.htm",
+          "index.html",
+          "iisstart.htm",
+          "default.aspx",
+          "index.php",
+          "hostingstart.html"
+        ],
+        "netFrameworkVersion": "v4.0",
```
wi-y, line 462: need to add environment variable settings. currently the app is running off of what is set in appsettings.json or environment variables, but no environment variables are being set here.

deploy/scripts/Create-IomtWebJobsEnvironment.ps1
```diff
@@ -0,0 +1,109 @@
...
+(
+    [Parameter(Mandatory = $true)]
+    [ValidateNotNullOrEmpty()]
+    [ValidateLength(5,12)]
+	[ValidateScript({
+        if ("$_" -cmatch "(^([a-z]|\d)+$)") {
+            return $true
+        }
+        else {
+			throw "Environment name must be lowercase and numbers"
```
ms-teli, line 16: Nit: please fix the indention

deploy/scripts/Create-IomtWebJobsEnvironment.ps1
```diff
@@ -0,0 +1,109 @@
...
+    [string]$FhirServiceAudience,
+
+    [Parameter(Mandatory = $false)]
+    [string]$EnvironmentDeploy = $true
+)
+
+Function BuildPackage() {
+    try {
+        Push-Location $currentPath
+        cd ../../src/console/
```
ms-teli, line 49: Just a question: do we assume that the script is being executed at a specific location, like "deploy/scripts/"? is that possible to set a working directory with ```dotnet```?

ms-teli (approved):
Overall look good to me, it would be helpful to see some testing output of the scripts.

deploy/templates/default-azuredeploy-webjobs.json
```diff
@@ -0,0 +1,604 @@
...
+        "P2",
+        "P3",
+        "P4"
+      ],
+      "defaultValue": "S1",
+      "metadata": {
+        "description": "The app service plan tier to use for hosting the required Azure Functions."
+      }
+    },
+    "ResourceLocation": {
```
namalu, line 78: Since Stream Analytics is not used, we can probably expand the list of allowed locations. The only dependency that might restict the region is the App Service Plan. https://azure.microsoft.com/en-us/global-infrastructure/services/?products=app-service&regions=all

deploy/scripts/Create-IomtWebJobsEnvironment.ps1
```diff
@@ -0,0 +1,109 @@
...
+    })]
+    [string]$EnvironmentName,
+
+    [Parameter(Mandatory = $false)]
+    [ValidateSet('Australia East','East US','East US 2', 'West US', 'West US 2','North Central US','South Central US','Southeast Asia','North Europe','West Europe','UK West','UK South')]
+    [string]$EnvironmentLocation = "North Central US",
+
+    [Parameter(Mandatory = $false)]
+    [ValidateSet('R4')]
+    [string]$FhirVersion = "R4",
```
namalu, line 28: Since we only support R4, do we need this parameter?

deploy/scripts/Create-IomtWebJobsEnvironment.ps1
```diff
@@ -0,0 +1,109 @@
...
+    [string]$EnvironmentDeploy = $true
+)
+
+Function BuildPackage() {
+    try {
+        Push-Location $currentPath
+        cd ../../src/console/
+        dotnet restore
+        dotnet build --output $buildPath /p:DeployOnBuild=true /p:DeployTarget=Package
+    } catch {
```
namalu, line 52: nit: I believe that you can just do try/finally in PS (omit the catch) since we aren't taking any action on the error.

deploy/scripts/Create-IomtWebJobsEnvironment.ps1
```diff
@@ -0,0 +1,109 @@
...
+
+        Publish-AzWebApp -ArchivePath "$currentPath\iomtwebjobs.zip" -ResourceGroupName $resourceGroupName -Name $webAppName
+    } catch {
+        throw
+    } finally {
+        Pop-Location
+    }
+}
+
+Function Clean-Path($WebJobName) {
```
namalu, line 84: nit: The method name uses an unapproved verb (https://docs.microsoft.com/en-us/powershell/scripting/developer/cmdlet/approved-verbs-for-windows-powershell-commands?view=powershell-7.1)

Maybe Clear-Path?

deploy/scripts/Create-IomtWebJobsEnvironment.ps1
dustinburson: I would be better to just resolve the path and build against it if possible.  Changing directory and keeping target can get tricky.

---
In reply to: [544493385](https://github.com/microsoft/iomt-fhir/pull/83#discussion_r544493385) [](ancestors = 544493385)

deploy/templates/default-azuredeploy-webjobs.json
```diff
@@ -0,0 +1,604 @@
+{
+  "$schema": "http://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
```
dustinburson, line 2: >$schema [](start = 3, length = 7)

We can have a separate template for now but ideally I would like to collapse this and others into one template with conditionals.  We have many *different* flavors of ARM templates today and they are becoming harder to manage and keep in sync with changes.

deploy/scripts/Create-IomtWebJobsEnvironment.ps1
```diff
@@ -0,0 +1,109 @@
+<#
+.SYNOPSIS
+Creates a new IoMT FHIR Connector for Azure without using Stream Analytics
+.DESCRIPTION
+#>
+param
```
namalu, line 6: It might actually be a nice feature to be able to specify the resource group name instead of assuming the resource group will be named the same as all of the resources.

namalu (approved):

deploy/templates/default-azuredeploy-webjobs.json
```diff
@@ -0,0 +1,604 @@
...
+        "siteConfig": {},
+        "scmSiteAlsoStopped": false,
+        "clientAffinityEnabled": true,
+        "clientCertEnabled": false,
+        "hostNamesDisabled": false,
+        "containerSize": 0,
+        "dailyMemoryTimeQuota": 0,
+        "httpsOnly": false,
+        "redundancyMode": "None"
+      }
```
dustinburson, line 439: Settings are missing.

deploy/templates/default-azuredeploy-webjobs.json
```diff
@@ -0,0 +1,604 @@
...
+    },
+    {
```
dustinburson, line 441: So using the setting "PROJECT": "relative path to webjob project here" + 
{
          "apiVersion": "2015-08-01",
          "name": "web",
          "type": "sourcecontrols",
          "condition": "[variables('deploy_source_code')]",
          "dependsOn": [
            "[variables('app_service_resource_id')]",
            "[resourceId('Microsoft.Web/sites/config', variables('app_service_name'), 'appsettings')]"
          ],
          "properties": {
            "RepoUrl": "[parameters('RepositoryURL')]",
            "branch": "[parameters('RepositoryBranch')]",
            "IsManualIntegration": true
          }
        }

doesn't work for web jobs?

dustinburson (changes requested):
:clock1:

deploy/scripts/Create-IomtWebJobsEnvironment.ps1
wi-y: thanks - updated

deploy/scripts/Create-IomtWebJobsEnvironment.ps1
wi-y: that's a good idea. updated

deploy/templates/default-azuredeploy-webjobs.json
wi-y: updated

deploy/scripts/Create-IomtWebJobsEnvironment.ps1
wi-y: It is not really needed at this time but the other scripts like the sandbox one have it. I think it was added there for some of the investigation work to support R5

deploy/templates/default-azuredeploy-webjobs.json
wi-y: This seems like it could work since webjobs appear to only care about the file names and locations of the files.

I'll try a few things today but if it doesn't work I'll create a future work item to investigate trying to get rid of the script and try using only arm templates for deploying the webjobs.

dustinburson (approved):

PR #87, opened 2021-01-26
Support managed identity for connecting to storage

wi-y:
This PR uses managed identity to connect to storage rather than connection strings.


deploy/templates/default-azuredeploy-webjobs.json
```diff
@@ -410,7 +411,22 @@
           "keySource": "Microsoft.Storage"
         },
         "accessTier": "Hot"
-      }
+      },
+      "resources": [
+        {
+          "type": "Microsoft.Storage/storageAccounts/providers/roleAssignments",
+          "apiVersion": "2020-03-01-preview",
```
namalu, line 418: Do we need to use the preview API?

src/console/Program.cs
```diff
@@ -27,47 +30,95 @@ public class Program
...
+            System.Console.WriteLine($"Reading from event hub: {eventHubName}");
+            System.Console.WriteLine($"Logs and Metrics will be written to Application Insights");
             var ct = new CancellationToken();
+            await eventHubReader.RunAsync(incomingEventReader, ct);
+        }
+
+        public static string GetEventHubName(IConfiguration config)
+        {
+            var eventHub = Environment.GetEnvironmentVariable("WEBJOBS_NAME");
+            if (eventHub == null)
```
namalu, line 63: eventHub is a string, we should probably use !string.IsNullOrWhitespace(EventHubs).

src/lib/Microsoft.Health.Events/EventCheckpointing/StorageCheckpointClient.cs
```diff
@@ -33,21 +34,30 @@ public StorageCheckpointClient(StorageCheckpointOptions options, ITelemetryLogge
...
+
             BlobPrefix = options.BlobPrefix;
 
             _lastCheckpointMaxCount = int.Parse(options.CheckpointBatchCount);
             _checkpoints = new ConcurrentDictionary<string, Checkpoint>();
             _lastCheckpointTracker = new ConcurrentDictionary<string, int>();
-            _storageClient = new BlobContainerClient(options.BlobStorageConnectionString, options.BlobContainerName);
+
+            var checkpointUri = new Uri($"https://{accountName}.blob.core.windows.net/{containerName}");
+            var credential = new DefaultAzureCredential();
```
namalu, line 49: I'm not sure if we should tackle this now (because of our short timeline), but I'm wondering if it makes sense to support both connection string and managed identity. When working with one of our customers, we found a case where the IT team could not deploy our the IoMT Connector ARM template that used MI. The reason was that the user did not have the required permissions in the AAD tenant.

namalu (approved):

src/lib/Microsoft.Health.Events/EventCheckpointing/StorageCheckpointClient.cs
```diff
@@ -33,21 +34,30 @@ public StorageCheckpointClient(StorageCheckpointOptions options, ITelemetryLogge
...
+            var containerName = EnsureArg.IsNotNullOrWhiteSpace(options.BlobContainerName);
+
             BlobPrefix = options.BlobPrefix;
 
             _lastCheckpointMaxCount = int.Parse(options.CheckpointBatchCount);
             _checkpoints = new ConcurrentDictionary<string, Checkpoint>();
             _lastCheckpointTracker = new ConcurrentDictionary<string, int>();
-            _storageClient = new BlobContainerClient(options.BlobStorageConnectionString, options.BlobContainerName);
+
+            var checkpointUri = new Uri($"https://{accountName}.blob.core.windows.net/{containerName}");
```
namalu, line 48: I think the options should contain the fill Uri instead of constructing it here. The reason is, the Uri may be different for Government and Sovereign regions.

rogordon01: As another option you could accept the full Url to the blob storage endpoint as well as the container name. Then join the two together to build the final checkpointUri. That would get around the issue of different Uri's for different Government/Sovereign regions and also allow you to connect to build uri's to multiple containers if need be.

deploy/templates/default-azuredeploy-webjobs.json
```diff
@@ -205,7 +205,8 @@
     "app_service_resource_id": "[resourceId('Microsoft.Web/sites', variables('app_service_name'))]",
     "deploy_source_code": "[and(not(empty(parameters('repositoryUrl'))),not(empty(parameters('repositoryBranch'))))]",
     "sender_role": "[concat('/subscriptions/', subscription().subscriptionId, '/providers/Microsoft.Authorization/roleDefinitions/', '2b629674-e913-4c01-ae53-ef4638d8f975')]",
-    "receiver_role": "[concat('/subscriptions/', subscription().subscriptionId, '/providers/Microsoft.Authorization/roleDefinitions/', 'a638d3c7-ab3a-418d-83e6-5f17a39d4fde')]"
+    "receiver_role": "[concat('/subscriptions/', subscription().subscriptionId, '/providers/Microsoft.Authorization/roleDefinitions/', 'a638d3c7-ab3a-418d-83e6-5f17a39d4fde')]",
+    "storage_blob_contributor_role": "[concat('/subscriptions/', subscription().subscriptionId, '/providers/Microsoft.Authorization/roleDefinitions/', 'ba92f5b4-2d11-453d-a403-e96b0029c9fe')]"
```
rogordon01, line 209: Fyi: You can also use the ARM template function `subscriptionResourceId` to build the string and not use `concat`:

`"storage_blob_contributor_role": "[subscriptionResourceId('Microsoft.Authorization/roleDefinitions', 'ba92f5b4-2d11-453d-a403-e96b0029c9fe')]",`

deploy/templates/default-azuredeploy-webjobs.json
```diff
@@ -410,7 +411,22 @@
...
+          "type": "Microsoft.Storage/storageAccounts/providers/roleAssignments",
+          "apiVersion": "2020-03-01-preview",
+          "name": "[concat(variables('storage_account_name'), '/Microsoft.Authorization/', guid(uniqueString(variables('storage_account_name'))))]",
+          "dependsOn": [
+            "[resourceId('Microsoft.Storage/storageAccounts', variables('storage_account_name'))]"
+
+          ],
+          "properties": {
+            "roleDefinitionId": "[variables('storage_blob_contributor_role')]",
+            "principalId": "[reference(variables('app_service_resource_id'), '2015-08-01', 'Full').Identity.principalId]"
```
rogordon01, line 426: I don't see a `scope` applied here which would limit this role to a specific storage account. What `scope` does this role get applied to in that case? In other words does the MI have `blob contributor` access to a specific storage account or all storage accounts at the resource group/subscription level?

wi-y: This role assignment is nested under the storage account created above and just has access to that storage account.

rogordon01 (approved):

src/console/Program.cs
```diff
@@ -27,47 +30,95 @@ public class Program
...
+
+            System.Console.WriteLine($"Reading from event hub: {eventHubName}");
+            System.Console.WriteLine($"Logs and Metrics will be written to Application Insights");
             var ct = new CancellationToken();
+            await eventHubReader.RunAsync(incomingEventReader, ct);
+        }
+
+        public static string GetEventHubName(IConfiguration config)
+        {
+            var eventHub = Environment.GetEnvironmentVariable("WEBJOBS_NAME");
```
dustinburson, line 62: >Environment.GetEnvironmentVariable("WEBJOBS_NAME"); [](start = 27, length = 51)

Where does the WEBJOBS_NAME environment variable get set?  I would expect it to be a property on the web config in the Web Template but I don't see it.

src/console/Program.cs
```diff
@@ -112,6 +161,15 @@ public static ServiceProvider GetRequiredServiceProvider(IConfiguration config,
             }
         }
 
+        public static void ConfigureLogging(IConfiguration config, IServiceCollection serviceCollection)
+        {
+            var instrumentationKey = config.GetSection("APPINSIGHTS_INSTRUMENTATIONKEY").Value;
```
dustinburson, line 166: >APPINSIGHTS_INSTRUMENTATIONKEY [](start = 56, length = 30)

What happens when this is null?

src/lib/Microsoft.Health.Events/EventCheckpointing/StorageCheckpointClient.cs
dustinburson: There should be an option to build this from the resource id.  That should be less error prone then using the URI as Rob and Nate have mentioned which is problematic when targeting different clouds.

---
In reply to: [566303793](https://github.com/microsoft/iomt-fhir/pull/87#discussion_r566303793) [](ancestors = 566303793)

src/lib/Microsoft.Health.Events/EventCheckpointing/StorageCheckpointClient.cs
dustinburson: Ideally this should be DIed.  Can support default like here, connection string later if needed, or cert based identity if needed in the future.

---
In reply to: [565584072](https://github.com/microsoft/iomt-fhir/pull/87#discussion_r565584072) [](ancestors = 565584072)

src/lib/Microsoft.Health.Events/EventCheckpointing/StorageCheckpointClient.cs
dustinburson: There is BlobURIBuilder.  Unfortunately it doesn't appear to support different clouds built in but it would be better to use than string interpolation itself. [https://docs.microsoft.com/en-us/dotnet/api/azure.storage.blobs.bloburibuilder?view=azure-dotnet](https://docs.microsoft.com/en-us/dotnet/api/azure.storage.blobs.bloburibuilder?view=azure-dotnet)

---
In reply to: [566498449](https://github.com/microsoft/iomt-fhir/pull/87#discussion_r566498449) [](ancestors = 566498449,566303793)

deploy/templates/default-azuredeploy-webjobs.json
dustinburson: Can you try 2020-04-08?  That seems like the latest GA version.

---
In reply to: [564892943](https://github.com/microsoft/iomt-fhir/pull/87#discussion_r564892943) [](ancestors = 564892943)

dustinburson (changes requested):
:clock1:

src/console/Program.cs
wi-y: Azure Web Jobs sets WEBJOBS_NAME automatically when a web job is created/running. The reason I used this rather than the app service environment variables is that we need to be running two web jobs (normalization, fhir conversion) on the same app service and needed some way for the web job to figure out what it should be doing. I'm open to better ways of doing this.

src/console/Program.cs
wi-y: Updated. Now it will continue and just not log to app insights if null/empty.

src/lib/Microsoft.Health.Events/EventCheckpointing/StorageCheckpointClient.cs
wi-y: I've updated the code so that this is injected.

src/lib/Microsoft.Health.Events/EventCheckpointing/StorageCheckpointClient.cs
wi-y: Added full url for now.

deploy/templates/default-azuredeploy-webjobs.json
wi-y: thanks - updated.

deploy/templates/default-azuredeploy-webjobs.json
wi-y: I tried 2020-04-08 but it said that version didn't support role assignments, so will instead go with the latest non-preview version which is 2018-07-01

Status Message: No registered resource provider found for location 'northcentralus' and API version '2020-04-08' for
type 'roleAssignments'. The supported api-versions are '2014-04-01-preview, 2014-07-01-preview, 2014-10-01-preview,
2015-05-01-preview, 2015-06-01, 2015-07-01, 2016-07-01, 2017-05-01, 2017-09-01, 2017-10-01-preview,
2018-01-01-preview, 2018-07-01, 2018-09-01-preview, 2018-12-01-preview, 2019-04-01-preview, 2020-03-01-preview,
2020-04-01-preview'.

dustinburson: Thanks for checking Will.  We should probably follow up with the Azure Storage team.  We don't want to take a dependency on an API in preview if we can avoid it. Would you be able to create a work item for tracking?

---
In reply to: [567489805](https://github.com/microsoft/iomt-fhir/pull/87#discussion_r567489805) [](ancestors = 567489805)

src/console/Program.cs
dustinburson: Okay, I am still not following.  I will try to dig in a little more to figure it out.  I may sync with you to understand better.

---
In reply to: [567487431](https://github.com/microsoft/iomt-fhir/pull/87#discussion_r567487431) [](ancestors = 567487431)

deploy/templates/default-azuredeploy-webjobs.json
dustinburson: Nevermind I misread, it looks like 2018-07-01 works, thanks for the update.

---
In reply to: [569032564](https://github.com/microsoft/iomt-fhir/pull/87#discussion_r569032564) [](ancestors = 569032564,567489805)

src/lib/Microsoft.Health.Common/Auth/Provider/AzureCredentialProvider.cs
```diff
@@ -0,0 +1,38 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using Azure.Identity;
+
+namespace Microsoft.Health.Common.Auth
+{
+    public class AzureCredentialProvider : IAzureCredentialProvider
```
dustinburson, line 11: >AzureCredentialProvider [](start = 17, length = 23)

We already have an interface for IAuthService for use with FHIR auth.  Do you think there is an option to unify? Or do you think they need to be separate?

dustinburson: Just to clarify, I am okay leveraging this work and removing the IAuthService pieces if that makes more sense.

---
In reply to: [569059401](https://github.com/microsoft/iomt-fhir/pull/87#discussion_r569059401) [](ancestors = 569059401)

src/lib/Microsoft.Health.Events/EventCheckpointing/StorageCheckpointClient.cs
```diff
@@ -29,25 +30,53 @@ public class StorageCheckpointClient : ICheckpointClient
...
+            else if (!string.IsNullOrWhiteSpace(connectionString))
+            {
+                return new BlobContainerClient(connectionString, blobUri.BlobContainerName);
+            }
+            else
+            {
+                var ex = new Exception($"Unable to create blob container client for {blobUri}");
+                log.LogError(ex);
+                throw ex;
+            }
```
dustinburson, line 72: Can this be internalized into the credential provider itself?  I.E. creating a storage client for the provider should be a responsibility outside the StorageCheckpointClient and this class should just ask for the provider.  This refactor is a large improvement though!

src/lib/Microsoft.Health.Common/Auth/Service/ICredentialOptions.cs
```diff
@@ -0,0 +1,11 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Common.Auth
+{
+    public interface ICredentialOptions
```
dustinburson, line 8: >ICredentialOptions [](start = 21, length = 18)

Is there a benefit of an empty interface here?  Are you anticipating other types of CredentialOptions in the future?

src/lib/Microsoft.Health.Common/Auth/Provider/AzureCredentialProvider.cs
```diff
@@ -0,0 +1,38 @@
...
+    {
+        private CredentialOptions _credentialOptions;
+
+        public AzureCredentialProvider(ICredentialOptions options)
```
dustinburson, line 15: >ICredentialOptions [](start = 39, length = 18)

Per my comment on the interface I would be in favor of just going with the credential class without the interface.

src/console/Program.cs
```diff
@@ -27,47 +31,91 @@ public class Program
...
 
-            var serviceProvider = GetRequiredServiceProvider(config, eventHub);
             var logger = serviceProvider.GetRequiredService<ITelemetryLogger>();
-            var eventConsumers = GetEventConsumers(config, eventHub, serviceProvider, logger);
+
+            var eventConsumers = GetEventConsumers(config, managedIdentityCredentialService, eventHubName, serviceProvider, logger);
 
             var storageOptions = new StorageCheckpointOptions();
             config.GetSection(StorageCheckpointOptions.Settings).Bind(storageOptions);
-            storageOptions.BlobPrefix = eventHub;
```
namalu, line 54: I wonder is there is a pattern here that would provide more flexibility in the future. What about creating an interface IStorageCheckpointClientProvider that has a GetClient(IConfiguration config). This would allow for different implementations of IStorageCheckpointClientProvider, the default could be for MI, but a consumer could provide an implementation that still uses a connection string.

This will provide flexibility by allowing consumers to create their own appsettings.json scheme that works with their own implementation of IStorageCheckpointClientProvider

This pattern could be used for other clients.

wi-y: Yeah, that is a good idea. I ended up moving things around which should allow the customer to have more flexibility. I decided to create a BlobContainerClientFactory which can take an overload of IAzureCredentialProvider which can construct to blob client with some custom credentials implementation. 

BlobContainerClientFactory will also by default support default azure credential and connection string, and the user can toggle between those by changing the config environment variables.

src/lib/Microsoft.Health.Events/EventCheckpointing/StorageCheckpointClient.cs
wi-y: Thanks for the feedback. Went with a factory approach and now pass a storage client in as an argument.

src/lib/Microsoft.Health.Common/Auth/Provider/AzureCredentialProvider.cs
wi-y: I'll try to unify these in the next PR that I do for the fhir server. I need to spend a bit more time understanding how auth with the fhir server works and how it differs. We should be able to unify once I understand though.

src/lib/Microsoft.Health.Common/Storage/BlobContainerClientFactory.cs
```diff
@@ -0,0 +1,46 @@
...
+
+using System;
+using Azure.Identity;
+using Azure.Storage.Blobs;
+using EnsureThat;
+using Microsoft.Health.Common.Auth;
+
+namespace Microsoft.Health.Common.Storage
+{
+    public static class BlobContainerClientFactory
```
dustinburson, line 14: >static [](start = 11, length = 6)

Is there a reason for static? Normally it is best to have a class/interface so we can have dependency injection.  Overall this is a great improvement though.  Thank you. If this can be made non-static I am fine doing that in a separate PR.  I made a note and will create a work item unless there is a specific reason for the static.

src/lib/Microsoft.Health.Events/EventCheckpointing/StorageCheckpointClient.cs
```diff
@@ -29,25 +29,25 @@ public class StorageCheckpointClient : ICheckpointClient
...
-            EnsureArg.IsNotNullOrWhiteSpace(options.BlobContainerName);
-            EnsureArg.IsNotNullOrWhiteSpace(options.CheckpointBatchCount);
-
             BlobPrefix = options.BlobPrefix;
 
             _lastCheckpointMaxCount = int.Parse(options.CheckpointBatchCount);
             _checkpoints = new ConcurrentDictionary<string, Checkpoint>();
             _lastCheckpointTracker = new ConcurrentDictionary<string, int>();
-            _storageClient = new BlobContainerClient(options.BlobStorageConnectionString, options.BlobContainerName);
+            _storageClient = containerClient;
```
dustinburson, line 40: >containerClient [](start = 29, length = 15)

Can we do an isnull check?

src/lib/Microsoft.Health.Events/Repository/StorageManager.cs
```diff
@@ -13,12 +14,10 @@ public class StorageManager : IRepositoryManager
...
         private BlobContainerClient _blobContainer;
 
-        public StorageManager(string connectionString, string blobContainerName)
+        public StorageManager(Uri containerUri, BlobContainerClient containerClient)
         {
-            EnsureArg.IsNotNull(connectionString);
-            EnsureArg.IsNotNull(blobContainerName);
-
-            _blobContainer = new BlobContainerClient(connectionString, blobContainerName);
+            EnsureArg.IsNotNull(containerUri);
```
dustinburson, line 19: >containerUri [](start = 32, length = 12)

We have an IsNotNull check but the URI is not used.  Do we still need?

src/lib/Microsoft.Health.Events/Repository/StorageManager.cs
```diff
@@ -13,12 +14,10 @@ public class StorageManager : IRepositoryManager
...
+            _blobContainer = containerClient;
```
dustinburson, line 20: >containerClient [](start = 29, length = 15)

Can we do a _blobContainer = EnsureArg.IsNotNull(containerClient, nameof(containerClient));?

dustinburson (approved):
:shipit:

src/lib/Microsoft.Health.Events/Repository/StorageManager.cs
wi-y: Nope. I was leftover from before we passed the client in as an argument - removed the argument and the null check.

src/lib/Microsoft.Health.Common/Storage/BlobContainerClientFactory.cs
wi-y: thanks - updated to a normal non-static class

PR #90, opened 2021-02-10
Support managed identity with normalized event hub

wi-y:


deploy/templates/default-azuredeploy-webjobs.json
```diff
@@ -353,6 +364,48 @@
         ]
       }
     },
+    {
+      "type": "Microsoft.EventHub/namespaces/eventhubs/providers/roleAssignments",
+      "apiVersion": "2018-07-01",
```
rogordon01, line 369: Since this apiVersion is duplicated consider making it a variable.

deploy/templates/default-azuredeploy-webjobs.json
```diff
@@ -197,11 +197,20 @@
...
+      "metadata": {
+        "description": "The domain of the Azure Event Hubs."
+      }
     }
+
   },
   "variables": {
     "asa_job_name": "[parameters('ServiceName')]",
     "eventhub_namespace_name": "[parameters('ServiceName')]",
+    "eventhub_fqdn": "[concat(variables('eventhub_namespace_name'), '.', parameters('EventHubsDomain'))]",
```
rogordon01, line 213: The below should work as a cross cloud solution for obtaining the FQDN of the EventHub:

```
[reference(resourceId('Microsoft.EventHub/namespaces', variables('eventhub_namespace_name'))).serviceBusEndpoint]
```

src/console/Program.cs
```diff
@@ -115,6 +103,26 @@ public static EventProcessorClient GetEventProcessorClient(BlobContainerClient b
             return eventHubReader;
         }
 
+        public static EventProcessorClientFactoryOptions GetEventProcessorFactoryOptions(IConfiguration config)
```
rogordon01, line 106: Will we be building these options differently for Paas? I ask because EventHub names in Paas will be prefixed by the Iotconnector name as they all share the same Event Hub Namespace.

src/lib/Microsoft.Health.Events/EventHubProcessor/EventProcessorClientFactory.cs
```diff
@@ -0,0 +1,52 @@
...
+            {
+                throw new Exception($"Unable to create Event Hub processor client for {options.EventHubName}");
+            }
+        }
+
+        public EventProcessorClient CreateProcessorClient(IAzureCredentialProvider provider, EventProcessorClientFactoryOptions options, EventProcessorClientOptions eventProcessorClientOptions)
+        {
+            EnsureArg.IsNotNull(options);
+            EnsureArg.IsNotNull(provider);
+
```
rogordon01, line 47: Should we also be checking for a null EventProcessorClientOptions?

src/lib/Microsoft.Health.Events/EventHubProcessor/EventProcessorClientFactory.cs
```diff
@@ -0,0 +1,52 @@
...
+        public EventProcessorClientFactory(BlobContainerClient blobContainerClient)
+        {
+            _blobContainerClient = EnsureArg.IsNotNull(blobContainerClient, nameof(blobContainerClient));
+        }
+
+        public EventProcessorClient CreateProcessorClient(EventProcessorClientFactoryOptions options, EventProcessorClientOptions eventProcessorClientOptions)
+        {
+            EnsureArg.IsNotNull(options);
+
+            if (options.ServiceManagedIdentityAuth)
```
dustinburson, line 28: >ServiceManagedIdentityAuth [](start = 24, length = 26)

Will this be extended to support cert based auth as well for customer facing MI?

dustinburson (approved):
:shipit:

ms-teli:
Apologies, I forgot to submit the pending review.

src/console/Program.cs
```diff
@@ -54,7 +48,11 @@ public static async Task Main()
             var storageCheckpointClient = GetStorageCheckpointClient(blobContainerClientFactory, checkpointContainerOptions, storageOptions, logger, eventHubName);
             var eventConsumerService = new EventConsumerService(eventConsumers, logger);
 
-            var incomingEventReader = GetEventProcessorClient(storageCheckpointClient.GetBlobContainerClient(), eventHubOptions);
+            var eventProcessorOptions = GetEventProcessorFactoryOptions(config);
+            var eventProcessorClientFactory = new EventProcessorClientFactory(storageCheckpointClient.GetBlobContainerClient());
```
namalu, line 52: I think in all cases where you are using new to create these items could be simplified by adding these options objects and factory objects to the Service Collection. Then result would probably be getting the service provider, getting the EventProcessor from the provider and calling RunAsync().

src/lib/Microsoft.Health.Events/EventHubProcessor/EventProcessorClientFactoryOptions.cs
```diff
@@ -5,18 +5,18 @@
...
-            EventHubConnectionString = connectionString;
-            EventHubName = name;
-        }
-
-        public string EventHubConnectionString { get; set; }
+        public string EventHubConsumerGroup { get; set; }
 
         public string EventHubName { get; set; }
+
+        public string ConnectionString { get; set; }
```
namalu, line 16: Rather than overloading these 2 options objects with both authentication methods, could we create separate implementations, create a common interface (IEventProcessorClientFactory), and just change which one gets added to the Service Collection at startup? We could just default to ConnectionString in the project, and allow consumers to just change it if needed.

e.g. EventProcessorClientConnectionStringFactoryOptions is injected into EventProcessorClientConnectionStringFactory.

There would be an implementation for Managed Identity - EventProcessorClientManagedIdentityFactoryOptions is injected into EventProcessorClientManagedIdentityFactory.

Both implement IEventProcessorClientFactory.

namalu (changes requested):

deploy/templates/default-azuredeploy-webjobs.json
wi-y: I tried this out but it seems to only work for resources that are "outputted" or already created.

For inputs it seemed to return this error: "The template function 'reference' is not expected at this location.". Will probably just leave as is for now.

src/console/Program.cs
wi-y: The names of the variables like "eventHubName" were a bit misleading since it not really an event hub name but rather the type of application we are running so I updated the variable to "applicationType".

Right now we have one code base and two "apps" and depending on whether we are a "devicedata" app or a "normazlieddata" app we are injecting the appropriate startup classes. The name of the actual event hubs doesn't matter as those get passed in as separate environment variables. The app type is an environment variable which I've updated to Console:ApplicationType.

src/lib/Microsoft.Health.Events/EventHubProcessor/EventProcessorClientFactory.cs
wi-y: For cert based auth I created a second method in EventProcessorClientFactory which takes an IAzureCredentialProvider which is a way to specify a custom way of retrieving a token. I was thinking of passing in a provider that does the cert based auth.

EventProcessorClient CreateProcessorClient(IAzureCredentialProvider provider, BlobContainerClient blobContainerClient, EventProcessorClientFactoryOptions options, EventProcessorClientOptions eventProcessorClientOptions)

src/lib/Microsoft.Health.Events/EventHubProcessor/EventProcessorClientFactory.cs
```diff
@@ -0,0 +1,52 @@
...
+            {
+                var tokenCredential = new DefaultAzureCredential();
```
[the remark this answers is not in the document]
wi-y: I was thinking that we would want to make it easy to deploy and use service managed identity (where ManagedIdentityCredential is most appropriate) but also allow develops to test/develop locally with their own credentials (where VisualStudio/VisualStudioCode/environment variables credentials might be most appropriate) and DefaultAzureCredential seemed to handle that nicely. I felt like that benefit of testing locally using DefaultAzureCredential was worth it over only using ManagedIdentityCredential and not really having an easy way of testing locally.

src/lib/Microsoft.Health.Events/EventHubProcessor/EventProcessorClientFactory.cs
```diff
@@ -0,0 +1,52 @@
...
+                return new EventProcessorClient(_blobContainerClient, options.EventHubConsumerGroup, options.EventHubNamespaceFQDN, options.EventHubName, tokenCredential, eventProcessorClientOptions);
+            }
+            else if (!string.IsNullOrEmpty(options.ConnectionString))
```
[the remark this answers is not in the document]
wi-y: Yeah we have a backlog item - right now none of this console stuff is documented because we were planning on doing a couple more iterations before putting out some documentation just we don't have to keep changing it.

src/console/Normalize/Processor.cs
```diff
@@ -74,21 +70,21 @@ public async Task ConsumeAsync(IEnumerable<IEventMessage> events)
...
 
-        private IAsyncCollector<IMeasurement> CreateCollector(string eventHubName, string connectionString, IOptions<EventHubMeasurementCollectorOptions> options)
+        private IAsyncCollector<IMeasurement> CreateCollector()
         {
-            var client = options.Value.GetEventHubClient(eventHubName, connectionString);
-            return new MeasurementToEventAsyncCollector(new EventHubService(client));
+            var eventHubProducerOptions = new EventHubProducerClientOptions();
+            _env.GetSection("OutputEventHub").Bind(eventHubProducerOptions);
+
+            var eventHubProducerFactory = new EventHubProducerFactory();
```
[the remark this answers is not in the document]
wi-y: Thanks - updated to inject most of these. I moved most of the logic to Startup.cs. It did reduce the amount of code.

src/console/Program.cs
wi-y: Updated - moved to the service collection.

src/lib/Microsoft.Health.Events/EventHubProcessor/EventProcessorClientFactoryOptions.cs
wi-y: I tried this out and like the idea of separating in principal, but the problem seems to be what methods IEventProcessorClientFactory should have. We want to be able to pass in options into the factories but if the options can either be EventProcessorClientConnectionStringFactoryOptions or EventProcessorClientManagedIdentityFactoryOptions then I think that means that IEventProcessorClientFactory would have to have methods that accept those options, or some sort of generic options (but then we have to check/validate the generic options that are being passed in are connection options or MI options which seems pretty similar to what the code is currently doing).

PR #79, opened 2020-12-02
Fix for "PrincipalNotFound" error during deployment of roleAssignments resources

wkilday:
Fixes #37 

While repeatedly testing greenfield deployments with the [managed-identity-azuredeploy.json](https://github.com/microsoft/iomt-fhir/blob/master/deploy/templates/managed-identity-azuredeploy.json) template, I kept getting `PrincipalNotFound` errors in both `Microsoft.EventHub/namespaces/eventhubs/providers/roleAssignments` resources.

This [documentation](https://github.com/microsoft/iomt-fhir/blob/master/deploy/templates/managed-identity-azuredeploy.json) suggests that newly created service principals can take a while to replicate and role assignments for those principals can sometimes fail.  The fix is to specify the `principalType` of the role assignment as `ServicePrincipal`.

After making this change, I haven't run into any more `PrincipalNotFound` errors.

PR #91, opened 2021-03-03
Fix Function App unable to start

dustinburson:
Recent package upgrade accidently included a .NET 5 package upgrade.  Because App Services aren't yet upgraded to .NET 5 loading framework dependencies causes the application to crash.

* Fix Microsoft.Extensions.Options incorrectly upgraded to 5.0.0 (set to version 3.1.12).
* Fix XUnit dependencies added to non test projects.
* Update packages to max .NET 3.0 version where available.
* Update API version and principal types for the Event Hub Role assignments per suggested changes in https://github.com/microsoft/iomt-fhir/pull/79

### DIFF
--- a/src/lib/Microsoft.Health.Common/Microsoft.Health.Common.csproj
+++ b/src/lib/Microsoft.Health.Common/Microsoft.Health.Common.csproj
@@ -18,8 +18,8 @@
   <ItemGroup>
     <PackageReference Include="Azure.Storage.Blobs" Version="12.8.0" />
     <PackageReference Include="Ensure.That" Version="9.2.0" />
-    <PackageReference Include="Microsoft.Extensions.Options" Version="5.0.0" />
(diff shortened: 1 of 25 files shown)


src/lib/Microsoft.Health.Common/Microsoft.Health.Common.csproj
```diff
@@ -18,19 +18,18 @@
   <ItemGroup>
     <PackageReference Include="Azure.Storage.Blobs" Version="12.8.0" />
     <PackageReference Include="Ensure.That" Version="9.2.0" />
-    <PackageReference Include="Microsoft.Extensions.Options" Version="5.0.0" />
```
dustinburson, line 21: This was the specific package upgrade in the previous update causing the function app to crash on Azure App Services.

wi-y (approved):

PR #92, opened 2021-03-04
Add custom auth option for event producer factory

wi-y:
- Allows for a custom auth option when sending events to an event hub
- Can now use the following 3 options to connect: the service managed identity, a connection string, or a custom auth option (if supplied).
- Refactor the event hub processor factory and storage to behave the same way.


src/lib/Microsoft.Health.Common/Storage/BlobContainerClientFactory.cs
```diff
@@ -28,19 +28,15 @@ public BlobContainerClient CreateStorageClient(BlobContainerClientOptions option
...
                 return new BlobContainerClient(containerUri.ToString(), blobUri.BlobContainerName);
             }
+            else if (provider != null)
+            {
+                var tokenCredential = provider.GetCredential();
+                return new BlobContainerClient(containerUri, tokenCredential);
+            }
             else
             {
                 throw new Exception($"Unable to create blob container client for {blobUri}");
```
namalu, line 38: I think we should be more explicit as to why this failed so it's easier to debug.

wi-y: Yep makes sense. Updated with error messages that explain the invalid config / how we got to this point.

src/lib/Microsoft.Health.Events/Common/EventHubFormatter.cs
```diff
@@ -0,0 +1,33 @@
...
+        {
+            EnsureArg.IsNotEmptyOrWhiteSpace(host);
+
+            if (Uri.IsWellFormedUriString(host, UriKind.Absolute))
+            {
+                var uri = new Uri(host);
+                host = uri.Host;
+            }
+
+            if (Uri.CheckHostName(host) != UriHostNameType.Unknown)
```
ms-teli, line 23: Nit, I personal prefer put validations first and let it fail as early as possible.

e.g.

if (!Uri.IsWellFormedUriString(host, UriKind.Absolute)) {
   throw new Exception("not well formatted.");
}

var uri = new Uri(host);
host = uri.Host;

if(Uri.CheckHostName(host) == UriHostNameType.Unknown)
 throw new Exception("host is unknown");
)

return host;

src/lib/Microsoft.Health.Events/EventHubProcessor/EventProcessorClientFactory.cs
```diff
@@ -23,27 +24,24 @@ public EventProcessorClient CreateProcessorClient(BlobContainerClient blobContai
             if (options.ServiceManagedIdentityAuth)
```
ms-teli, line 24: Nit, would that make sense to add an enum property in the EventProcessorClientFactoryOptions to explicitly say the type of connection/authentication? The if-else seems able to handle overriding logic flow in the code itself but I'm not sure whether that's a valid use case.  Other than that, I think a single property to specify which type of authentication would make code cleaner.

ms-teli (approved):
Left some suggestions.

src/lib/Microsoft.Health.Events/EventHubProcessor/IEventProcessorClientFactory.cs
```diff
@@ -11,8 +11,6 @@ namespace Microsoft.Health.Events.EventHubProcessor
 {
     public interface IEventProcessorClientFactory
     {
-        EventProcessorClient CreateProcessorClient(BlobContainerClient blobContainerClient, EventProcessorClientFactoryOptions options, EventProcessorClientOptions eventProcessorClientOptions);
-
-        EventProcessorClient CreateProcessorClient(IAzureCredentialProvider provider, BlobContainerClient blobContainerClient, EventProcessorClientFactoryOptions options, EventProcessorClientOptions eventProcessorClientOptions);
+        EventProcessorClient CreateProcessorClient(BlobContainerClient blobContainerClient, EventProcessorClientFactoryOptions options, EventProcessorClientOptions eventProcessorClientOptions, IAzureCredentialProvider provider = null);
```
dustinburson, line 14: Should the provider be part of the method signature?  I feel like it can be injected into the ctor for the factory and resolved by DI and if it is needed it will be used.  I.E. The factory needs to be resolve any configuration we should support so it should have provider available.

Does the provider need to vary based on the context?  The concern here is we have an option signature and we allow the options to require the Azure provider but invoke could fail to provide causing an error.

src/lib/Microsoft.Health.Events/EventHubProcessor/EventProcessorClientFactory.cs
dustinburson: If we have a known set an enum seems warrented.

dustinburson (commented):
Okay as is but suggested improvments

src/lib/Microsoft.Health.Events/EventHubProcessor/EventProcessorClientFactory.cs
wi-y: Is the idea to have a "authentication type" property in EventProcessorClientFactoryOptions that is set to the type of connection/authentication and is calculated based on what options (connection string, event hub mi settings, other) are provided? I agree this could make the code cleaner, but just making sure I understand the comment.

Or do we want to user to explicitly set this new "authentication type" property, in addition to supplying the required connection options properties.

src/lib/Microsoft.Health.Events/EventHubProcessor/IEventProcessorClientFactory.cs
wi-y: I think I did this originally as part of the factory instead of the method, but we changed to the method because it might make sense to be able to generate multiple processors or producers with different configs. Right now there is not a requirement to do so but it could add flexibility in the future.

We were thinking it would be better to have 1 factory that could generate pretty much every kind of client by putting IAzureCredentialProvider in the method signature. If IAzureCredentialProvider is part of the factory instead of the method then we may need to generate a new factory for each custom auth credential provider if there is a requirement to do so in the future.

src/lib/Microsoft.Health.Events/EventHubProcessor/EventProcessorClientFactory.cs
ms-teli: I was thinking the second one in here, as user, I will have to tell this factory explicitly which auth option I want to use by setting the enum field.

ms-teli: Thanks for making the changes! They look good to me.

ms-teli (approved):

PR #94, opened 2021-03-16
Console app - use managed identity when connecting to fhir server

wi-y:
As part of this update I also removed the IAuthService interface and modified that two classes that implemented it to use TokenCredential instead. 

There seems recommendation from Azure to migrate from AppAuthentication (legacy) to Azure.Identity so I went ahead and updated the classes. Link to guidance: https://docs.microsoft.com/en-us/dotnet/api/overview/azure/app-auth-migration


src/lib/Microsoft.Health.Extensions.Fhir.R4/FhirClientFactory.cs
```diff
@@ -71,19 +86,26 @@ private static FhirClient CreateConfidentialApplicationClient()
 
         private class BearerTokenAuthorizationMessageHandler : HttpClientHandler
         {
-            public BearerTokenAuthorizationMessageHandler(IAuthService authService)
+            public BearerTokenAuthorizationMessageHandler(Uri uri, TokenCredential tokenCredentialProvider)
             {
-                AuthService = EnsureArg.IsNotNull(authService, nameof(authService));
+                TokenCredential = EnsureArg.IsNotNull(tokenCredentialProvider, nameof(tokenCredentialProvider));
+                Uri = EnsureArg.IsNotNull(uri);
```
namalu, line 92: nit: named(uri)

namalu (approved):

PR #95, opened 2021-03-18
Add unit tests for event batching and consumption

wi-y:


test/Microsoft.Health.Events.UnitTest/EventConsumerServiceTests.cs
```diff
@@ -0,0 +1,39 @@
...
+{
+    public class EventConsumerServiceTests
+    {
+        [Fact]
+        public async void GivenEventConsumer_WhenEventConsumerThrowsException_ThenRetryEventConsumer_Test()
+        {
+            var retries = 0;
+            var initialBatch = new List<EventMessage>()
+            {
+                new EventMessage("0", new ReadOnlyMemory<byte>(), 1, 1, new DateTime(2020, 12, 31, 5, 10, 20), new Dictionary<string, object>(), new ReadOnlyDictionary<string, object>(new Dictionary<string, object>()))
```
ms-teli, line 21: Nit, not sure if the EventMessage supports parameterless constructor so we can use the object initializer with sending properties to make this a little more readable.

ms-teli (approved):

test/Microsoft.Health.Events.UnitTest/EventBatchingTests.cs
```diff
@@ -0,0 +1,194 @@
...
+using Xunit;
+
+namespace Microsoft.Health.Events.UnitTest
+{
+    public class EventBatchingTests
+    {
+        [Fact]
+        public async void GivenBatchServiceCreated_WhenFirstEventReceived_ThenCreateQueueAndWindow_Test()
+        {
+            var eventConsumerService = Substitute.For<IEventConsumerService>();
```
namalu, line 19: Just a quick note, I see these objects getting created at the start of each test. You can actually do this in the constructor of the test class and set private fields that each test can use. The constructor is called at the start of each test so there is no concern of shared context between tests. The behavior is essentially the same, it's just less lines of code.

```c#
private IEventConsumerService _eventConsumerService;
private ICheckpointClient _checkpointClient;
private ITelemetryLogger _logger;
private EventBatchingOptions _options;

public EventBatchingTests()
{
     _eventConsumerService = Substitute.For<IEventConsumerService>();
     _checkpointClient = Substitute.For<ICheckpointClient>();
     _logger = Substitute.For<ITelemetryLogger>();
     _options = new EventBatchingOptions();
}
```

namalu (approved):

test/Microsoft.Health.Events.UnitTest/EventBatchingTests.cs
wi-y: Thanks for the suggestion. Updated.

PR #100, opened 2021-04-01
Log device data ingress as metric

wi-y:


src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementEventNormalizationService.cs
```diff
@@ -88,6 +90,13 @@ private async Task StartConsumer(ISourceBlock<EventData> producer, IAsyncCollect
                             (DateTime.UtcNow - evt.SystemProperties.EnqueuedTimeUtc).TotalMilliseconds);
 
                         var token = _converter.Convert(evt);
+                        var tokenString = JsonConvert.SerializeObject(token);
+                        var megabytes = Encoding.Unicode.GetBytes(tokenString).Length / Math.Pow(1000, 3); // Base 10 definition of MB;
+
+                        _log.LogMetric(
+                            IomtMetrics.DeviceEventIngress(),
+                            megabytes);
```
wi-y, line 98: Looks like Azure Monitor recommends using Bytes as the unit. Will update to Bytes

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementEventNormalizationService.cs
```diff
@@ -88,6 +90,13 @@ private async Task StartConsumer(ISourceBlock<EventData> producer, IAsyncCollect
                             (DateTime.UtcNow - evt.SystemProperties.EnqueuedTimeUtc).TotalMilliseconds);
 
                         var token = _converter.Convert(evt);
```
dustinburson, line 92: I am wondering if we have an opportunity to do something more efficiently in the EventDataWithJsonBodyToJTokenConverter or prior based off the raw EventData.

The EventData contains the byte array for the body so we can just record the size of that for the *data*.  The question is do we include the size of the properties/system properties (if we want to which your current method does).  If the performance impact ends up being high enough I think it is a reasonable compromise to just meter the 

It also might be worth looking at the latest https://github.com/Azure/azure-sdk-for-net/tree/master/sdk/eventhub/Azure.Messaging.EventHubs SDK to see if there is a method for calculating the size of the event message built in.

The one benefit of your approach that I like is it applies no matter the source as long as we convert the payload to JSON I just have a feeling we will want to end up with something more specialized per converter implementation to get the necessary efficiency to sustain throughput.

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementEventNormalizationService.cs
```diff
@@ -88,6 +90,13 @@ private async Task StartConsumer(ISourceBlock<EventData> producer, IAsyncCollect
                             (DateTime.UtcNow - evt.SystemProperties.EnqueuedTimeUtc).TotalMilliseconds);
 
                         var token = _converter.Convert(evt);
+                        var tokenString = JsonConvert.SerializeObject(token);
+                        var bytes = Encoding.Unicode.GetBytes(tokenString).Length;
+
+                        _log.LogMetric(
```
dustinburson, line 96: This will work for data collection but if we want to use this for a billing meter later we will need to think through placement/dedupping charges.

If the message fails after though should we charge for it?  Similarly if we have a compute failure and we reprocess a set of messages again because a failure on our I don't think it is appropriate to charge the customer again.  

Once again good enough for data collection but we probably need to more if we want to use it in a billing meter in the future.

dustinburson (dismissed):
Overall looks good. The one thing I would like addressed before committing is making it configurable with the default disabled to avoid impact on existing implementations and also better metric name.

Other comments are optional/thinking out loud/for later consideration

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementEventNormalizationService.cs
```diff
@@ -88,6 +90,13 @@ private async Task StartConsumer(ISourceBlock<EventData> producer, IAsyncCollect
                             (DateTime.UtcNow - evt.SystemProperties.EnqueuedTimeUtc).TotalMilliseconds);
 
                         var token = _converter.Convert(evt);
+                        var tokenString = JsonConvert.SerializeObject(token);
```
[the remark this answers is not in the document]
wi-y: I've added a setting "NormalizationService:LogDeviceIngressSizeBytes" to make this optional. It will default to false. Let me know if you have feedback on how this setting is injected. I tried to follow the approach extension/bindings/attribute as opposed to just passing a boolean in.

src/lib/Microsoft.Health.Fhir.Ingest/Telemetry/Metrics/IomtMetrics.cs
```diff
@@ -92,6 +92,15 @@ public static class IomtMetrics
                 { _operationDimension, ConnectorOperation.Normalization },
             });
 
+        private static Metric _deviceEventIngress = new Metric(
+            "DeviceEventIngress",
```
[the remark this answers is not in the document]
wi-y: Yep makes sense. Updated to DeviceIngressSizeBytes

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementEventNormalizationService.cs
wi-y: That is a good point. I plan on addressing this, but have not done so yet. I will address as part of this PR.

dustinburson: Thoughts on what we would use?  Event Sequence number?

dustinburson: I will hold off signing off till the changes make into the PR.  Other changes look good, thank you!  If you end up wanting to do the uniqueness change separately happy to sign off on the content of this PR as is now.  Thank you!

wi-y: I was considering a few solutions for de-duplication that cover the scenarios you mentioned.

Scenarios:
(1) Downstream failures that trigger a retry should not rebill for events when attempting a retry.
(2) App restarted/crashed and we re-read events. We don't want to bill for these events again.

Possible solutions for Scenario (1):
(1) Only logging ingress metric if downstream consumption succeeds. (Pros: seems simple? Cons: slightly different implementations for Functions and Console. Possible CPU hit if we need to iterate over the events again).
(2) Passing the metric to some other service/class with a unique guid or id (Event Sequence number + partition number) and having the service deal with deduplication. Could just be a hashmap that gets flushed every once in a while. (Pros: seems like it also could work. Cons: Uses Memory)
(3) Keeping a dictionary of recently logged metrics for each ingress event batches. If the event batch is already in the dictionary then don't log it again. Periodically purge dictionary values after a certain time period. (Cons: complicated, same memory concerns as above)

For OSS, I am thinking of going with solution (1) to only log the ingress metric if downstream consumption succeeds.

For OSS Functions, it does not appear that the NormalizeDeviceData function currently will retry if there is a downstream failure. It appears to logs an exception and continue on with the next batch. However, it may be best to only log the ingress metric if [dataNormalizationService.ProcessAsync](https://github.com/microsoft/iomt-fhir/blob/7794cbcc463e8d26c3097cd5e2243d770f26fe45/src/func/Microsoft.Health.Fhir.Ingest.Host/IomtConnectorFunctions.cs#L76) succeeds, so was thinking of logging the ingress metric right after the "await". The downside is we may need to iterate of the events to get their size. This will be off my default though.
 
For the OSS console based app, we do retry if there is a downstream failure. The retry occurs in [EventConsumerService ](https://github.com/microsoft/iomt-fhir/blob/7794cbcc463e8d26c3097cd5e2243d770f26fe45/src/lib/Microsoft.Health.Events/EventConsumers/Service/EventConsumerService.cs) and we essentially just call the Normalize processor again. In order to only log the ingress metric once, I am thinking of moving the ingress logging right after [dataNormalizationService.ProcessAsync](https://github.com/microsoft/iomt-fhir/blob/7794cbcc463e8d26c3097cd5e2243d770f26fe45/src/console/Normalize/Processor.cs#L77) so that it only gets logged if the await succeeds. 

Possible solutions for Scenario (2):
(1) If there is a compute failure or restart we need to persist some sort of state of what ingress metrics we have logged so far. If we periodically push some sort of ingress metric datetime in storage that contains a datetime value of that corresponds to the "last" ingress metric we logged for a particular event. We could then modify the code not to log ingress metrics if the event datetime < event datetime in storage. The "checkpoints" would essentially be like our event checkpoints but perhaps we could update these at a more frequent interval.
(2) Like Scenario 1 we could emit a unique id with the metric and have some downstream system consume all the metrics. We could do a query that excludes duplicate ids.

I think Scenario (2) mainly needs to be addressed for PaaS and not OSS at this time, so I am not planning on addressing in this PR.

Thoughts?

wi-y: I went with solution 1 in the latest commit but am open to other ideas too.

src/func/Microsoft.Health.Fhir.Ingest.Host/IomtConnectorFunctions.cs
```diff
@@ -74,6 +77,15 @@ public IomtConnectorFunctions(ITelemetryLogger logger)
 
                 IDataNormalizationService<EventData, IMeasurement> dataNormalizationService = new MeasurementEventNormalizationService(_logger, template);
                 await dataNormalizationService.ProcessAsync(events, output).ConfigureAwait(false);
+
+                if (normalizationSettings.Value.LogDeviceIngressSizeBytes)
+                {
+                    IEventProcessingMeter meter = new EventProcessingMeter();
```
rogordon01, line 83: Do we need to create a separate instance of `IEventProcessingMeter` for each invocation of `NormalizeDeviceData`? If not it might make sense to initialize one inside of the constructor. Or, pass in a singleton via dependency injection.

src/lib/Microsoft.Health.Fhir.Ingest/Service/EventProcessingMeter.cs
```diff
@@ -0,0 +1,41 @@
...
+    public class EventProcessingMeter : IEventProcessingMeter
+    {
+        public Task<EventStats> CalculateEventStats(EventData[] events)
+        {
+            double ingressSizeBytes = 0;
+            foreach (var e in events)
+            {
+                var body = e.Body.Array.Length;
+
+                foreach (KeyValuePair<string, object> entry in e.Properties)
```
rogordon01, line 22: Fyi; some nice one-liners to to replace these loops using Linq:

```
ingressSizeBytes = e.Properties.Aggregate(0.0, (current, entry) => current + Encoding.UTF8.GetBytes(entry.Key + entry.Value).Length)
ingressSizeBytes = e.SystemProperties.Aggregate(ingressSizeBytes, (current, entry) => current + Encoding.UTF8.GetBytes(entry.Key + entry.Value).Length)
```

src/func/Microsoft.Health.Fhir.Ingest.Host/IomtConnectorFunctions.cs
wi-y: Yes I think ideal performance would be to inject IDataNormalizationService<EventData, IMeasurement> and also IEventProcessingMeter but I do not think it is a bottleneck currently, so I did not invest time into doing it.

src/lib/Microsoft.Health.Fhir.Ingest/Service/EventProcessingMeter.cs
wi-y: Thanks, that is a lot cleaner. will update

src/lib/Microsoft.Health.Fhir.Ingest/Service/EventProcessingMeter.cs
```diff
@@ -0,0 +1,35 @@
...
+// -------------------------------------------------------------------------------------------------
+
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Azure.EventHubs;
+
+namespace Microsoft.Health.Fhir.Ingest.Service
+{
+    public class EventProcessingMeter : IEventProcessingMeter
```
rogordon01, line 13: A test case for this class would be nice

wi-y: makes sense. added

src/console/Normalize/Processor.cs
```diff
@@ -54,27 +54,45 @@ public async Task ConsumeAsync(IEnumerable<IEventMessage> events)
                 IomtMetrics.DeviceEvent(),
                     events.Count());
 
+            var eventBatchSizeBytes = 0;
```
dustinburson, line 57: >eventBatchSizeBytes [](start = 16, length = 19)

Can we use the Event Processing Meter here as well?  I like how you did that an it is a shame we aren't using it here.  This should technically perform better but I like the cleaner approach and the extra overhead shouldn't be too much now that we switched from JSON serialization.

src/lib/Microsoft.Health.Fhir.Ingest/Service/EventProcessingMeter.cs
```diff
@@ -0,0 +1,35 @@
...
+    {
+        public Task<EventStats> CalculateEventStats(EventData[] events)
+        {
+            double ingressSizeBytes = 0;
+
+            foreach (var e in events)
+            {
+                var bodySizeBytes = e.Body.Array.Length;
+                ingressSizeBytes = ingressSizeBytes + bodySizeBytes;
+                ingressSizeBytes = e.Properties.Aggregate(ingressSizeBytes, (current, entry) => current + Encoding.UTF8.GetBytes(entry.Key + entry.Value).Length);
```
dustinburson, line 23: >entry.Key + entry.Value [](start = 129, length = 23)

.NET strings are immutable, this will cause a new string to be created in memory.  Would be better to do GetBytes on each string section and add.  Also suggest using Encoding.UTF8.GetByteCount() over GetBytes.Length.

dustinburson (changes requested):
:clock1:

src/console/Normalize/Processor.cs
wi-y: Yep. I didn't use the meter due to the slight performance improvement from not having to iterate again, but since it is only a slight improvement I went ahead and changed console app to now use the Event Processing meter.

src/lib/Microsoft.Health.Fhir.Ingest/Service/EventProcessingMeter.cs
wi-y: That makes sense. Updated. Agreed GetByteCount is better

dustinburson (approved):
:shipit:

src/lib/Microsoft.Health.Fhir.Ingest/Service/EventProcessingMeter.cs
```diff
@@ -0,0 +1,36 @@
...
+    {
+        public Task<EventStats> CalculateEventStats(IEnumerable<EventData> events)
+        {
+            double ingressSizeBytes = 0;
+
+            foreach (var e in events)
+            {
+                var bodySizeBytes = e.Body.Array.Length;
+                ingressSizeBytes = ingressSizeBytes + bodySizeBytes;
+                ingressSizeBytes = e.Properties.Aggregate(ingressSizeBytes, (current, entry) => current + Encoding.UTF8.GetByteCount(entry.Key) + Encoding.UTF8.GetByteCount(entry.Value.ToString()));
```
dustinburson, line 24: >Aggregate [](start = 48, length = 9)

Looks like a forgot to publish this comment.  Suggestion: refactor into a common function for dictionaries.

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementEventNormalizationService.cs
dustinburson: Sounds good to me.  Thanks for the detailed break down.  For scenario 2 holding off here makes sense.  I personally like option 2 but it depends on what the billing system will support.

PR #101, opened 2021-04-12
Add additional event hub and fhir server error metrics

wi-y:
The bulk of the changes for adding event hub error metrics and FHIR server error metrics are in EventProcessor.cs and FhirClientFactory.cs. There were also some changes needed to move some of the logging around so that it was available at the time the FhirClientFactory is injected.


src/lib/Microsoft.Health.Extensions.Fhir.R4/FhirClientFactory.cs
```diff
@@ -105,6 +114,13 @@ protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage
                 var accessToken = await TokenCredential.GetTokenAsync(requestContext, CancellationToken.None);
                 request.Headers.Authorization = new AuthenticationHeaderValue("Bearer", accessToken.Token);
                 var response = await base.SendAsync(request, cancellationToken);
+
+                if (Logger != null && !response.IsSuccessStatusCode)
+                {
+                    var statusDescription = response.ReasonPhrase.Replace(" ", string.Empty);
+                    Logger.LogMetric(FhirClientMetrics.HandledException($"FhirServerError{statusDescription}", ConnectorOperation.FHIRConversion), 1);
```
dustinburson, line 121: Do we know how many different status descriptions we can expect?

dustinburson (approved):

src/lib/Microsoft.Health.Events/Telemetry/Metrics/EventMetrics.cs
```diff
@@ -100,5 +103,24 @@ public static Metric EventWatermark(string partitionId, DateTime dateTime)
...
+        public static Metric HandledException(string exceptionName, string connectorStage)
+        {
+            return new Metric(
+                exceptionName,
+                new Dictionary<string, object>
+                {
+                    { _nameDimension, exceptionName },
+                    { _categoryDimension, Category.Errors },
+                    { _errorTypeDimension, ErrorType.GeneralError },
+                    { _errorSeverityDimension, ErrorSeverity.Critical },
```
namalu, line 121: Should all of the EventHubsExceptions be critical?

src/lib/Microsoft.Health.Extensions.Fhir.R4/Telemetry/Metrics/FhirClientMetrics.cs
```diff
@@ -0,0 +1,38 @@
...
+        public static Metric HandledException(string exceptionName, string connectorStage)
+        {
+            return new Metric(
+                exceptionName,
+                new Dictionary<string, object>
+                {
+                    { _nameDimension, exceptionName },
+                    { _categoryDimension, Category.Errors },
+                    { _errorTypeDimension, ErrorType.GeneralError },
+                    { _errorSeverityDimension, ErrorSeverity.Critical },
```
namalu, line 33: Similarly to the EventHub error metric, should these all be critical too? For example, if the FHIR server returned a 429 because the customer had an unusual amount of data, could eventually correct itself. I'm also wondering if the FHIR server returns 404s for Device or Patients that are not found, which if the connector is set up to Create, this is a normal condition.

src/lib/Microsoft.Health.Extensions.Fhir.R4/FhirClientFactory.cs
wi-y: I had not tried to generate every possible http response, but wanted to capture any non-success error that could be returned. The main errors expect are "Forbidden", "Unauthorized", "Too Many Requests", and "Internal Server Error".

src/lib/Microsoft.Health.Extensions.Fhir.R4/Telemetry/Metrics/FhirClientMetrics.cs
wi-y: Yeah, I suppose 429 is more informational than critical. I can update that,

I did some testing (correct me if this is totally wrong) and believe the FHIR server returns a 200 when the Device or Patient are not found with an empty bundle instead of a 404, so I think we don't need to worry about any errors getting logged in that particular case.

src/lib/Microsoft.Health.Events/Telemetry/Metrics/EventMetrics.cs
wi-y: I think we should default to critical for now. I have only been able to generate these types of errors when there is some major connectivity or config issue that would prevent the connecting from fully functioning. 

If there was some non-critical exception that was thrown then we can handle it as a non-critical error, but so far have not seen one that is non-critical.

src/lib/Microsoft.Health.Extensions.Fhir.R4/Telemetry/Metrics/FhirClientMetrics.cs
namalu: You might be correct about the 200. It depends on whether or not the FHIR request is a search or a full path to a resource. In other words, if a search is performed with an identifier, it would return an empty bundle. However, if the url is constructed with the FHIR type and FHIR id, I believe the result will be a 404.

I'm not sure that we ever use the FHIR id and only search.

src/lib/Microsoft.Health.Events/Telemetry/Metrics/EventMetrics.cs
namalu: Sounds good. We can always make updates as needed.

src/lib/Microsoft.Health.Extensions.Fhir.R4/Telemetry/Metrics/FhirClientMetrics.cs
dustinburson: It is a 200 now because we do a search not a get since we need to look up by an identifier instead of the internal id.

A 404 could happen if the resource was deleted while we have it cached but that should trigger a retry.

src/lib/Microsoft.Health.Extensions.Fhir.R4/FhirClientFactory.cs
dustinburson: Thanks,  We may want to have a special classification for Forbidden/Unauthorized so we can really make them stand out but we can do that later if necessary.

PR #102, opened 2021-04-13
Add metrics for fhir observations created or updated

wi-y:


src/lib/Microsoft.Health.Fhir.Ingest/Telemetry/Metrics/IomtMetrics.cs
```diff
@@ -175,6 +176,23 @@ public static Metric NotSupported()
...
+        /// </summary>
+        /// <param name="resourceType">The type of FHIR resource that was saved.</param>
+        /// <param name="resourceOperation">The operation performed on the FHIR resource.</param>
+        public static Metric FhirResourceSaved(ResourceType resourceType, ResourceOperation resourceOperation)
+        {
+            return new Metric(
+                $"FhirResourceSaved",
+                new Dictionary<string, object>
+                {
+                    { _nameDimension, $"{resourceType}FhirResource{resourceOperation}" },
```
rogordon01, line 190: Will this output the ordinal of the enum, of the Name of the enum?

dustinburson: Should be the name.  This is an implicit ToString()

rogordon01 (approved):

src/lib/Microsoft.Health.Fhir.Ingest/Telemetry/Metrics/IomtMetrics.cs
dustinburson, line 190: So this will result in ObservationFhirResourceCreated & ObservationFhirResourceUpdated?  Right now we are only using this in exceptions so we have exceptions like DeviceNotFoundException, PatientNotFoundException, DeviceIdentityNotDefinedException, etc.  It is currently implicit that these are FHIR Resources.  Since we are capturing the operation dimension as FHIRConversion the FHIR Resource aspect seems implicit.  Thoughts on removing for succinctness and alignment with other telemetry?

dustinburson (approved):
Fine with this as is but added a suggestion on the naming.

src/lib/Microsoft.Health.Fhir.Ingest/Telemetry/Metrics/IomtMetrics.cs
wi-y: Agree it makes sense to align. Updated. It will now result in: ObservationCreated & ObservationUpdated

PR #108, opened 2021-06-04
Update OSS Sandbox deployment scripts for IoMT Connector

pallar-ms:
* Feature - [USER STORY 82590]( https://microsofthealth.visualstudio.com/Health/_workitems/edit/82590)
* Changes -
    * Update provisioning of resources with the managed identity setup to be the default setup for sandbox environment
    * Update sandbox creation script to reflect the above changes
    * Remove legacy auth setup script
    * Configure RBAC for the Azure API for FHIR and set the role for FHIR Data Contributor
    * Configure RBAC for AKVs and set the role for Key Vault User
    * Update documentation to reflect changes
* Note - 
    * Retaining other templates such as - consumption/premium/flexible (previously default) as is for now, although the sandbox script will not work as is if the IomtConnectorUrlTemplate is replaced with any of these, instead of the current default (managed-identity-azuredeploy).
    * If the linked template deployment fails, the RBAC for FHIR will not be setup as it has a dependency on the linked template. Successful redeployment will trigger that.
    * Seeing inconsistent failures with the app service deployment -
       - BadRequest - Service Unavailable for Microsoft.Web/sites/host
       - Principal does not exist - even though the principal type is set, but after updating to a diff api version, haven't seen this repro.
* Testing - 
     * Validated end to end script deployment, verified the resources were provisioned and the auth setup shows the roles assigned as intended.
image
image
     * Validated running the new script on a previous setup works fine too
     * Validated that the data is present in the FHIR server
image


docs/ARMInstallation.md
```diff
@@ -2,11 +2,11 @@
 This article details provisioning and installation of the IoMT FHIR Connector for Azure using an ARM template.
 
 ## ARM Template Provisioning
-<a href="https://portal.azure.com/#create/Microsoft.Template/uri/https%3A%2F%2Fraw.githubusercontent.com%2FMicrosoft%2Fiomt-fhir%2Fmaster%2Fdeploy%2Ftemplates%2Fdefault-azuredeploy.json" target="_blank">
+<a href="https://portal.azure.com/#create/Microsoft.Template/uri/https%3A%2F%2Fraw.githubusercontent.com%2FMicrosoft%2Fiomt-fhir%2Fmaster%2Fdeploy%2Ftemplates%2Fflexible-azuredeploy.json" target="_blank">
```
pallar-ms, line 5: Let me know if we don't want to keep this anymore. At first, I was planning on removing it, but from the readme it seemed like we want to keep this as a flexible template.

deploy/templates/default-azuredeploy-sandbox.json
```diff
@@ -238,12 +215,6 @@
       "location": "[parameters('FhirServiceLocation')]",
       "tags": {},
       "properties": {
-        "accessPolicies": "[parameters('FhirServiceAccessPolicies')]",
-        "authenticationConfiguration": {
-          "audience": "[parameters('FhirServiceResource')]",
-          "authority": "[parameters('FhirServiceAuthority')]",
-          "smartProxyEnabled": true
```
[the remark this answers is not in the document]
pallar-ms: Thanks for the context, added it back. 
And verified - https://ms.portal.azure.com/#@microsoft.onmicrosoft.com/resource/subscriptions/4092fd42-145e-4e91-aa15-96f66d2b6deb/resourceGroups/pallarsd12/providers/Microsoft.HealthcareApis/services/pallarsd12/authentication

docs/ARMInstallationManagedIdentity.md
```diff
@@ -51,7 +51,7 @@ The following parameters are provided by the ARM template:
...
 
 ## Post Deployment
-After the ARM template is successfully deployed, add the Managed Identity ID output by the ARM deployment to the Allowed Object IDs on the Authentication page of your Azure API for FHIR. This is the identity in Azure Active Directory that the IoMT FHIR Connector for Azure uses to connect to Azure API for FHIR. Also, the Authority on the Authentication page of your Azure API for FHIR should NOT be changed from the Azure Active Directory in your subscription, or this connection will be broken.
+After the ARM template is successfully deployed, [configure Azure RBAC for your Azure API for FHIR](https://docs.microsoft.com/en-us/azure/healthcare-apis/fhir/configure-azure-rbac) using the Managed Identity ID output by the ARM deployment in order to assign access for the IoMT FHIR Connector for Azure to the Azure API for FHIR data plane. Also, the Authority on the Authentication page of your Azure API for FHIR should NOT be changed from the Azure Active Directory in your subscription, or this connection will be broken.
```
namalu, line 54: "configure Azure RBAC for your Azure API for FHIR ... using the Managed Identity ID output by the ARM deployment" - Is this actually possible? I have tried to use the Managed Identity ID to set up RBAC, but when I enter the value in the search box, the identity is not listed. I have only been able to search using the name of the App Service Plan.

namalu (approved):

deploy/templates/flexible-azuredeploy.json
```diff
@@ -1,881 +1,881 @@
...
-        "attributes": {
-          "enabled": true
-        }
-      }
-    }
-  ],
-  "outputs": {
-  }
+{
+  "$schema": "http://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
```
dustinburson, line 2: >"http://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#", [](http://example.com/codeflow?start=13&length=81)

Just to verify, this is a rename of the old default template?

docs/ARMInstallation.md
dustinburson: I think it would be better to keep this page but update it to reference the default-managed-identity-azure template.  Basically we should use the MI template as the default and the customer can choose if they want to do the ARM deployment or the Sandox for it.

dustinburson: As follow up, we would functionality delete this and rename the ARMInstallationManagedIdentity to this file name update links as needed.

dustinburson (approved):
Looks good.  Left a few comments re: getting rid of the old template (now called flexible) in favor of just managed identity.

deploy/templates/default-managed-identity-azuredeploy.json
```diff
@@ -409,7 +410,7 @@
     },
     {
       "type": "Microsoft.EventHub/namespaces/eventhubs/providers/roleAssignments",
-      "apiVersion": "2018-07-01",
+      "apiVersion": "2018-09-01-preview",
```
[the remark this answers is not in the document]
dustinburson: Agreed we should avoid using preview APIs where we can

deploy/templates/default-managed-identity-azuredeploy.json
```diff
@@ -721,7 +722,7 @@
     },
     {
       "type": "Microsoft.KeyVault/vaults",
-      "apiVersion": "2016-10-01",
+      "apiVersion": "2021-04-01-preview",
```
dustinburson, line 725: We shouldn't use preview APIs

deploy/templates/default-managed-identity-azuredeploy.json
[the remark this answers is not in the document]
pallar-ms: For roleAssignments, I'm only seeing preview versions listed, and the one I used is the min version needed to fix the intermittent failures due to replication delay. Didn't want to update to latest preview in case it brought in some unwanted changes that we aren't ready for. But let me know if I am looking at the wrong place, or if its okay to update to the latest preview version listed.
https://docs.microsoft.com/en-us/azure/templates/microsoft.authorization/roleassignments?tabs=json

image

docs/ARMInstallationManagedIdentity.md
pallar-ms: Ahh good catch, it can be done via powershell through managed identity id, but that link only shows the portal setup and you are right, you can only search by name there. I'll update this to reflect the name of the app service. I don't think we need to output that though.

deploy/templates/flexible-azuredeploy.json
pallar-ms: Yes, given that in the ReadMe, we have this listed:
Open Source Deployment - Flexible: Describes how to deploy the IoMT FHIR Connector for Azure using another FHIR server or identity provider. 
Decided to let this be instead of removing it, and rename it instead.

docs/ARMInstallation.md
pallar-ms: Yeah that was what I gathered earlier as well, however in the ReadMe I read this and figured we'd want to have this as an option too if using another identity provider for whatever reason:
Open Source Deployment - Flexible: Describes how to deploy the IoMT FHIR Connector for Azure us
ing another FHIR server or identity provider.
If not, then yes makes sense to only just have the MI one(renamed to this) as have that be the only reference.

deploy/templates/default-managed-identity-azuredeploy.json
pallar-ms: The documentation around enableRbacAuthorization mentioned its only in preview, but I just checked version 2019-09-01 and it seems to work with that too.

deploy/templates/default-managed-identity-azuredeploy.json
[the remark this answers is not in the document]
pallar-ms: I'm actually quite confused on how to interpret the versions to use for role assignments. We aren't going by the Microsoft.Authorization/roleAssignments in these templates, but using resourcetype/providers/roleAssignments instead because of scope issues. 
The only reason I updated it here because it seems like to address the "Principal doesn't exist" based on this https://docs.microsoft.com/en-us/azure/role-based-access-control/role-assignments-template#new-service-principal

dustinburson (approved):

PR #114, opened 2021-06-18
Clear Event Hub Device Watermarks when source event hub changed

pallar-ms:
Description -
When customer changes their source of events (source event hub) the IoT Connector watermark should be reset and process all data available from the earliest available point in the event hub.

Changes - 
* Append event hub name to the checkpoint blob path
* Delete checkpoints corresponding to an older source event hub
* Note - This change is applicable to the event hubs for both the app types (i.e. watermarks will be handled in the same fashion for input event hub or the normalization event hub, but they are independently maintained.)

Testing - 
* Manual testing by running the Fhir.Ingest console app and verifying the various scenarios
    - Change event hub namespace
    - Change event hub within the same event hub namespace
    - Same event hub
* Added UTs for reset checkpoints

PENDING - 
Once this change takes effect, the checkpoints in the older path will not be honored, which will trigger reprocessing of the data. To avoid this, we can copy over the older checkpoints to the new location. Dustin is helping gauge how many active Gen2 connectors we have and the size of data ingested, to assess the impact and take the call accordingly. Will not be closing this PR until then, but putting this out now for first pass of review.

### DIFF
--- a/src/console/Startup.cs
+++ b/src/console/Startup.cs
@@ -106,8 +106,18 @@ namespace Microsoft.Health.Fhir.Ingest.Console
             var checkpointBlobClient = factory.CreateStorageClient(checkpointContainerOptions);
             var logger = serviceProvider.GetRequiredService<ITelemetryLogger>();
 
-            storageOptions.BlobPrefix = applicationType;
-            var checkpointClient = new StorageCheckpointClient(checkpointBlobClient, storageOptions, logger);
+            var eventProcessorOptions = new EventHubClientOptions();
+            if (applicationType == _normalizationAppType)
+            {
+                Configuration.GetSection("InputEventHub").Bind(eventProcessorOptions);
+            }
+            else if (applicationType == _measurementToFhirAppType)
+            {
+                Configuration.GetSection("NormalizationEventHub").Bind(eventProcessorOptions);
+            }
+
+            storageOptions.BlobPrefix = $"{applicationType}/{storageOptions.BlobPrefix}";
+            var checkpointClient = new StorageCheckpointClient(checkpointBlobClient, storageOptions, eventProcessorOptions, logger);
             return checkpointClient;
         }
 
--- a/src/lib/Microsoft.Health.Events/EventCheckpointing/StorageCheckpointClient.cs
+++ b/src/lib/Microsoft.Health.Events/EventCheckpointing/StorageCheckpointClient.cs
@@ -15,6 +15,7 @@ using Azure;
 using Azure.Storage.Blobs;
 using Azure.Storage.Blobs.Models;
 using EnsureThat;
+using Microsoft.Health.Events.Common;
 using Microsoft.Health.Events.Model;
 using Microsoft.Health.Events.Telemetry;
 using Microsoft.Health.Logging.Telemetry;
@@ -23,17 +24,25 @@ namespace Microsoft.Health.Events.EventCheckpointing
 {
     public class StorageCheckpointClient : ICheckpointClient
     {
-        private ConcurrentDictionary<string, Checkpoint> _checkpoints;
-        private ConcurrentDictionary<string, int> _lastCheckpointTracker;
-        private int _lastCheckpointMaxCount;
-        private BlobContainerClient _storageClient;
-        private ITelemetryLogger _log;
+        private readonly string _blobCheckpointPrefix;
+        private readonly string _blobPath;
+        private readonly ConcurrentDictionary<string, Checkpoint> _checkpoints;
+        private readonly int _lastCheckpointMaxCount;
+        private readonly ConcurrentDictionary<string, int> _lastCheckpointTracker;
+        private readonly ITelemetryLogger _log;
+        private readonly BlobContainerClient _storageClient;
 
-        public StorageCheckpointClient(BlobContainerClient containerClient, StorageCheckpointOptions options, ITelemetryLogger log)
+        public StorageCheckpointClient(BlobContainerClient containerClient, StorageCheckpointOptions options, EventHubClientOptions eventHubClientOptions, ITelemetryLogger log)
         {
             EnsureArg.IsNotNull(containerClient);
             EnsureArg.IsNotNull(options);
-            BlobPrefix = options.BlobPrefix;
+            EnsureArg.IsNotNull(eventHubClientOptions);
+
+            // Blob path for checkpoints includes the event hub name to scope the checkpoints per source event hub.
+            _blobCheckpointPrefix = $"{options.BlobPrefix}/checkpoint/";
+            _blobPath = eventHubClientOptions != default ?
+                $"{_blobCheckpointPrefix}{eventHubClientOptions.EventHubNamespaceFQDN}/{eventHubClientOptions.EventHubName}/" :
+                _blobCheckpointPrefix;
 
             _lastCheckpointMaxCount = int.Parse(options.CheckpointBatchCount);
             _checkpoints = new ConcurrentDictionary<string, Checkpoint>();
@@ -41,8 +50,6 @@ namespace Microsoft.Health.Events.EventCheckpointing
             _storageClient = containerClient;
             _log = log;
         }
-
-        public string BlobPrefix { get; }
 
         public BlobContainerClient GetBlobContainerClient()
         {
@@ -55,7 +62,7 @@ namespace Microsoft.Health.Events.EventCheckpointing
             EnsureArg.IsNotNullOrWhiteSpace(checkpoint.Id);
             var lastProcessed = EnsureArg.IsNotNullOrWhiteSpace(checkpoint.LastProcessed.DateTime.ToString("MM/dd/yyyy hh:mm:ss.fff tt"));
 
-            var blobName = $"{BlobPrefix}/checkpoint/{checkpoint.Id}";
+            var blobName = $"{checkpoint.Prefix}{checkpoint.Id}";
             var blobClient = _storageClient.GetBlobClient(blobName);
 
             var metadata = new Dictionary<string, string>()
@@ -87,7 +94,7 @@ namespace Microsoft.Health.Events.EventCheckpointing
 
         public Task<Checkpoint> GetCheckpointForPartitionAsync(string partitionIdentifier)
         {
-            var prefix = $"{BlobPrefix}/checkpoint/{partitionIdentifier}";
+            var prefix = $"{_blobPath}{partitionIdentifier}";
 
             Task<Checkpoint> GetCheckpointAsync()
             {
@@ -103,7 +110,7 @@ namespace Microsoft.Health.Events.EventCheckpointing
                         DateTimeOffset.TryParse(str, null, DateTimeStyles.AssumeUniversal, out lastEventTimestamp);
                     }
 
-                    checkpoint.Prefix = BlobPrefix;
+                    checkpoint.Prefix = _blobPath;
                     checkpoint.Id = partitionId;
                     checkpoint.LastProcessed = lastEventTimestamp;
                 }
@@ -136,7 +143,7 @@ namespace Microsoft.Health.Events.EventCheckpointing
                 var checkpoint = new Checkpoint();
                 checkpoint.LastProcessed = eventArgs.EnqueuedTime;
                 checkpoint.Id = partitionId;
-                checkpoint.Prefix = BlobPrefix;
+                checkpoint.Prefix = _blobPath;
 
                 _checkpoints[partitionId] = checkpoint;
                 var count = _lastCheckpointTracker.AddOrUpdate(partitionId, 1, (key, value) => value + 1);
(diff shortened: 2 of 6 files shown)


src/lib/Microsoft.Health.Events/EventCheckpointing/StorageCheckpointClient.cs
```diff
@@ -23,17 +24,25 @@ namespace Microsoft.Health.Events.EventCheckpointing
...
+
+        public StorageCheckpointClient(BlobContainerClient containerClient, StorageCheckpointOptions options, EventHubClientOptions eventHubClientOptions, ITelemetryLogger log)
         {
             EnsureArg.IsNotNull(containerClient);
             EnsureArg.IsNotNull(options);
-            BlobPrefix = options.BlobPrefix;
+            EnsureArg.IsNotNull(eventHubClientOptions);
+
+            // Blob path for checkpoints includes the event hub name to scope the checkpoints per source event hub.
+            _blobCheckpointPrefix = $"{options.BlobPrefix}/checkpoint/";
```
rogordon01, line 42: I'm not too familiar with the `default` keyword. My assumption is that `default` here would evaluate to `null` since `eventHubClientOptions` is a reference type.  If that is the case, then the check isn't needed as you already check that via EnsureArg above.

src/lib/Microsoft.Health.Events/EventHubProcessor/EventProcessor.cs
```diff
@@ -36,6 +36,9 @@ public async Task RunAsync(EventProcessorClient processor, CancellationToken ct)
         {
             EnsureArg.IsNotNull(processor);
 
+            // Reset previous checkpoints corresponding to an older source event hub (i.e. applicable if the source event hub changes)
+            await _checkpointClient.ResetCheckpointsAsync();
```
rogordon01, line 40: @wi-y  @dustinburson Since we are dealing with Async methods, what do you think about starting to use CancellationTokens where needed? For instance, the `CancellationToken` in scope here could be passed into `ResetCheckpointsAsync` and used in the subsequent async calls.

src/lib/Microsoft.Health.Events/EventCheckpointing/StorageCheckpointClient.cs
```diff
@@ -161,5 +168,28 @@ public async Task PublishCheckpointAsync(string partitionId)
...
         }
+
+        /// <summary>
+        /// Deletes the previously recorded checkpoints if the current checkpoint blob path (corresponding to the input event hub) has changed.
+        /// </summary>
+        public async Task ResetCheckpointsAsync()
+        {
+            try
+            {
+                foreach (BlobItem blob in _storageClient.GetBlobs(states: BlobStates.All, prefix: _blobCheckpointPrefix, cancellationToken: CancellationToken.None))
```
namalu, line 179: I'm not sure that this matters, but if the options.BlobPrefix ever changes, previous checkpoints will not be deleted. The only way I can see to address this issue would be to delete all of the blobs in the container, but then we would need 1 container per StorageCheckpointClient.

namalu (approved):

src/lib/Microsoft.Health.Events/EventCheckpointing/StorageCheckpointClient.cs
dustinburson: correct, EnureArg.IsNotNull should do it.  If it was a struct then this would be different.

src/console/Startup.cs
```diff
@@ -106,8 +106,18 @@ public virtual StorageCheckpointClient ResolveCheckpointClient(IServiceProvider
...
+            if (applicationType == _normalizationAppType)
+            {
+                Configuration.GetSection("InputEventHub").Bind(eventProcessorOptions);
+            }
+            else if (applicationType == _measurementToFhirAppType)
+            {
+                Configuration.GetSection("NormalizationEventHub").Bind(eventProcessorOptions);
+            }
+
+            storageOptions.BlobPrefix = $"{applicationType}/{storageOptions.BlobPrefix}";
```
dustinburson, line 119: >BlobPrefix [](http://example.com/codeflow?start=76&length=10)

Is this getting set today?  Do we have examples?

dustinburson: In the previous code this was only being set interally now we prefixing the application type in addition what is in the config.

src/lib/Microsoft.Health.Events/EventCheckpointing/StorageCheckpointClient.cs
```diff
@@ -161,5 +168,28 @@ public async Task PublishCheckpointAsync(string partitionId)
...
+                {
+                    if (!blob.Name.Contains(_blobPath, StringComparison.OrdinalIgnoreCase))
```
dustinburson, line 181: >Contains [](http://example.com/codeflow?start=35&length=8)

It would be good to add logging (info) so we can see when this is or is not occuring for troubleshooting purposes.

src/lib/Microsoft.Health.Events/EventCheckpointing/StorageCheckpointClient.cs
dustinburson: orhpaned prefixes aren't a concern if we as long as the value isn't revert back and they become in use again.

dustinburson (changes requested):
:clock1:

src/lib/Microsoft.Health.Events/EventHubProcessor/EventProcessor.cs
dustinburson: We should update and add proper cancellation token support but  lets create a backlog item and do it outside this PR.

src/console/Startup.cs
dustinburson: To Nate's comment if we allow this here the root can change and clean up could be prevented.

src/lib/Microsoft.Health.Events/EventCheckpointing/StorageCheckpointClient.cs
```diff
@@ -23,17 +24,25 @@ namespace Microsoft.Health.Events.EventCheckpointing
...
             EnsureArg.IsNotNull(containerClient);
             EnsureArg.IsNotNull(options);
-            BlobPrefix = options.BlobPrefix;
+            EnsureArg.IsNotNull(eventHubClientOptions);
+
+            // Blob path for checkpoints includes the event hub name to scope the checkpoints per source event hub.
+            _blobCheckpointPrefix = $"{options.BlobPrefix}/checkpoint/";
+            _blobPath = eventHubClientOptions != default ?
+                $"{_blobCheckpointPrefix}{eventHubClientOptions.EventHubNamespaceFQDN}/{eventHubClientOptions.EventHubName}/" :
+                _blobCheckpointPrefix;
```
wi-y, line 45: EventHubNamespaceFQDN isn't required to be set in all cases. When using a connection string, the user only has to specify a connection string and the consumer group. 

If the user does not specify EventHubNamespaceFQDN, then the blob path looking like this:
checkpoint/Normalization//checkpoint//devicedata

We can still make EventHubNamespaceFQDN required if we want. Just wanted to point out that it is not currently required by the EventProcessorClient constructor when using a connection string. EventHubNamespaceFQDN  is only required by the EventProcessorClient constructor when using managed identity.

[EventProcessorClientFactory](https://github.com/microsoft/iomt-fhir/blob/7794cbcc463e8d26c3097cd5e2243d770f26fe45/src/lib/Microsoft.Health.Events/EventHubProcessor/EventProcessorClientFactory.cs)

src/lib/Microsoft.Health.Events/EventCheckpointing/StorageCheckpointClient.cs
```diff
@@ -23,17 +24,25 @@ namespace Microsoft.Health.Events.EventCheckpointing
...
+        public StorageCheckpointClient(BlobContainerClient containerClient, StorageCheckpointOptions options, EventHubClientOptions eventHubClientOptions, ITelemetryLogger log)
         {
             EnsureArg.IsNotNull(containerClient);
             EnsureArg.IsNotNull(options);
-            BlobPrefix = options.BlobPrefix;
+            EnsureArg.IsNotNull(eventHubClientOptions);
+
+            // Blob path for checkpoints includes the event hub name to scope the checkpoints per source event hub.
+            _blobCheckpointPrefix = $"{options.BlobPrefix}/checkpoint/";
+            _blobPath = eventHubClientOptions != default ?
```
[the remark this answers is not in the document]
pallar-ms: Based on yours and Will's comment, added checks that ensure either `EventHubNamespaceFQDN `and `EventHubName ` or the corresponding parsed equivalents from `ConnectionString` is set.

src/lib/Microsoft.Health.Events/EventCheckpointing/StorageCheckpointClient.cs
pallar-ms: Oops , my bad, meant to remove it. Thanks for catching.

src/lib/Microsoft.Health.Events/EventCheckpointing/StorageCheckpointClient.cs
pallar-ms: Yes, that is correct that if the blob prefix changes, then checkpoints in the previous prefix will not be deleted, however, we will be reading from the new location and so the data ingestion will still be reset.
Having said that, I don't anticipate the prefix changing, right now in PaaS, we seem to be hardcoding the blobprefix with "devicedata", which actually wasn't being honored until now (we were always setting the prefix to app type eg: Normalization). The new prefix now will be *apptype*/*checkpoint prefix* while it was previously *apptype*/
So we will not be deleting the checkpoints that were already created till now (which I believe is okay(?) given we don't have many active iot connectors created through PaaS), but going forward since we don't expect to change the prefix, things should be cleared as intended.

src/console/Startup.cs
pallar-ms: So there is a property `"Checkpoint:BlobPrefix"` that was intended to be used as the blob prefix and which is parsed into storageOptions.BlobPrefix, but was actually not being honored, given that we were overriding it with app type. 
Today we seem to be hardcoding this value in PaaS to "devicedata"
https://microsofthealth.visualstudio.com/Health/_git/health-paas?path=%2Fsrc%2FResourceProviderApplication%2FResourceProvider.Iot.Workspace%2FModels%2FNormalizationEnvironmentVariables.cs&version=GBmaster&line=32&lineEnd=32&lineStartColumn=13&lineEndColumn=49&lineStyle=plain&_a=contents

It seemed like we wanted to expose this as an input but we weren't using it, so decided to append it to the app type (since we also want to scope the checkpoints to the app type as well). 
As explained in the other comment, with this change, we will not be cleaning up checkpoints up until this point that were in *app type*/checkpoint, but the data ingestion will be reset since its looking at the new path. Which I believe is okay? (given that we decided offline its potentially okay given we don't have many "production" Gen2 IoT connectors setup)
As for the property being taken as input, given that its hardcoded, I don't see the value personally, but maybe it was intended probably for people using the OSS ingest console app to be able to configure this property?

src/lib/Microsoft.Health.Events/EventCheckpointing/StorageCheckpointClient.cs
pallar-ms: Added, although it will log for each blob in that path. Which I think is okay in case we ran into issue deleting a particular one, but wanted to call out the possible bloat in logging.

src/lib/Microsoft.Health.Events/EventCheckpointing/StorageCheckpointClient.cs
pallar-ms: Updated to check the connection string properties if that is the auth type selected. 
(Thanks for bringing it up, didn't cross my mind, since I was using the default appsettings.json which had the FQDN and eventhub name params in there!)

src/lib/Microsoft.Health.Events/EventHubProcessor/EventProcessor.cs
pallar-ms: Agreed, I thought of using it for this call, but given the pattern in this file didn't seem to be doing that, decided to defer for now.
Created this WI -https://microsofthealth.visualstudio.com/Health/_workitems/edit/83006

src/console/Startup.cs
wi-y: The reason we do not honor Checkpoint:BlobPrefix is because we deploy OSS on Webjobs. Unfortunately because both of the Normalize and MeasurementToFhir Webjobs run on a single App Service, this means the environment variables are shared. Because the environment variables are shared, we run into issues where we need Normalization and MeasurementToFhir environment variables like Checkpoint:BlobPrefix and Console:ApplicationType to be set to different values. We could have created new environment variables for each application, or solved this another way, but happened to choose to override the two application specific variables Checkpoint:BlobPrefix and Console:ApplicationType variables at startup.

The main purpose for Checkpoint:BlobPrefix was just so we didn't hardcode anything in the class and had a way to to pass the prefix in as a parameter, but the intent was still to set upstream depending on the application type.

wi-y (approved):

src/console/Startup.cs
pallar-ms: Agreed, that makes sense for it to be different per app type. Which is why I retained that and just appended the passed in blob prefix param value.
As discussed offline with Will, we can technically do without blob prefix being passed in as an env variable, since we care about differentiating it based on app type. But say if in the future, we have more "data" types per app type, then appending that config driven blob prefix would help.
So the path should look like - Normalization/devicedata/*eventHubFQDN*/*eventHubName*

dustinburson: Thanks I really appreciate the discussion here.  This all makes sense :)  Can we have a comment (or comments) where appropriate so the expected path is clear.  Part of my concern if we have different layers touching the path and result takes a bit to decipher.

src/lib/Microsoft.Health.Events/EventCheckpointing/StorageCheckpointClient.cs
dustinburson: Thanks I think this is fine.  Should only happen on start up so bloat should be minimal.  It might be worth while to have a trace at the very beginning, i.e. Entering Reset Checkpoints help if there ever a scenario we blobs we thought should be deleted arent (i.e. ensure we are hitting the rest checkpoint code)/

dustinburson (approved):
Overall looks good.  Added a few more suggestions.

src/lib/Microsoft.Health.Events/EventHubProcessor/EventProcessor.cs
pallar-ms: Updated WI - https://microsofthealth.visualstudio.com/Health/_workitems/edit/83122

PR #115, opened 2021-06-21
Personal/rogordon/add expression content template

rogordon01:
This PR creates a new template type called `CalculatedFunctionContentTemplate`. The new type allows a user to choose an expression language for each available expression within the template. Besides `JsonPath` it also allows the use of `JMESPath`.


src/lib/Microsoft.Health.Fhir.Ingest.Template/CalculatedFunction/CalculatedFunctionContentTemplateFactory.cs
```diff
@@ -0,0 +1,33 @@
...
+// -------------------------------------------------------------------------------------------------
+
+using EnsureThat;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Health.Fhir.Ingest.Template.CalculatedFunction
+{
+    public class CalculatedFunctionContentTemplateFactory : HandlerProxyTemplateFactory<TemplateContainer, IContentTemplate>
+    {
+        private const string TargetTypeName = "CalculatedFunctionContent";
```
rogordon01, line 13: I'm considering renaming this to just `CalculatedContent`. Thoughts? I'd also update all classes and fields to be the same.

src/lib/Microsoft.Health.Fhir.Ingest.Template/CalculatedFunction/CalculatedFunctionContentTemplate.cs
```diff
@@ -0,0 +1,210 @@
...
+        [JsonConverter(typeof(StringEnumConverter))]
+        public virtual ExpressionLanguage DefaultExpressionLanguage { get; set; } = ExpressionLanguage.JsonPath;
+
+        [JsonIgnore]
+        public virtual IExpressionEvaluatorFactory ExpressionEvaluatorFactory { get; set; } = new CachingExpressionEvaluatorFactory();
+
+        public virtual IEnumerable<Measurement> GetMeasurements(JToken token)
+        {
+            EnsureArg.IsNotNull(token, nameof(token));
+            EnsureArg.IsTrue(token is JObject);
```
rogordon01, line 65: Based on the [IConverter](https://github.com/microsoft/iomt-fhir/blob/7794cbcc463e8d26c3097cd5e2243d770f26fe45/src/lib/Microsoft.Health.Fhir.Ingest/Data/EventDataWithJsonBodyToJTokenConverter.cs) we are using for Normalization, it appears that IContentTemplate will always receive a `JObject` here. I enforce this as I use functions related to JObjects later on in the code.

src/lib/Microsoft.Health.Fhir.Ingest.Template/CalculatedFunction/ExpressionEvaluatorFactory.cs
```diff
@@ -0,0 +1,52 @@
...
+    public class ExpressionEvaluatorFactory : IExpressionEvaluatorFactory
+    {
+        private JmesPath _jmesPath;
+
+        public ExpressionEvaluatorFactory()
+        {
+            _jmesPath = new JmesPath();
+
+            /*
+             * register additional custom functions.
```
rogordon01, line 21: I'm thinking that we can create a separate Nuget to hold these functions. That way they could be shared across teams. Thoughts?

ms-teli: I also like the idea of separating them out.

src/lib/Microsoft.Health.Fhir.Ingest.Template/CalculatedFunction/JMESPathExpressionEvaluator.cs
```diff
@@ -0,0 +1,50 @@
...
+            {
+                throw new ExpressionException($"Multiple tokens were returned using expression ${_expression.Value}");
+            }
+
+            return result;
+        }
+
+        public IEnumerable<JToken> SelectTokens(JToken data)
+        {
+            var result = _jmespathExpression.Transform(data).AsJToken();
```
rogordon01, line 40: We shouldn't throw an exception here. If we return a non-Array we should simply wrap the value inside of an Enumerable.

src/lib/Microsoft.Health.Fhir.Ingest.Template/CalculatedFunction/CalculatedFunctionContentTemplate.cs
```diff
@@ -0,0 +1,210 @@
...
+        public const string MatchedToken = "matchedToken";
+
+        [JsonProperty(Required = Required.Always)]
+        public virtual string TypeName { get; set; }
+
+        [JsonProperty(Required = Required.Always)]
+        public virtual string TypeMatchExpression { get; set; }
+
+        public virtual ExpressionLanguage? TypeMatchExpressionLanguage { get; set; }
+
```
rogordon01, line 29: The reason for a separate `XxxMatchExpression` and `XxxMatchExpressionLanguage` fields was to make the Swagger definition for this new template easier to write. I do think it would be easier to make the `XxxMatchExpression` an object that contains the expression and language. Doing this would reduce the duplication that we see here. So if we can produce Swagger that would support that I can make that change. Doesn't need to be in this PR, we can do it in a subsequent one.

src/lib/Microsoft.Health.Fhir.Ingest.Template/CalculatedFunction/CalculatedFunctionContentTemplate.cs
dustinburson: I believe there is an EnsureArg function to enforce a type.

src/lib/Microsoft.Health.Fhir.Ingest.Template/CalculatedFunction/ExpressionEvaluatorFactory.cs
dustinburson: Yes a separate library which will result in a separate package makes sense.  Long term we will want to migrate common code into the health shared repo.  The OSS project predates the existence of that repo and we haven't updated yet.

src/lib/Microsoft.Health.Fhir.Ingest.Template/CalculatedFunction/CalculatedFunctionContentTemplate.cs
```diff
@@ -0,0 +1,210 @@
...
+            EnsureArg.IsNotNull(token, nameof(token));
+            EnsureArg.IsTrue(token is JObject);
+
+            foreach (var typeToken in MatchTypeTokens(token as JObject))
+            {
+                yield return CreateMeasurementFromToken(typeToken);
+            }
+        }
+
+        protected T EvalExpression<T>(JToken token, params Expression[] expressions)
```
namalu, line 73: It seems like a lot of this code is duplicated from JsonPathContentTemplate; I'm wondering if it makes sense to either move the duplicate code to another class, or have the 2 inherit from a common class.

src/lib/Microsoft.Health.Fhir.Ingest.Template/CalculatedFunction/CachingExpressionEvaluatorFactory.cs
```diff
@@ -0,0 +1,41 @@
...
+
+using System.Collections.Generic;
+using EnsureThat;
+
+namespace Microsoft.Health.Fhir.Ingest.Template.CalculatedFunction
+{
+    public class CachingExpressionEvaluatorFactory : IExpressionEvaluatorFactory
+    {
+        private IExpressionEvaluatorFactory _wrappedExpressionEvaluatorFactory;
+        private IDictionary<string, IExpressionEvaluator> _expressionCache;
```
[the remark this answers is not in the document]
rogordon01: I believe you're correct, and that I need to update this to be thread safe. Each `EventData` is currently processed in a separate thread, so this class could called from multiple threads. @dustinburson and I had discussed adjusting that logic within the `MeasurementEventNormalizationService`, but I feel we should leave it unless there is data showing that otherwise.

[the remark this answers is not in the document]
dustinburson: Just a note on thread safety, the IDictionary should be fine to be accessed from multiple threads.  What is not thread safe is concurrent gets with writes.  So if you create the dictionary and after it is created it is immutable you should be fine.  That said we should enforce the behavior contractually using IReadonlyDictionary over a normal IDictionary https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.ireadonlydictionary-2?view=net-5.0

This would be preferable to a concurrent dictionary since it would avoid the locking overhead on reads when don't need it.

src/lib/Microsoft.Health.Fhir.Ingest.Template/CalculatedFunction/JMESPathExpressionEvaluator.cs
```diff
@@ -0,0 +1,50 @@
...
+            JmesPath.Expression jmespathExpression,
+            Expression templateExpression)
+        {
+            _jmespathExpression = EnsureArg.IsNotNull(jmespathExpression, nameof(jmespathExpression));
+            _expression = EnsureArg.IsNotNull(templateExpression, nameof(templateExpression));
+        }
+
+        public JToken SelectToken(JToken data)
+        {
+            var result = _jmespathExpression.Transform(data).AsJToken();
```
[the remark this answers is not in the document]
rogordon01: The call to transform always returns a `JmesPathArgument` and doesn't throw. Same thing for `AsJToken`.  As it turns out, if there is no match it will return a `JToken` with a `JTokenType` of NULL. So I'll update the code to account for that.

src/lib/Microsoft.Health.Fhir.Ingest.Template/CalculatedFunction/CalculatedFunctionContentTemplate.cs
rogordon01: I chose to leave them separate for now, as this new content template essentially replaces `JsonPathContentTemplate`. I could update it to make use of the new `Expression` class and then colocate common features, but since its legacy it feels like overkill. I could be persuaded otherwise, though.

rogordon01: After thinking about it a bit more the JsonPathContentTemplate and its subclass could benefit from the additional logging that I've added into the new `CalculatedContentTemplate` class. This could prove useful to folks who don't migrate right away. 
I'll go ahead with the refactor. If its the only thing holding up this PR I'll do it within another PR.

rogordon01: I've refactored the code to separate out the data from the behavior for ContentTemplates. Behavior has been moved into a class called `MeasurementExtractor` and the ContentTemplates themselves are just simple data containers.

src/lib/Microsoft.Health.Fhir.Ingest.Template/CalculatedFunction/JMESPathExpressionEvaluator.cs
```diff
@@ -0,0 +1,66 @@
...
+        public JMESPathExpressionEvaluator(
+            JmesPath.Expression jmespathExpression)
+        {
+            _jmespathExpression = EnsureArg.IsNotNull(jmespathExpression, nameof(jmespathExpression));
+        }
+
+        public JToken SelectToken(JToken data)
+        {
+            EnsureArg.IsNotNull(data);
+            var jmePathArgument = _jmespathExpression.Transform(data);
```
namalu, line 26: nit: jmesPathArgument

src/lib/Microsoft.Health.Fhir.Ingest.Template/CalculatedFunction/MeasurementExtractor.cs
```diff
@@ -0,0 +1,166 @@
...
+            // Current assumption is that the expressions should match a single element and will error otherwise.
+
+            string deviceId = GetDeviceId(token);
+            DateTime? timestamp = GetTimestamp(token);
+
+            string correlationId = null;
+            if (IsExpressionDefined(Template.CorrelationIdExpression))
+            {
+                correlationId = GetCorrelationId(token);
+            }
```
namalu, line 138: I see below that we simple call GetPatientId(token) and GetEncounterId(token) without doing this "IsExpressionDefined" check - Is there a reason we need to treat the correlationId differently?

src/lib/Microsoft.Health.Fhir.Ingest.Template/CalculatedFunction/MeasurementExtractor.cs
```diff
@@ -0,0 +1,166 @@
...
+                Type = Template.TypeName,
+                PatientId = GetPatientId(token),
+                EncounterId = GetEncounterId(token),
+                CorrelationId = correlationId,
+            };
+
+            if (Template.Values != null)
+            {
+                for (var i = 0; i < Template.Values.Count; i++)
+                {
```
namalu, line 153: Is there a reason we can't just use "foreach" here?

namalu (approved):

src/lib/Microsoft.Health.Fhir.Ingest.Template/CalculatedFunction/JMESPathExpressionEvaluator.cs
```diff
@@ -0,0 +1,50 @@
...
+// -------------------------------------------------------------------------------------------------
+
+using System.Collections.Generic;
+using DevLab.JmesPath;
+using EnsureThat;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Health.Fhir.Ingest.Template.CalculatedFunction
+{
+    public class JMESPathExpressionEvaluator : IExpressionEvaluator
```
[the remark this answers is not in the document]
rogordon01: Yup, will fix

src/lib/Microsoft.Health.Fhir.Ingest.Template/CalculatedFunction/MeasurementExtractor.cs
rogordon01: Previously I was constructing the name of the expression using `ValueExpression{idx}` but  then realized each ValueExpression had its own name. I'll switch to use a `foreach`.

src/lib/Microsoft.Health.Fhir.Ingest.Template/CalculatedFunction/MeasurementExtractor.cs
rogordon01: Interesting question. This logic is in the existing [JsonPathContentTemplate](https://github.com/microsoft/iomt-fhir/blob/master/src/lib/Microsoft.Health.Fhir.Ingest.Template/JsonPathContentTemplate.cs#L115). So I kept that logic here. 

My assumption was that `CorrelationId` isn't required, but if its specified then it must return a value.

src/lib/Microsoft.Health.Fhir.Ingest.Template/CalculatedFunction/IotJsonPathLegacyMeasurementExtractor.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    public class IotJsonPathLegacyMeasurementExtractor : LegacyMeasurementExtractor
+    {
+        private IotJsonPathContentTemplate _template;
+
+        public IotJsonPathLegacyMeasurementExtractor(
+            IotJsonPathContentTemplate template)
+            : base(new JsonPathCalculatedFunctionContentTemplateFacade<IotJsonPathContentTemplate>(template), new JsonPathExpressionEvaluatorFactory())
+        {
+            _template = template;
```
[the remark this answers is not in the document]
rogordon01: I'll add in a null check

src/lib/Microsoft.Health.Fhir.Ingest.Template/CalculatedFunction/JmesPathExpressionEvaluator.cs
```diff
@@ -0,0 +1,66 @@
...
+        public JmesPathExpressionEvaluator(
+            JmesPath.Expression jmespathExpression)
+        {
+            _jmespathExpression = EnsureArg.IsNotNull(jmespathExpression, nameof(jmespathExpression));
+        }
+
+        public JToken SelectToken(JToken data)
+        {
+            EnsureArg.IsNotNull(data);
+            var jmesPathArgument = _jmespathExpression.Transform(data);
```
pallar-ms, line 26: Does this throw and if so should this be wrapped in a try catch with a specific throw exception?

rogordon01: This doesn't throw. If no data can be found it will return a JToken with a type of 'Null'.

src/lib/Microsoft.Health.Fhir.Ingest.Template/CalculatedFunction/ExpressionEvaluatorFactory.cs
```diff
@@ -0,0 +1,59 @@
...
+            EnsureArg.IsNotEmptyOrWhiteSpace(expression?.Value, nameof(expression.Value));
+
+            var expressionLanguage = expression.Language ?? _defaultExpressionLanguage;
+
+            switch (expressionLanguage)
+            {
+                case ExpressionLanguage.JsonPath:
+                    return new JsonPathExpressionEvaluator(expression.Value);
+                case ExpressionLanguage.JmesPath:
+                    try
```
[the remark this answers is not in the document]
rogordon01: Sure that makes sense. I'll make the adjustment

src/lib/Microsoft.Health.Fhir.Ingest.Template/CalculatedFunction/CalculatedFunctionContentTemplate.cs
```diff
@@ -0,0 +1,45 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System.Collections.Generic;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Converters;
+
+namespace Microsoft.Health.Fhir.Ingest.Template.CalculatedFunction
```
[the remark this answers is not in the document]
dustinburson: Another option is to simply put this in the template namespace.  Especially since you are core functionality that the objects in the Template namespace are leveraging.

src/lib/Microsoft.Health.Fhir.Ingest.Template/CalculatedFunction/CalculatedFunctionContentTemplateFactory.cs
```diff
@@ -0,0 +1,95 @@
...
+
+            var calculatedFunctionTemplate = jsonTemplate.Template.ToValidTemplate<CalculatedFunctionContentTemplate>();
+            var measurementExtractor = new MeasurementExtractor(calculatedFunctionTemplate, CreateExpressionEvaluatorFactory(calculatedFunctionTemplate));
+            return measurementExtractor;
+        }
+
+        private IExpressionEvaluatorFactory CreateExpressionEvaluatorFactory(CalculatedFunctionContentTemplate template)
+        {
+            var evaluatorCache = new Dictionary<string, IExpressionEvaluator>();
+            IExpressionEvaluatorFactory expressionEvaluator = new ExpressionEvaluatorFactory(_jmesPath, template.DefaultExpressionLanguage);
```
dustinburson, line 50: This might be better as a dependency injected to aid with testing.

src/lib/Microsoft.Health.Fhir.Ingest.Template/CalculatedFunction/CalculatedFunctionContentTemplateFactory.cs
```diff
@@ -0,0 +1,95 @@
...
+
+            return new CachingExpressionEvaluatorFactory(new ReadOnlyDictionary<string, IExpressionEvaluator>(evaluatorCache));
+        }
+
+        private void AddExpression(
+            IDictionary<string, IExpressionEvaluator> cache,
+            Expression expression,
+            IExpressionEvaluatorFactory expressionEvaluator)
+        {
+            if (expression != null)
```
dustinburson, line 72: Thoughts on introducing an ILogger to this class and logging these expressions?  Would be good to have a log for the null condition as well.

src/lib/Microsoft.Health.Fhir.Ingest.Template/CalculatedFunction/ExpressionEvaluatorFactory.cs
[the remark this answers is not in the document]
dustinburson: That seems like a better approach to me as well

src/lib/Microsoft.Health.Fhir.Ingest.Template/CalculatedFunction/ExpressionEvaluatorFactory.cs
```diff
@@ -0,0 +1,59 @@
...
+            _defaultExpressionLanguage = defaultLanguage;
+        }
+
+        public IExpressionEvaluator Create(Expression expression)
+        {
+            EnsureArg.IsNotEmptyOrWhiteSpace(expression?.Value, nameof(expression.Value));
+
+            var expressionLanguage = expression.Language ?? _defaultExpressionLanguage;
+
+            switch (expressionLanguage)
```
dustinburson, line 39: Suggestion: If you do Nate's suggestion you can switch to a switch expression which I feel is more concise and a little easier to read: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/switch-expression

dustinburson (commented):
Still reviewing but posting comments I have so far.  Most of the comments are minor/suggestions.

src/lib/Microsoft.Health.Fhir.Ingest.Template/CalculatedFunction/ExpressionJsonConverter.cs
```diff
@@ -0,0 +1,43 @@
...
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Health.Fhir.Ingest.Template.CalculatedFunction
+{
+    public class ExpressionJsonConverter : JsonConverter
```
dustinburson, line 12: Is the reason for the converter for backwards compatibility with existing templates?

dustinburson: If so it would be good to add comment to this class to explain its use.

src/lib/Microsoft.Health.Fhir.Ingest.Template/CalculatedFunction/MeasurementExtractor.cs
dustinburson: That is correct.  The correlation id support came later.  When I added it I thought about doing the same check for PatientId and EncounterId but didn't because it would change current behavior.

We can revisit if we would like.  Same question was brought up in the PR that introduced the correlation id support; https://github.com/microsoft/iomt-fhir/pull/47/files

src/lib/Microsoft.Health.Fhir.Ingest.Template/CalculatedFunction/MeasurementExtractor.cs
```diff
@@ -0,0 +1,165 @@
...
+using Microsoft.Health.Fhir.Ingest.Data;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Health.Fhir.Ingest.Template.CalculatedFunction
+{
+    public class MeasurementExtractor : IContentTemplate
+    {
+        public const string MatchedToken = "matchedToken";
+
+        public MeasurementExtractor(
```
dustinburson, line 20: It would be good to take this opportunity to inject an ILogger here for better traceability.

src/lib/Microsoft.Health.Fhir.Ingest.Template/CalculatedFunction/ExpressionJsonConverter.cs
rogordon01: Its a custom converter which allows use to build `Expression` objects that are represented as either String or JObjects. For example, it supports the below Json snippets:

```
 {
  "typeMatchExpression" : "@.heartRate"
}
```

and 

```
 {
  "typeMatchExpression" : {
       "value" : "@.heartRate",
       "language" : "JsonPath"
  }
}
```

src/lib/Microsoft.Health.Fhir.Ingest.Template/CalculatedFunction/ExpressionEvaluatorFactory.cs
rogordon01: Will do

src/lib/Microsoft.Health.Fhir.Ingest.Template/CalculatedFunction/CalculatedFunctionContentTemplateFactory.cs
rogordon01: Will do

src/lib/Microsoft.Health.Fhir.Ingest.Template/CalculatedFunction/CalculatedFunctionContentTemplateFactory.cs
rogordon01: Sure thing. After working in the code base I think that we could benefit from a general refactor to move more of our initialization code into the DI framework. For example, `CollectionContentTemplateFactory` could be a DI Singleton versus referencing it via the static class instance as we do today. In fact it would be nice to do what @namalu did for the Workspace Platform repo and refactor each application into its own HostedService. I can do that in a follow on PR, as this one is getting a bit big.

src/lib/Microsoft.Health.Fhir.Ingest.Template/CalculatedFunction/CalculatedFunctionContentTemplate.cs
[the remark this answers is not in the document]
rogordon01: The reason I had moved this into its own namespace is that I felt that the `Template` namespace was getting pretty big, and didn't want to add to it with 20+ new classes. But you're right in that there is a strong coupling to the core functionality in the Template namespace. So I'll just get rid of `CalculatedFunction` namespace and move everything  into the `Template` namespace.

src/lib/Microsoft.Health.Fhir.Ingest.Template/CalculatedFunction/ExpressionEvaluatorFactory.cs
rogordon01: Actually, I had tried using switch expressions before and realized I couldn't. It's a function of C# 8.0+ and this project is set to use `netstandard2.0` and `LanguageVersion=7.3`. So if we want to do this I believe that we'd need to update our target framework, which might also mean updating our build pipeline.

rogordon01: Nevermind; looks like you can update the language version independently of the target framework.

src/lib/Microsoft.Health.Fhir.Ingest.Template/CalculatedFunction/CalculatedFunctionContentTemplateFactory.cs
rogordon01: The `null` case is valid; it may be the user hasn't supplied a value for a non-required expression. What might be useful is to add in a parameter `isRequired` and throw an exception if a required expression is `null`. I do that when performing the actual Normalization operation but doing it earlier when we construct the content template couldn't hurt.

src/lib/Microsoft.Health.Fhir.Ingest.Template/CalculatedFunction/MeasurementExtractor.cs
rogordon01: I don't have any specific logs/telemetry to produce here at the moment. But I've added the `iLogger` as a member of the factory that creates this class, so if we decide on telemetry here it'll be trivial to pass an `iLogger` in.

src/lib/Microsoft.Health.Fhir.Ingest.Template/Microsoft.Health.Fhir.Ingest.Template.csproj
```diff
@@ -1,10 +1,9 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
```
rogordon01, line 3: I've gone ahead and updated this to `netcoreapp3.1`. This was needed in order to use the `Microsoft.Health.Logging` package. This hasn't caused any adverse affects AFAIK but please let me know if this could cause issues.

src/lib/Microsoft.Health.Fhir.Ingest.Template/CalculatedFunction/CalculatedFunctionContentTemplateFactory.cs
rogordon01: Done

rogordon01:
All changes have been addressed. I've deployed these changes to a running test instance of IotConnector. I've tested this by sending events targeting `JsonPathContentTemplate` and `CalculatedFunctionContentTemplate`. All looks well.

src/lib/Microsoft.Health.Expressions/Microsoft.Health.Expressions.csproj
```diff
@@ -1,6 +1,6 @@
 <Project Sdk="Microsoft.NET.Sdk">
 	<PropertyGroup>
-		<TargetFramework>netstandard2.0</TargetFramework>
+		<TargetFramework>netcoreapp3.1</TargetFramework>
```
dustinburson, line 3: This is technically a breaking change.  I assume this is needed because of the addition of JmesPath.Net?  This is currently used in the RP for validation.  I believe netcore3.1 will be compatible there but can you add a work item to update and verify after this PR is complete?

rogordon01: This is because I wanted to incorporate `Microsoft.Health.Logging` which is `netcoreapp3.1`

rogordon01: Actually it appears that the project in the RP that is using `Microsoft.Health.Fhir.Ingest.Template` has a targetFramework of `net472`, which only supports `netstandard2.0`. So I don't think that this will work. I'm assuming that the RP needs .netframework and runs on Windows so I'll have to adjust things here.

dustinburson: Sorry yes, the template will be used in the RP code.  It should be fine but we should verify.

rogordon01: I just built a local `Microsoft.Health.Fhir.Ingest.Template` nuget and added it to the  RP project. It doesn't work; apparently netcoreapp3.1 supports netstandard2.1 and `net472` only supports `netstandard2.0`. I've reverted this back to `netstandard2.0`. I'm also going to swtich `Microsoft.Health.Logging` to `netstandard2.0` as it doesn't need to be set to `netcoreapp3.1`

dustinburson: Actually, I just tested locally and it doesn't seem like there is any reason why the logging library can't be netstandard2.0.  Can we update it and change these back to netstandard2.0?

rogordon01: Yup, I've done just that and pushed up the changes.

src/lib/Microsoft.Health.Fhir.Ingest.Template/Microsoft.Health.Fhir.Ingest.Template.csproj
```diff
@@ -1,10 +1,9 @@
...
     <CodeAnalysisRuleSet>..\..\..\CustomAnalysisRules.ruleset</CodeAnalysisRuleSet>
     <HighEntropyVA>true</HighEntropyVA>
     <RootNamespace>Microsoft.Health.Fhir.Ingest.Template</RootNamespace>
-    <LangVersion>7.3</LangVersion>
```
dustinburson, line 7: ><LangVersion>7.3</LangVersion> [](http://example.com/codeflow?start=4&length=30)

Please add this back.  You can update it to <LangVersion>8.0</LangVersion> which it appears you need with the latest changes but without builds can fail.  I.E. the latest fails to build on my dev machine.

rogordon01: I actually added that in earlier this morning. Strange its not showing up in github... will try refreshing my browser

rogordon01: There it is. Must have been some Github weirdness.

dustinburson:
>         public CollectionContentTemplateFactory(params ITemplateFactory<TemplateContainer, IContentTemplate>[] factories)

If we add a new ctor with IEnumerable<ITemplateFactory<TemplateContainer, IContentTemplate>> you can leverage DI to define these. 

---
Refers to: src/lib/Microsoft.Health.Fhir.Ingest.Template/CollectionContentTemplateFactory.cs:18 in f7db2e2. [](commit_id = f7db2e2961e2ad47e695aced2e98a350742d8272, deletion_comment = False)

src/console/Startup.cs
```diff
@@ -188,5 +194,28 @@ public string GetConsoleApplicationType()
...
             return applicationType;
         }
+
+        private void AddContentTemplateFactories(IServiceCollection services)
+        {
+            services.AddSingleton<IExpressionRegister>(sp => new AssemblyExpressionRegister(typeof(IExpressionRegister).Assembly, sp.GetRequiredService<ITelemetryLogger>()));
+            services.AddSingleton<IExpressionEvaluatorFactory>(
+                sp =>
+                {
+                    var jmesPath = new JmesPath();
```
dustinburson, line 204: >var jmesPath = new JmesPath(); [](http://example.com/codeflow?start=20&length=30)

I think a better approach here would be to define the singleton for building the JmesPath that includes setting the expressionRegister, and using normal DI for the jmesPath object for the TemplateExpressionEvaluatorFactory

dustinburson: Main weakness here is if anything else needs the JmesPath object in the future it won't properly instantiated.

src/console/Startup.cs
```diff
@@ -188,5 +194,28 @@ public string GetConsoleApplicationType()
...
+                });
+            services.AddSingleton<ITemplateFactory<TemplateContainer, IContentTemplate>, JsonPathContentTemplateFactory>();
+            services.AddSingleton<ITemplateFactory<TemplateContainer, IContentTemplate>, IotJsonPathContentTemplateFactory>();
+            services.AddSingleton<ITemplateFactory<TemplateContainer, IContentTemplate>, IotCentralJsonPathContentTemplateFactory>();
+            services.AddSingleton<ITemplateFactory<TemplateContainer, IContentTemplate>, CalculatedFunctionContentTemplateFactory>();
+            services.AddSingleton<CollectionTemplateFactory<IContentTemplate, IContentTemplate>>(
+                sp =>
+                {
+                    var contentTemplateFactories = sp.GetRequiredService<IEnumerable<ITemplateFactory<TemplateContainer, IContentTemplate>>>();
+                    return new CollectionContentTemplateFactory(contentTemplateFactories.ToArray());
```
dustinburson, line 217: If you add the ctor for IEnumerable to CollectionContentTemplateFactory you can avoid this here.

src/lib/Microsoft.Health.Fhir.Ingest.Template/CalculatedFunctionContentTemplateExtensions.cs
```diff
@@ -0,0 +1,51 @@
...
+namespace Microsoft.Health.Fhir.Ingest.Template
+{
+    public static class CalculatedFunctionContentTemplateExtensions
+    {
+        /// <summary>
+        /// Iterates over all expressions belonging to the content template. If the expression language has not been set it will be set
+        /// to default language of the content template.
+        /// </summary>
+        /// <param name="template">The template to retrieve expressions from</param>
+        public static void EnsureExpressionLanguageIsSet(this CalculatedFunctionContentTemplate template)
```
dustinburson, line 18: >void [](http://example.com/codeflow?start=22&length=4)

Suggestion: Return the CalculatedFunctionContentTemplate.  This will support chaining functions together in fluent like style.

src/lib/Microsoft.Health.Fhir.Ingest.Template/CalculatedFunctionContentTemplateExtensions.cs
```diff
@@ -0,0 +1,51 @@
...
+        {
+            EnsureArg.IsNotNull(template, nameof(template));
+
+            foreach (var tuple in template.GetExpressions())
+            {
+                if (tuple.expression != null)
+                {
+                    tuple.expression.Language ??= template.DefaultExpressionLanguage;
```
dustinburson, line 26: >tuple.expression.Language ??= template.DefaultExpressionLanguage; [](http://example.com/codeflow?start=20&length=65)

Is there a way to control this as part of the definition of TemplateExpression.  Ideally the default would be used automatically if not defined.

src/func/Microsoft.Health.Fhir.Ingest.Host/IomtConnectorFunctions.cs
```diff
@@ -25,10 +25,17 @@ namespace Microsoft.Health.Fhir.Ingest.Service
     public class IomtConnectorFunctions
     {
         private readonly ITelemetryLogger _logger;
+        private readonly CollectionContentTemplateFactory _collectionContentTemplateFactory;
```
dustinburson, line 28: >CollectionContentTemplateFactory [](http://example.com/codeflow?start=25&length=32)

Have we done an e2e test to verify the data flow for the old Azure functions?

src/lib/Microsoft.Health.Fhir.Ingest.Template/CalculatedFunctionContentTemplateExtensions.cs
rogordon01: Interesting question. Since it's optional for a user to specify the language it's left blank during deserialization. I briefly played around with NewtonSoft and the JsonConverter to see if I could set it but I didn't see an easy way to obtain the 'defaultExpressionLanguage' that is set on the root of the ContentTemplate itself. So I choose this extension as an alternative.

src/func/Microsoft.Health.Fhir.Ingest.Host/IomtConnectorFunctions.cs
rogordon01: No I have not, but I will do that

src/lib/Microsoft.Health.Fhir.Ingest.Template/CalculatedFunctionContentTemplateFactory.cs
```diff
@@ -0,0 +1,102 @@
...
+{
+    public class CalculatedFunctionContentTemplateFactory : HandlerProxyTemplateFactory<TemplateContainer, IContentTemplate>
+    {
+        private const string TargetTypeName = "CalculatedContent";
+        private readonly ITelemetryLogger _logger;
+        private readonly IExpressionEvaluatorFactory _expressionEvaluatorFactory;
+
+        public CalculatedFunctionContentTemplateFactory(
+            IExpressionEvaluatorFactory expressionEvaluatorFactory,
+            ITelemetryLogger logger)
```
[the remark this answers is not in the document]
rogordon01: I've actually used some of the logs produced by the logger during testing. Could we keep it as-is and update it when we  tackle the backlog item for log standardization?

src/lib/Microsoft.Health.Fhir.Ingest.Template/CalculatedFunctionContentTemplateExtensions.cs
dustinburson: Okay thanks for checking!

src/lib/Microsoft.Health.Fhir.Ingest.Template/CalculatedFunctionContentTemplateFactory.cs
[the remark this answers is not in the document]
dustinburson: Okay, that works if you found the added logging useful :)

dustinburson (commented):
Changes look really good.  Before completing:

Please do an E2E scenario using the old Azure Function approach to ensure no regressions.
A few suggestions to improve the DI.

src/func/Microsoft.Health.Fhir.Ingest.Host/IomtConnectorFunctions.cs
rogordon01: Deployed Azure Functions using `Create-IomtFhirSandboxEnvironment` script. Submitted events to event hub and noted that NormalizationProcessor successfully processed the events. I didn't notice the events flowing into the FhirConverter Function, or into the NormalizedData event hub. I don't see any errors, but will continue to explore this.


image

rogordon01: Turned out to be a misconfigured device content template. Now data flows from normalization -> fhirconversion -> fhir

image

image

PR #116, opened 2021-06-24
Align Blood Pressure typeNames

SharonHart:
'typeName' attribute was misaligned between the sandbox devicecontent.json and fhirmapping.json (bp vs bloddpressure).
As a result, bloodpressure data wasn't propagated to the fhir server.
Replaced the usage of 'bp' in favor or 'bloodpressure', in production code, as well as in tests for the sake of conformity.

### DIFF
--- a/sample/templates/sandbox/devicecontent.json
+++ b/sample/templates/sandbox/devicecontent.json
@@ -64,7 +64,7 @@
     {
       "templateType": "IotCentralJsonPathContent",
       "template": {
-        "typeName": "bp",
+        "typeName": "bloodpressure",
         "typeMatchExpression": "$..[?(@telemetry.BloodPressure.Systolic && @telemetry.BloodPressure.Diastolic)]",
         "patientIdExpression": "$.deviceId",
         "values": [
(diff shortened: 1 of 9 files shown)


pallar-ms:
Oops, didn't realize when I was updating the device content that there was this mismatch since it was always "bp" to start with. Thanks for catching and fixing!
Just curious though (since I'm still fairly new to all this), alternatively would changing the typename in fhirmapping to "bp" fix this, or are there more changes with that route?

pallar-ms:
Thanks for catching and fixing! I'm in favour of elaborating the name to "bloodpressure" to be more explicit, but just curious if alternatively would changing the type name in fhirmapping to "bp" result in the same fix, or are there more changes with that route?

dustinburson (approved):

dustinburson:
> Thanks for catching and fixing! I'm in favour of elaborating the name to "bloodpressure" to be more explicit, but just curious if alternatively would changing the type name in fhirmapping to "bp" result in the same fix, or are there more changes with that route?

It would.  It would also have the benefit of working with any normalized data already enqueued with the bp type.

PR #120, opened 2021-07-09
Replacing System.AggregateException with custom aggregate exception

rogordon01:
When a `System.AggregateException` contains more than 10 inner exceptions, the AI Telemetry Pipeline [converts](https://github.com/microsoft/ApplicationInsights-dotnet/blob/81af973ca106c01b449047a193e03bb5e682fea5/BASE/src/Microsoft.ApplicationInsights/DataContracts/ExceptionTelemetry.cs#L403) this into a custom, internal exception ([InnerExceptionCountExceededException](https://github.com/microsoft/ApplicationInsights-dotnet/blob/master/BASE/src/Microsoft.ApplicationInsights/DataContracts/ExceptionTelemetry.cs#L434)). This results in the following:

- The code truncates the inner exceptions to 10
- A bug within the logic prevents the `Message` attribute of the surrounding `ExceptionTelemety` object from getting populated. 

This results in no actionable message being shown to a developer/customer. Even if there was one, it would be constructed with only the first 10 exceptions.

This PR creates a new Aggregate Exception which will not be converted by the current AI logic above. This will result in the expected aggregate message being displayed to the customer.


rogordon01:
Will also file the discovered bug with the AI Telemetry team

src/lib/Microsoft.Health.Common/SimpleAggregateException.cs
```diff
@@ -0,0 +1,54 @@
...
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+using EnsureThat;
+
+namespace Microsoft.Health.Common
+{
+    public class SimpleAggregateException : Exception
```
dustinburson, line 13: Would be good to double check if we don't have any thing in the health care shared components that does something similar.  If not this might be a good time to add a dependency to that library and add this there so it can be reused, https://github.com/microsoft/healthcare-shared-components

src/lib/Microsoft.Health.Common/Service/ParallelTaskWorker.cs
```diff
@@ -66,7 +66,7 @@ await consumer.Completion
                     {
                         if (!exceptions.IsEmpty)
                         {
-                            throw new AggregateException(exceptions);
+                            throw new SimpleAggregateException(exceptions);
```
dustinburson, line 69: Another alternative is we log all the individual exceptions here.  That still might be a good thing to do even with this change to ensure we have discrete messages for every error for readability.

dustinburson: This will also ensure each exception's stack trace is recorded which we are currently missing.

rogordon01: I had considered doing that but wasn't sure if we wanted to report 1 exception per event or 1 per batch like we do now. I'm fine with reporting each exception in order to gain the stack trace and removing this new exception class as it won't be used

src/lib/Microsoft.Health.Common/SimpleAggregateException.cs
rogordon01: There is nothing similar in the shared components. But since we'll be logging the individual exceptions anyway this class doesn't seem to be needed.

src/lib/Microsoft.Health.Common/Service/ParallelTaskWorker.cs
rogordon01: Done

namalu (approved):

PR #123, opened 2021-07-09
Use Polly to perform retries on failed event processing

rogordon01:
Currently, when an event incurs an error the entire batch is retried forever. This PR utilizes `Polly` to perform the retry logic. Only a specific set of exceptions will be allowed to retry.

### DIFF
--- a/src/console/MeasurementCollectionToFhir/Processor.cs
+++ b/src/console/MeasurementCollectionToFhir/Processor.cs
@@ -58,7 +58,8 @@ namespace Microsoft.Health.Fhir.Ingest.Console.MeasurementCollectionToFhir
         {
             bool ExceptionRetryableFilter(Exception ee)
             {
-                logger.LogError(new Exception("Encountered retryable exception", ee));
+                logger.LogTrace($"Encountered retryable exception {ee.GetType()}");
+                logger.LogError(ee);
                 TrackExceptionMetric(ee, logger);
                 return true;
             }
--- a/src/console/MeasurementCollectionToFhir/Processor.cs
+++ b/src/console/MeasurementCollectionToFhir/Processor.cs
@@ -34,17 +34,16 @@ namespace Microsoft.Health.Fhir.Ingest.Console.MeasurementCollectionToFhir
             [MeasurementFhirImport] MeasurementFhirImportService measurementImportService,
             ITelemetryLogger logger)
         {
-            _templateDefinition = templateDefinition;
-            _templateManager = templateManager;
-            _measurementImportService = measurementImportService;
-            _logger = logger;
+            _templateDefinition = EnsureArg.IsNotNullOrWhiteSpace(templateDefinition, nameof(templateDefinition));
+            _templateManager = EnsureArg.IsNotNull(templateManager, nameof(templateManager));
+            _measurementImportService = EnsureArg.IsNotNull(measurementImportService, nameof(measurementImportService));
+            _logger = EnsureArg.IsNotNull(logger, nameof(logger));
             _retryPolicy = CreateRetryPolicy(logger);
         }
 
         public async Task ConsumeAsync(IEnumerable<IEventMessage> events)
         {
             EnsureArg.IsNotNull(events);
-            EnsureArg.IsNotNull(_templateDefinition);
 
             await _retryPolicy.ExecuteAsync(async () => await ConsumeAsyncImpl(events, _templateManager.GetTemplateAsString(_templateDefinition)));
         }
--- a/src/console/MeasurementCollectionToFhir/Processor.cs
+++ b/src/console/MeasurementCollectionToFhir/Processor.cs
@@ -3,17 +3,20 @@
 // Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // -------------------------------------------------------------------------------------------------
 
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Threading.Tasks;
 using EnsureThat;
 using Microsoft.Azure.WebJobs;
+using Microsoft.Health.Common.Telemetry;
 using Microsoft.Health.Events.EventConsumers;
 using Microsoft.Health.Events.Model;
 using Microsoft.Health.Fhir.Ingest.Console.Template;
 using Microsoft.Health.Fhir.Ingest.Host;
 using Microsoft.Health.Fhir.Ingest.Service;
 using Microsoft.Health.Logging.Telemetry;
+using Polly;
 
 namespace Microsoft.Health.Fhir.Ingest.Console.MeasurementCollectionToFhir
 {
@@ -23,6 +26,7 @@ namespace Microsoft.Health.Fhir.Ingest.Console.MeasurementCollectionToFhir
         private MeasurementFhirImportService _measurementImportService;
         private string _templateDefinition;
         private ITelemetryLogger _logger;
+        private AsyncPolicy _retryPolicy;
 
         public Processor(
             [Blob("template/%Template:FhirMapping%", FileAccess.Read)] string templateDefinition,
@@ -34,6 +38,7 @@ namespace Microsoft.Health.Fhir.Ingest.Console.MeasurementCollectionToFhir
             _templateManager = templateManager;
             _measurementImportService = measurementImportService;
             _logger = logger;
+            _retryPolicy = CreateRetryPolicy(logger);
         }
 
         public async Task ConsumeAsync(IEnumerable<IEventMessage> events)
@@ -41,8 +46,42 @@ namespace Microsoft.Health.Fhir.Ingest.Console.MeasurementCollectionToFhir
             EnsureArg.IsNotNull(events);
             EnsureArg.IsNotNull(_templateDefinition);
 
-            var templateContent = _templateManager.GetTemplateAsString(_templateDefinition);
+            await _retryPolicy.ExecuteAsync(async () => await ConsumeAsyncImpl(events, _templateManager.GetTemplateAsString(_templateDefinition)));
+        }
+
+        private async Task ConsumeAsyncImpl(IEnumerable<IEventMessage> events, string templateContent)
+        {
             await _measurementImportService.ProcessEventsAsync(events, templateContent, _logger).ConfigureAwait(false);
+        }
+
+        private static AsyncPolicy CreateRetryPolicy(ITelemetryLogger logger)
+        {
+            bool ExceptionRetryableFilter(Exception ee)
+            {
+                logger.LogError(new Exception("Encountered retryable exception", ee));
+                TrackExceptionMetric(ee, logger);
+                return true;
+            }
+
+            return Policy
+                .Handle<Exception>(ExceptionRetryableFilter)
+                .WaitAndRetryForeverAsync(retryCount => TimeSpan.FromSeconds(Math.Min(30, Math.Pow(2, retryCount))));
+        }
+
+        private static void TrackExceptionMetric(Exception exception, ITelemetryLogger logger)
+        {
+            var type = exception.GetType().ToString();
+            var ToMetric = new Metric(
+                type,
+                new Dictionary<string, object>
+                {
+                    { DimensionNames.Name, type },
+                    { DimensionNames.Category, Category.Errors },
+                    { DimensionNames.ErrorType, ErrorType.GeneralError },
+                    { DimensionNames.ErrorSeverity, ErrorSeverity.Warning },
+                    { DimensionNames.Operation, ConnectorOperation.FHIRConversion},
+                });
+            logger.LogMetric(ToMetric, 1);
         }
     }
 }
--- a/src/console/Normalize/Processor.cs
+++ b/src/console/Normalize/Processor.cs
@@ -114,19 +114,6 @@ namespace Microsoft.Health.Fhir.Ingest.Console.Normalize
         {
             bool ExceptionRetryableFilter(Exception ee)
             {
-                var type = ee.GetType().ToString();
-                var ToMetric = new Metric(
-                type,
-                new Dictionary<string, object>
-                {
-                        { DimensionNames.Name, type },
-                        { DimensionNames.Category, Category.Errors },
-                        { DimensionNames.ErrorType, ErrorType.DeviceMessageError },
-                        { DimensionNames.ErrorSeverity, ErrorSeverity.Warning },
-                        { DimensionNames.Operation, ConnectorOperation.Normalization},
-                });
-                logger.LogMetric(ToMetric, 1);
-
                 switch (ee)
                 {
                     case AggregateException ae when ae.InnerExceptions.Any(ExceptionRetryableFilter):
@@ -137,16 +124,34 @@ namespace Microsoft.Health.Fhir.Ingest.Console.Normalize
                     case RequestFailedException _:
                         break;
                     default:
+                        TrackExceptionMetric(ee, logger);
                         return false;
                 }
 
                 logger.LogError(new Exception("Encountered retryable exception", ee));
+                TrackExceptionMetric(ee, logger);
                 return true;
             }
 
             return Policy
                 .Handle<Exception>(ExceptionRetryableFilter)
                 .WaitAndRetryForeverAsync(retryCount => TimeSpan.FromSeconds(Math.Min(30, Math.Pow(2, retryCount))));
+        }
+
+        private static void TrackExceptionMetric(Exception exception, ITelemetryLogger logger)
+        {
+            var type = exception.GetType().ToString();
+            var ToMetric = new Metric(
+            type,
+            new Dictionary<string, object>
+            {
+                { DimensionNames.Name, type },
+                { DimensionNames.Category, Category.Errors },
+                { DimensionNames.ErrorType, ErrorType.DeviceMessageError },
+                { DimensionNames.ErrorSeverity, ErrorSeverity.Warning },
+                { DimensionNames.Operation, ConnectorOperation.Normalization},
+            });
+            logger.LogMetric(ToMetric, 1);
         }
     }
 }
--- a/src/console/Normalize/Processor.cs
+++ b/src/console/Normalize/Processor.cs
@@ -129,7 +129,8 @@ namespace Microsoft.Health.Fhir.Ingest.Console.Normalize
                         return false;
                 }
 
-                logger.LogError(new Exception("Encountered retryable exception", ee));
+                logger.LogTrace($"Encountered retryable exception {ee.GetType()}");
+                logger.LogError(ee);
                 TrackExceptionMetric(ee, logger);
                 return true;
             }
--- a/src/console/Normalize/Processor.cs
+++ b/src/console/Normalize/Processor.cs
@@ -58,14 +58,13 @@ namespace Microsoft.Health.Fhir.Ingest.Console.Normalize
         public async Task ConsumeAsync(IEnumerable<IEventMessage> events)
         {
             EnsureArg.IsNotNull(events);
-            await _retryPolicy.ExecuteAsync(async () => await ConsumeAsyncImpl(events));
+            EnsureArg.IsNotNull(_templateDefinition);
+
+            await _retryPolicy.ExecuteAsync(async () => await ConsumeAsyncImpl(events, _templateManager.GetTemplateAsString(_templateDefinition)));
         }
 
-        private async Task ConsumeAsyncImpl(IEnumerable<IEventMessage> events)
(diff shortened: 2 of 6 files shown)


src/lib/Microsoft.Health.Events/EventConsumers/Service/EventConsumerService.cs
```diff
@@ -35,44 +50,46 @@ public async Task ConsumeEvents(IEnumerable<IEventMessage> events)
...
-                catch (Exception e)
-#pragma warning restore CA1031
+                switch (ee)
                 {
-                    _logger.LogError(e);
+                    case AggregateException ae when ae.InnerExceptions.All(ExceptionRetryableFilter):
+                    case OperationCanceledException _:
+                    case HttpRequestException _:
+                    case EventHubsException _:
+                    case RequestFailedException _:
```
rogordon01, line 80: These should retry on all Http, EventHub and BlobStorage related exceptions.

src/lib/Microsoft.Health.Events/EventConsumers/Service/EventConsumerService.cs
```diff
@@ -6,24 +6,40 @@
...
+using Azure.Identity;
+using Azure.Messaging.EventHubs;
+using EnsureThat;
 using Microsoft.Health.Events.Model;
 using Microsoft.Health.Logging.Telemetry;
+using Polly;
 
 namespace Microsoft.Health.Events.EventConsumers.Service
 {
     public class EventConsumerService : IEventConsumerService
```
dustinburson, line 21: This also impacts the Normalization to FHIR logic.  That has a separate mechanism for non-recoverable errors.  This would fail to retry on 429 for example due to throttling from the FHIR service.

src/lib/Microsoft.Health.Events/EventConsumers/Service/EventConsumerService.cs
```diff
@@ -35,44 +51,42 @@ public async Task ConsumeEvents(IEnumerable<IEventMessage> events)
...
+                    case HttpRequestException _:
+                    case EventHubsException _:
+                    case AuthenticationFailedException _:
+                    case RequestFailedException _:
+                        break;
+                    default:
+                        return false;
                 }
+
+                logger.LogError(new Exception("Encountered retryable exception", ee));
```
dustinburson, line 83: We should emit telemetry here.

src/lib/Microsoft.Health.Events/EventConsumers/Service/EventConsumerService.cs
dustinburson: To avoid breaking the FHIR conversion logic we should make this optional.

dustinburson (changes requested):
Please make option to avoid impacting Normalization to FHIR and add telemetry.

src/lib/Microsoft.Health.Events/EventConsumers/Service/EventConsumerService.cs
rogordon01: Right now this would log the exception using the AI SDK, which would get recored as both an Exception Metric and Exception record. Are you referring to logging a different metric? If so, what metric would you like to see?

src/lib/Microsoft.Health.Events/EventConsumers/Service/EventConsumerService.cs
rogordon01: We currently retry on `HttpExceptions`. Wouldn't that handle the 429 scenario?

rogordon01: I see the custom exception handling now: https://github.com/microsoft/iomt-fhir/blob/master/src/lib/Microsoft.Health.Fhir.Ingest/Telemetry/ExceptionTelemetryProcessor.cs. Let me update this to make it optional.

dustinburson: I don't know if it comes back as an HTTP request exception or some other type.  The FHIR logic takes the opposite approach, we retry all exceptions except the ones we explicitly capture and emit telemetry for.  With out further analysis I am concerned this could introduce unintended side effects for the FHIR conversion. For now I think it would be safe to only apply this during normalization and we can have a work item to unify methods if appropriate in the future.

src/lib/Microsoft.Health.Events/EventConsumers/Service/EventConsumerService.cs
dustinburson: For FHIR conversion we have explicit error codes for various non-recoverable errors.  See https://github.com/microsoft/iomt-fhir/blob/master/src/lib/Microsoft.Health.Fhir.Ingest/Telemetry/ExceptionTelemetryProcessor.cs and here is an example of a metric that emitted by one these exceptions https://github.com/microsoft/iomt-fhir/blob/master/src/lib/Microsoft.Health.Fhir.Ingest/Service/PatientDeviceMismatchException.cs

For this work item it is too large a scope to classify the various template errors we can see the same way but it would be good to make sure all the errors are getting tagged as general normalization errors.  If that is occurring upstream, great but I suspect it isn't.  Will may be able to help with more details since he did a lot of the wiring here to support Azure Monitor.

rogordon01: I think that is fine, but agree that is out of scope for this task. When exceptions are thrown during Normalization they should wrapped in a `NormalizationException` and perhaps even implement `ITelemetryFormattable` to be turned into a metric. 
I can create a backlog item to capture that work.

dustinburson: For clarification, the ask for this PR is ensure the metrics for errors thrown here have the following attributes:
``
        public Metric ToMetric => new Metric(
            "NormalizationException",
            new Dictionary<string, object>
            {
                { DimensionNames.Name, "NormalizationException" },
                { DimensionNames.Category, Category.Errors },
                { DimensionNames.ErrorType, ErrorType.DeviceMessageError },
                { DimensionNames.ErrorSeverity, ErrorSeverity.Warning },
                { DimensionNames.Operation, ConnectorOperation.Normalization},
            });
``

This should at least distinguish these errors in custom facing telemetry such that they can identify there is an issue with the device messages.  The exact nature will not be known, but that is what the follow up work item will seek to address.

src/lib/Microsoft.Health.Events/EventConsumers/Service/EventConsumerService.cs
```diff
@@ -35,44 +51,42 @@ public async Task ConsumeEvents(IEnumerable<IEventMessage> events)
...
-
-                    await TryOperationAsync(eventConsumer, events).ConfigureAwait(false);
-                    break;
-                }
-#pragma warning disable CA1031
-                catch (Exception e)
-#pragma warning restore CA1031
-                {
-                    _logger.LogError(e);
+                    case AggregateException ae when ae.InnerExceptions.All(ExceptionRetryableFilter):
```
dustinburson, line 72: Should this be Any()? Basically in a mixed exception setting (some retryable some fatal) this will be cause the retryable exceptions to be skipped.  Shouldn't we retry the whole batch unless only fatal errors remain then bubble up?

rogordon01: Yea that makes sense. I'll change this.

src/lib/Microsoft.Health.Events/EventConsumers/Service/EventConsumerService.cs
rogordon01: Spoke with @dustinburson  offline. We'll add in a custom ExceptionTelemetry processor to this class. The Normalization Service can use that to produce custom telemetry for each processed exception.

src/console/Startup.cs
```diff
@@ -125,7 +126,30 @@ public virtual IEventConsumerService ResolveEventConsumerService(IServiceProvide
         {
```
namalu, line 126: With the current changes I have in workspace-platform, this won't actually fix the issue. The workspace platform changes moves away from calling these "Resolve*" methods and instead just adds objects to the ServiceProvider, so it can handle the resolution.

We can definitely just add code there to to fix it, but it's important that we remember to do it, or the bug will still be there in Prod.

src/lib/Microsoft.Health.Events/EventConsumers/Service/EventConsumerService.cs
```diff
@@ -35,44 +48,52 @@ public async Task ConsumeEvents(IEnumerable<IEventMessage> events)
             {
                 foreach (IEventConsumer eventConsumer in _eventConsumers)
                 {
-                    await OperationWithRetryAsync(eventConsumer, events);
+                    try
+                    {
+                        await RetryPolicy.ExecuteAsync(async () => await eventConsumer.ConsumeAsync(events).ConfigureAwait(false));
```
namalu, line 53: I wonder if it makes more sense to move this code to the actual event consumer - In this case it would be the Microsoft.Health.Fhir.Ingest.Console.Normalize.Processor class, specifically in the ConsumeAsync() method. It appears that, considering the proposed changes in Startup.cs, we only want these changes in the normalization processor. Moving the retry logic there would encapsulate it to just normalization, and would mean we would not need to update the workspace-platform code.

namalu (changes requested):

src/lib/Microsoft.Health.Events/EventConsumers/Service/EventConsumerService.cs
rogordon01: Great point; I'll go ahead and do that.

rogordon01: Done

src/console/Startup.cs
rogordon01: Moved retry logic into Normalization Processor and reverted this back to its original form.

src/lib/Microsoft.Health.Events/EventConsumers/Service/EventConsumerService.cs
```diff
@@ -35,44 +35,16 @@ public async Task ConsumeEvents(IEnumerable<IEventMessage> events)
             {
                 foreach (IEventConsumer eventConsumer in _eventConsumers)
                 {
-                    await OperationWithRetryAsync(eventConsumer, events);
-                }
-            }
-        }
-
-        private async Task OperationWithRetryAsync(IEventConsumer eventConsumer, IEnumerable<IEventMessage> events)
```
dustinburson, line 43: >OperationWithRetryAsync [](http://example.com/codeflow?start=27&length=23)

Can we add this back in and make it optional in ctor?  Right now this will be break the FHIR Conversions.  Retryable errors like 429's or unable to auth to the FHIR service will error and cause loss of data.

dustinburson (changes requested):
The changes in the normalize processor look good.  The ask is to keep the existing retry logic in the EventConsumerService but make it optional and off for normalization and on for FHIR conversion (or you could look at moving it to the processor for the FHIR converter).

src/console/Normalize/Processor.cs
```diff
@@ -90,5 +109,49 @@ public async Task ConsumeAsync(IEnumerable<IEventMessage> events)
...
+                .Handle<Exception>(ExceptionRetryableFilter)
+                .WaitAndRetryForeverAsync(retryCount => TimeSpan.FromSeconds(Math.Min(30, Math.Pow(2, retryCount))));
+        }
+
+        private static void TrackExceptionMetric(Exception exception, ITelemetryLogger logger)
+        {
+            var type = exception.GetType().ToString();
+            var ToMetric = new Metric(
+            type,
+            new Dictionary<string, object>
```
dustinburson, line 146: nit: looks like indent is missing

src/lib/Microsoft.Health.Events/EventConsumers/Service/EventConsumerService.cs
rogordon01: This would cause the FHIR Converter to retry forever on all exceptions. I'm assuming that is desired, since the Fhir Converter has custom exception handling logic? If thats the case I'll just move this into the Fhir Converter Processor as per @namalu comment moving it into the Startup.cs will be tricky given his refactor

dustinburson: The FHIR converter doesn't retry forever on all exceptions.  It uses the [https://github.com/microsoft/iomt-fhir/blob/master/src/lib/Microsoft.Health.Fhir.Ingest/Telemetry/ExceptionTelemetryProcessor.cs](https://github.com/microsoft/iomt-fhir/blob/master/src/lib/Microsoft.Health.Fhir.Ingest/Telemetry/ExceptionTelemetryProcessor.cs) so known exceptions are "handled" and converted to telemetry that is emitted.  Basically any exception that makes it this far will cause it to retry indefinititely but that is intended.  Non-recoverable errors due to bad data are supposed to be caught and handled in the ExceptionTelemetryProcessor.

rogordon01: Yup; sorry that is what I had meant; it'll retry forever on all exceptions not handled by that custom exception logic.

src/console/Normalize/Processor.cs
```diff
@@ -90,5 +109,49 @@ public async Task ConsumeAsync(IEnumerable<IEventMessage> events)
...
+            bool ExceptionRetryableFilter(Exception ee)
+            {
+                switch (ee)
+                {
+                    case AggregateException ae when ae.InnerExceptions.Any(ExceptionRetryableFilter):
+                    case OperationCanceledException _:
+                    case HttpRequestException _:
+                    case EventHubsException _:
+                    case AuthenticationFailedException _:
+                    case RequestFailedException _:
```
wi-y, line 124: Since we are calling the Azure.Messaging.EventHubs.Producer.EventHubProducerClient downstream, I believe the Producer will throw an Azure.Messaging.EventHubs.EventHubsException instead of a Microsoft.Azure.EventHubs.EventHubsException (legacy).

I think this should be updated to reference the new SDK exception (or both). Feel free to double check. I testing by changing my Normalization Event Hub to be something bogus in order to log an error

rogordon01: Thanks for pointing that out. Does this entire class need to updated to use the new API?

rogordon01: Actually I see that other interfaces we created use the old API. Its a bit confusing... are their plans to update the project to use a single API?

wi-y: It would be good to transition to the new one, but that should probably be its own work item. It may be harder than just doing a find and replace.

wi-y: Question - When an EventHubsException occurs would we expect `ExceptionRetryableFilter` to return `true` and retry?

For some reason for me when I intentionally log an EventHubsException the code seems to still hit the default block in that switch statement and will return false. I am wondering if it is because it is a `System.Exception {Azure.Messaging.EventHubs.EventHubsException}` or something else with the switch statement that is going on, or if I am misunderstanding how `ExceptionRetryableFilter` works. I was thinking `ExceptionRetryableFilter` would hit the block of code at the at the end that returns true so that it retries.

wi-y: Nevermind - it works fine when pulling in the latest changes.

rogordon01:
Yeah I changed the exception to that of the new SDK per your suggestion.

Get Outlook for iOS<https://aka.ms/o0ukef>

src/console/MeasurementCollectionToFhir/Processor.cs
```diff
@@ -34,15 +38,50 @@ public class Processor : IEventConsumer
...
+        private async Task ConsumeAsyncImpl(IEnumerable<IEventMessage> events, string templateContent)
+        {
             await _measurementImportService.ProcessEventsAsync(events, templateContent, _logger).ConfigureAwait(false);
         }
+
+        private static AsyncPolicy CreateRetryPolicy(ITelemetryLogger logger)
+        {
+            bool ExceptionRetryableFilter(Exception ee)
+            {
+                logger.LogError(new Exception("Encountered retryable exception", ee));
```
dustinburson, line 61: >logger.LogError(new Exception("Encountered retryable exception", ee)); [](http://example.com/codeflow?start=16&length=70)

Would suggest just logging the original exception. We only get the first inner exception in the logs and if the original exception is already wrapped we will mis the details.  I like the message though, can we switch the message to info? i.e. "Encountered retryable exception {0}", ee.GetType()

src/console/MeasurementCollectionToFhir/Processor.cs
```diff
@@ -34,15 +38,50 @@ public class Processor : IEventConsumer
...
+                TrackExceptionMetric(ee, logger);
+                return true;
+            }
+
+            return Policy
+                .Handle<Exception>(ExceptionRetryableFilter)
+                .WaitAndRetryForeverAsync(retryCount => TimeSpan.FromSeconds(Math.Min(30, Math.Pow(2, retryCount))));
```
dustinburson, line 68: >30 [](http://example.com/codeflow?start=86&length=2)

30 seconds may be too long for an initial retry. Would suggest 5 to 10.

dustinburson, line 68: > Math.Pow(2, retryCount) [](http://example.com/codeflow?start=89&length=24)

We will probably want to cap the max retry time.  Say 10 minutes?

dustinburson (approved):
Looks good.  Added some suggestions.

src/console/MeasurementCollectionToFhir/Processor.cs
rogordon01: The time between retries will be the lower of 30 seconds and the Math.Pow function. Since the Pow function is based on the retry count, the initial retry should be 2 ^ 1 = 2 seconds.

rogordon01: Or do you mean that we should lower the maximum time between all retries to 10 seconds from 30?

src/console/MeasurementCollectionToFhir/Processor.cs
rogordon01: Will do

src/console/MeasurementCollectionToFhir/Processor.cs
rogordon01: Just to confirm, you're saying that we no longer should retry forever, as we were doing before, and instead limit the overall retry operation to 10 minutes?

dustinburson: No sorry, the intent is to cap the timespan for exponential back off.  I misread the code, I think we are covered.  Based on your above comment the longest we will wait for a retry is 30 seconds.

dustinburson: I misread the code line.  Thanks for the clarification.  I think we are good here.

src/console/MeasurementCollectionToFhir/Processor.cs
```diff
@@ -34,15 +38,51 @@ public class Processor : IEventConsumer
             _templateManager = templateManager;
             _measurementImportService = measurementImportService;
             _logger = logger;
+            _retryPolicy = CreateRetryPolicy(logger);
         }
 
         public async Task ConsumeAsync(IEnumerable<IEventMessage> events)
         {
             EnsureArg.IsNotNull(events);
             EnsureArg.IsNotNull(_templateDefinition);
```
namalu, line 47: If the _templateDefinition is null, I think we will end up with the 100% cpu issue here. I think if you do something similar to what you did in the Normalize Processor, that would address this issue.

namalu (approved):

src/console/MeasurementCollectionToFhir/Processor.cs
rogordon01: The EnsureArg is called outside of the retryPolicy.ExecutAsync, so throwing an exception here shouldn't cause us to retry forever. 
I do, however, see the EnsureArg check being thrown inside of the retryPolicy.ExecuteAsync in the Normalization Processor. I'll adjust that

rogordon01: @wi-y  Any particular reason why we check the validity of the `_templateDefinition` outside of the constructor? From the looks of it, If its null/empty there is never an opportunity to update it.

PR #124, opened 2021-07-15
Creating documentation for Calculated Functions

rogordon01:
This PR updates the documentation to detail the use of the new Expression based template. In addition it also details how to make use of Calculated Functions and describes the set of custom functions created as part of this project.


docs/Configuration.md
```diff
@@ -244,227 +395,151 @@ The JsonPathContentTemplate allows matching on and extracting values from an Eve
...
             }
         ]
     }
 }
 ```
+
 ---
 
-**Blood Pressure**
+**Select and tranform incoming data**
```
rogordon01, line 467: @namalu @dustinburson I believe that this example is supported by our system. Here we have two templates with the same `typeName` producing the same values, just in different ways and using different `typeMatches`. Looking at the code this should produce different `Measurements` and they'd be aggregated on the fhir transformation correctly. Is my understanding correct?

docs/CustomFunctions.md
```diff
@@ -0,0 +1,132 @@
...
+| {"left" : 0, "right" : 50}    | pow(left, right) | 0                          |
+| {"left" : 100, "right" : 0.5} | pow(left, right) | 10                         |
+
+### subtract
+
+```
+number subtract(number $left, number $right)
+```
+
+Returns the result of subtracting the left argument from the right.
```
namalu, line 89: Based on the table below it looks like we are actually subtracting the right argument **from** the left.

docs/CustomFunctions.md
```diff
@@ -0,0 +1,132 @@
...
+
+## String Functions
+
+### fromUnixTimestamp
+
+```
+string fromUnixTimestamp(number $unixTimestamp)
+```
+
+Produces an [ISO 8061](https://en.wikipedia.org/wiki/ISO_8601) compliant time stamp from the given Unix timestamp.
```
namalu: I gave feedback on your other PR about Unix timestamps being both seconds and milliseconds from reference. We should update the documentation when we decide on the change.

docs/Legacy-Configuration.md
```diff
@@ -0,0 +1,420 @@
+## Mapping with JSON Path
+
+The two device content template types supported today rely on JSON Path to both match the desired template and extract values.  
+More information on JSON Path can be found [here](https://goessner.net/articles/JsonPath/). Both template types use the [JSON .NET implementation](https://www.newtonsoft.com/json/help/html/QueryJsonSelectTokenJsonPath.htm) for resolving JSON Path expressions. Additional examples can be found in the [unit tests](../test/Microsoft.Health.Fhir.Ingest.UnitTests/Template/JsonPathContentTemplateTests.cs).
```
namalu, line 4: Now that we have more than 2 device content templates, we probably need to update the first sentence - It looks like we have 3 defined in this file and 1 new type that does not 'rely' on JSON Path, but can use it.

docs/Legacy-Configuration.md
```diff
@@ -0,0 +1,420 @@
...
+
```
namalu, line 5: unit tests is a dead link.

namalu (changes requested):

docs/Configuration.md
```diff
@@ -44,27 +44,178 @@ Below is a conceptual example of what happens during normalization.
...
+    {
+      "systolic": "122",
+      "diastolic": "82",
+      "date": "2021-07-13T17:28:01.061122Z"
+    }
+  ]
+}
+```
+
+_Template_
```
rbhaiya, line 135: nit: please change "Template" to "Template with a calculation"

docs/Configuration.md
```diff
@@ -44,27 +44,178 @@ Below is a conceptual example of what happens during normalization.
...
+
+```
+"templateType": "CalculatedContentTemplate",
+    "template": {
+        "typeName": "heartrate",
+        "typeMatchExpression": "$..[?(@systolic && @diastolic)]",
+        ...
```
rbhaiya, line 142: Recommend having a whole JSON content. as most of the customers like to copy and paste it while testing.

docs/Legacy-Configuration.md
```diff
@@ -0,0 +1,420 @@
...
+        ]
+    }
+}
+```
+
+### **IotJsonPathContentTemplate**
+
+The IotJsonPathContentTemplate is similar to the JsonPathContentTemplate except the DeviceIdExpression and TimestampExpression are not required.
+
+The assumption when using this template is the messages being evaluated were sent using the [Azure IoT Hub Device SDKs](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-sdks#azure-iot-hub-device-sdks) or [Export Data (legacy)](https://docs.microsoft.com/en-us/azure/iot-central/core/howto-export-data-legacy) feature of [Azure IoT Central](https://docs.microsoft.com/en-us/azure/iot-central/core/howto-export-data). When using these SDKs the device identity (assuming the device id from Iot Hub/Central is registered as an identifer for a device resource on the destination FHIR server) is known as well as the timestamp of the message.  If you are using Azure IoT Hub Device SDKs but are using custom properties in the message body for the device identity or measurement timestamp you can still use the JsonPathContentTemplate.
```
rbhaiya, line 222: nit: Please change "Iot Hub" to "IoT Hub"

rbhaiya (commented):
Agree with Nates comments, and added a couple additional ones

docs/Configuration.md
```diff
@@ -44,27 +44,178 @@ Below is a conceptual example of what happens during normalization.
...
+    }
+```
+
+Two matches will be extracted using the above expression and used to create JTokens. Subsequent expressions will be evaluated using the following JTokens:
```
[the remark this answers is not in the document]
rogordon01: This is being done so users can access the parent scope and the item returned by the `typeMatchExpression` within subsequent expressions. For example:

```json
{
"deviceId" : 1234,
    {
        "systolic" : "abc",
        "diastolic": "def"
    }
}
```

A `typeMatchExpression` of `$..[?(@systolic && @diastolic)]` would normally return:

```json
{
        "systolic" : "abc",
        "diastolic": "def"
 }
```json

We wouldn't be able to access `deviceId` in the `deviceIdExpression` as it is out of scope. Now we are returning:

{
   "deviceId" : 1234,
    {
          "systolic" : "abc",
          "diastolic": "def"
    },
   "matchedToken" : {
          "systolic" : "abc",
          "diastolic": "def"
   }
}

And the `deviceId` can now be accessed.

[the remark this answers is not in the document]
Comment (rogordon01) [reply docs/Configuration.md:146]: For your second question; if the typeMatchExpression matched the root of the json object, yes, the entire object would be added to `matchedToken`.

File: docs/CustomFunctions.md [hunk 10]
```diff
@@ -0,0 +1,132 @@
...
+
+## Mathmatical Functions
+
+### add
+
+```
+number add(number $left, number $right)
+```
+
+Returns the result of adding the left argument from the right.
```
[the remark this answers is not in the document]
rogordon01: Copy and paste is not my friend, haha

docs/CustomFunctions.md
rogordon01: I added an additional function `fromUnixTimestampMs` to support the milliseconds scenario. I've also updated the documentation to reflect this.

docs/Legacy-Configuration.md
rogordon01: Thinking about it more, I think that that I'll create separate pages for each Template Type. I can then link to them from the main Configuration page with a statement of what each one is and why'd you'd want to use it.

docs/Configuration.md
```diff
@@ -44,27 +44,178 @@ Below is a conceptual example of what happens during normalization.
...
+_Template_
+
+```
+"templateType": "CalculatedContentTemplate",
+    "template": {
+        "typeName": "heartrate",
+        "typeMatchExpression": "$..[?(@systolic && @diastolic)]",
+        ...
+    }
+```
```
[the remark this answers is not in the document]
rogordon01: I've updated the sample. Also, I've created a new unit test to test out each sample payload/template. This test will

- Verify the json template can be deserialized into a ContentTemplate 
- Verify the payload can be deserialized
- Verify that the expected Measurements are produced

namalu (approved):

PR #125, opened 2021-07-28
Data Mapper: JSON validation of Device Data Sample

kyclai:
Updated the IoMT Connector Data Mapper as follows:
- Replaced Newtonsoft.Json with System.Text.Json to validate JSON syntax when testing normalization and FHIR transformation. System.Text.Json is used because it follows RFC 8259 specification, which, for example, does not allow trailing commas.
- Use JSON Lint to validate JSON syntax as device data sample is input, and display validation result. JSON Lint is used because it considers trailing commas as invalid and it provides an error message with the line number of the error.
- Use CodeMirror as the editor for device data sample to improve the experience of entering the JSON and of finding JSON validation errors.
- Corrected spelling in Device Mapping tutorial.

### DIFF
--- a/tools/data-mapper/Microsoft.Health.Tools.DataMapper/ClientApp/src/components/mapping/Detail.Widgets.Test.tsx
+++ b/tools/data-mapper/Microsoft.Health.Tools.DataMapper/ClientApp/src/components/mapping/Detail.Widgets.Test.tsx
@@ -5,6 +5,7 @@
 
 import * as React from 'react';
 import { Input, Col, Row } from 'reactstrap';
+import * as JsonLint from 'jsonlint-mod';
 
 import { Mapping } from '../../store/Mapping';
 import TestService from '../../services/TestService';
@@ -12,6 +13,9 @@ import { PlayCircleIcon } from '../Icons';
 import * as Utility from './Utility';
 
 const MappingTestWidget = (props: { data: Mapping }) => {
+    const [dataSampleResult, setDataSampleResult] = React.useState('');
+    const [dataSampleValid, setDataSampleValid] = React.useState(true);
+
     const [normTestResult, setNormTestResult] = React.useState('Normalization test result output...');
     const [normTestResultBadge, setNormTestResultBadge] = React.useState('');
     const [normTestInProgress, setNormTestInProgress] = React.useState(false);
@@ -22,6 +26,19 @@ const MappingTestWidget = (props: { data: Mapping }) => {
 
     const dataSampleRef = React.useRef<HTMLInputElement>() as React.RefObject<HTMLInputElement>;
     const identityResolutionTypeRef = React.useRef<HTMLInputElement>() as React.RefObject<HTMLInputElement>;
+
+    const handleDataSampleChange = () => {
+        const dataSample = dataSampleRef.current?.value;
+        try {
+            JsonLint.parse(dataSample);
+            setDataSampleResult('Valid JSON');
+            setDataSampleValid(true);
+        }
+        catch (err) {
+            setDataSampleResult(err.toString());
+            setDataSampleValid(false);
+        }
+    }
 
     const startNormalizationTest = () => {
         setNormTestInProgress(true);
@@ -105,7 +122,11 @@ const MappingTestWidget = (props: { data: Mapping }) => {
                             <Input
                                 type="textarea" name="devicedatasample" id="devicedatasample" rows={10}
                                 placeholder="Paste your device data sample here..." innerRef={dataSampleRef}
+                                onChange={handleDataSampleChange}
                             />
+                            <pre className={`iomt-cm-data-result overflow-auto p-2 ${dataSampleValid ? "text-success" : "text-danger"}`}>
+                                {dataSampleResult}
+                            </pre>
                         </div>
                     </Col>
                     <Col sm={4}>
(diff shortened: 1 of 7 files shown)


wi-y (approved):

namalu (approved):
This is great. I launched the mapping tool on my machine and was able to test the error condition - I wish the message would just state there is a trailing comma, instead of an expected string, but there is not much we can do about that.

tools/data-mapper/Microsoft.Health.Tools.DataMapper/ClientApp/src/components/mapping/Detail.Widgets.Test.tsx
```diff
@@ -20,12 +31,62 @@ const MappingTestWidget = (props: { data: Mapping }) => {
...
 
-    const dataSampleRef = React.useRef<HTMLInputElement>() as React.RefObject<HTMLInputElement>;
     const identityResolutionTypeRef = React.useRef<HTMLInputElement>() as React.RefObject<HTMLInputElement>;
 
+    var codeEditor: CodeMirror.EditorFromTextArea;
+    var dataSampleErrorLine: number | null = null;
+
+    React.useEffect(() => {
+        codeEditor = CodeMirror.fromTextArea(
+            (document.getElementById('devicedatasample') as HTMLTextAreaElement),
```
ms-teli, line 41: Can this work: https://reactjs.org/docs/refs-and-the-dom.html

kyclai:
> This is great. I launched the mapping tool on my machine and was able to test the error condition - I wish the message would just state there is a trailing comma, instead of an expected string, but there is not much we can do about that.

@namalu - I agree about the message description. If we want to specifically target certain errors (e.g. trailing commas), perhaps we could also try
1. keeping a dictionary of JSON Lint messages to our custom messages that we can use to override (though at the risk of JSON Lint messages being not unique or changing in the future), or
2. writing our own regex to catch these errors and outputting our custom messages.

However, if the user runs the normalization test with a trailing comma, System.Test.Json's error message actually mentions the trailing comma.

tools/data-mapper/Microsoft.Health.Tools.DataMapper/ClientApp/src/components/mapping/Detail.Widgets.Test.tsx
```diff
@@ -20,12 +31,62 @@ const MappingTestWidget = (props: { data: Mapping }) => {
...
+            setDataSampleResult('Valid JSON');
+            setDataSampleValid(true);
+            highlightDataSampleErrorLine(null);
+        }
+        catch (err) {
+            setDataSampleResult(err.toString());
+            setDataSampleValid(false);
+            const lineMatches = err.message.match(/line ([0-9]+)/);
+            if (lineMatches) {
+                highlightDataSampleErrorLine(Number(lineMatches[1]) - 1);
```
pallar-ms, line 70: Just curious - are we taking 1st index because the 0th one would be the string "line xx"?

tools/data-mapper/Microsoft.Health.Tools.DataMapper/ClientApp/src/components/mapping/Detail.Widgets.Test.tsx
```diff
@@ -103,9 +164,11 @@ const MappingTestWidget = (props: { data: Mapping }) => {
                                 <span className="h6">Device Data Sample</span>
                             </div>
                             <Input
-                                type="textarea" name="devicedatasample" id="devicedatasample" rows={10}
-                                placeholder="Paste your device data sample here..." innerRef={dataSampleRef}
+                                type="textarea" name="devicedatasample" id="devicedatasample"
```
pallar-ms, line 167: nit: can we camel case this

tools/data-mapper/Microsoft.Health.Tools.DataMapper/ClientApp/src/components/mapping/Detail.Widgets.Test.tsx
kyclai: Yes, exactly.

tools/data-mapper/Microsoft.Health.Tools.DataMapper/ClientApp/src/components/mapping/Detail.Widgets.Test.tsx
kyclai: Good eye, will do. Thanks!

dustinburson (commented):
Changes look good to me.  I will differ approval to those more familiar with React so feel free to commit once you have that. Thanks!

ms-teli (approved):
thanks for addressing the comment.

tools/data-mapper/Microsoft.Health.Tools.DataMapper/ClientApp/src/components/mapping/Detail.Widgets.Test.tsx
```diff
@@ -20,12 +32,65 @@ const MappingTestWidget = (props: { data: Mapping }) => {
...
 
-    const dataSampleRef = React.useRef<HTMLInputElement>() as React.RefObject<HTMLInputElement>;
+    const dataSampleRef = React.useRef<HTMLTextAreaElement>() as React.RefObject<HTMLTextAreaElement>;
     const identityResolutionTypeRef = React.useRef<HTMLInputElement>() as React.RefObject<HTMLInputElement>;
 
+    var codeEditor: CodeMirror.EditorFromTextArea;
+    var dataSampleErrorLine: number | null = null;
+
+    React.useEffect(() => {
+        if (dataSampleRef.current) {
```
ms-teli, line 42: just curious what if it's undefined or aka null, what would happen?

ms-teli (approved):

PR #126, opened 2021-07-30
Handle event processing exceptions based on type.

pallar-ms:
Bug: Upon handling errors when processing events, the corresponding exception was always being cast to EventHubException even when the exception was not of that type causing an invalid cast exception in the error handling and as a result no errors/metrics were being logged.
Fix: Process the exception according to its type.

### DIFF
--- a/src/lib/Microsoft.Health.Events/EventHubProcessor/EventProcessor.cs
+++ b/src/lib/Microsoft.Health.Events/EventHubProcessor/EventProcessor.cs
@@ -4,26 +4,23 @@
 // -------------------------------------------------------------------------------------------------
 
 using System;
-using System.Net.Sockets;
 using System.Threading;
 using System.Threading.Tasks;
 using Azure.Messaging.EventHubs;
 using Azure.Messaging.EventHubs.Consumer;
 using Azure.Messaging.EventHubs.Processor;
 using EnsureThat;
-using Microsoft.Health.Common.Telemetry;
 using Microsoft.Health.Events.EventCheckpointing;
 using Microsoft.Health.Events.EventConsumers.Service;
 using Microsoft.Health.Events.Model;
 using Microsoft.Health.Events.Telemetry;
+using Microsoft.Health.Events.Telemetry.Exceptions;
 using Microsoft.Health.Logging.Telemetry;
 
 namespace Microsoft.Health.Events.EventHubProcessor
 {
     public class EventProcessor
     {
-        private readonly string _eventHubErrorPrefix = "EventHubError";
-
         private IEventConsumerService _eventConsumerService;
         private ICheckpointClient _checkpointClient;
         private ITelemetryLogger _logger;
@@ -64,9 +61,7 @@ namespace Microsoft.Health.Events.EventHubProcessor
             // todo: consider retry
             Task ProcessErrorHandler(ProcessErrorEventArgs eventArgs)
             {
-                (var exception, string reason) = ProcessErrorEventException(eventArgs.Exception);
-                _logger.LogError(exception);
-                _logger.LogMetric(EventMetrics.HandledException(reason, ConnectorOperation.Setup), 1);
+                EventHubExceptionTelemetryProcessor.ProcessException(eventArgs.Exception, _logger);
 
                 return Task.CompletedTask;
             }
@@ -88,7 +83,7 @@ namespace Microsoft.Health.Events.EventHubProcessor
 #pragma warning restore CA1031
                 {
                     _logger.LogTrace($"Failed to initialize partition {partitionId} from checkpoint");
-                    _logger.LogError(ex);
+                    EventHubExceptionTelemetryProcessor.ProcessException(ex, _logger, errorMetricName: EventHubErrorCode.EventHubPartitionInitFailed.ToString());
                 }
             }
 
@@ -111,19 +106,6 @@ namespace Microsoft.Health.Events.EventHubProcessor
                 processor.ProcessEventAsync -= ProcessEventHandler;
                 processor.ProcessErrorAsync -= ProcessErrorHandler;
                 processor.PartitionInitializingAsync -= ProcessInitializingHandler;
-            }
-        }
-
-        private (Exception exception, string reason) ProcessErrorEventException(Exception exception)
-        {
-            switch (exception)
-            {
-                case EventHubsException eventHubsException:
-                    return (eventHubsException, $"{_eventHubErrorPrefix}{eventHubsException.Reason}");
-                case SocketException socketException:
-                    return (socketException, $"{_eventHubErrorPrefix}{socketException.SocketErrorCode}");
-                default:
-                    return (exception, $"{_eventHubErrorPrefix}GeneralError");
             }
         }
     }
--- a/src/lib/Microsoft.Health.Events/EventHubProcessor/EventProcessor.cs
+++ b/src/lib/Microsoft.Health.Events/EventHubProcessor/EventProcessor.cs
@@ -4,6 +4,7 @@
 // -------------------------------------------------------------------------------------------------
 
 using System;
+using System.Net.Sockets;
 using System.Threading;
 using System.Threading.Tasks;
 using Azure.Messaging.EventHubs;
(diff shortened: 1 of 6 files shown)


wi-y (approved):

rogordon01 (approved):

src/lib/Microsoft.Health.Events/EventHubProcessor/EventProcessor.cs
```diff
@@ -111,5 +113,18 @@ async Task ProcessInitializingHandler(PartitionInitializingEventArgs initArgs)
                 processor.PartitionInitializingAsync -= ProcessInitializingHandler;
             }
         }
+
+        private (Exception exception, string reason) ProcessErrorEventException(Exception exception)
```
dustinburson, line 117: Would it be better to have our own exception type for these scenarios?  We can use the original exception as the inner exception but have a strong typed exception class with error codes and other properties.

src/lib/Microsoft.Health.Events/EventHubProcessor/EventProcessor.cs
```diff
@@ -111,5 +113,18 @@ async Task ProcessInitializingHandler(PartitionInitializingEventArgs initArgs)
...
+        {
+            switch (exception)
+            {
+                case EventHubsException eventHubsException:
```
dustinburson, line 121: This will cover authorization and invalid host exceptions?

src/lib/Microsoft.Health.Events/EventHubProcessor/EventProcessor.cs
```diff
@@ -111,5 +113,18 @@ async Task ProcessInitializingHandler(PartitionInitializingEventArgs initArgs)
...
+                    return (eventHubsException, $"{_eventHubErrorPrefix}{eventHubsException.Reason}");
```
dustinburson, line 122: @wi-y Because we are logging this as a metric above will these by exposed to PaaS users as is or is there additional wiring up needed to facilitate that?

wi-y: Yeah it will be exposed as an error metric to PaaS users without additional wiring

dustinburson: Thanks for confirming Will, that is what I thought.  @pallar-ms before we commit I think we need to make sure these are the values we customers & users to see in the product.  Do you have some examples that would be generated?

pallar-ms: Couple examples - 
* EventHubErrorResourceNotFound - frrom the EventHubsException when the event hub provided is not found.
* EventHubErrorHostNotFound - from the SocketException, when the event hub fqdn is invalid.
* EventHubErrorGeneralError - catch all for other exceptions. 
Unfortunately from the documentation, I don't see all the expected exception types that we could specifically handle.

src/lib/Microsoft.Health.Events/EventHubProcessor/EventProcessor.cs
pallar-ms: I feel like bulk of the event hub processing exceptions are captured in EventHubsException with explicit reason/error  [codes](https://docs.microsoft.com/en-us/dotnet/api/azure.messaging.eventhubs.eventhubsexception.failurereason?view=azure-dotnet) and other properties like IsTransient, EventHubName, Message etc. Its just the invalid host which isn't captured in here.
The only thing I can think of to wrap in our own exception class would be if we further want to classify the exceptions from our perspective of IsClient/Service - eg: if invalid event hub was provided then its a client error for us, etc.
If we see value in that, then can add a new exception type.

src/lib/Microsoft.Health.Events/EventHubProcessor/EventProcessor.cs
dustinburson: Okay thanks for the examples Pallavi.  My one concern here is we are relying on the Reason imbedded in the Event Hub exception to form the name we are emitting.  If there is change in the Event Hub library if we could generate different telemetry for the same issue.  I would like to instead have constants we control for these categories for consistency such scenario.  Does that make sense?  With that change other suggestions like potentially refactoring into a dedicated exception type can be deferred.

dustinburson (changes requested):
Rather than string interpolation for error names based off the exception reason.  Would prefer we have constants for these names that we are in control of.

src/lib/Microsoft.Health.Events/EventHubProcessor/EventProcessor.cs
pallar-ms: Yeah I sort of agree, I had the same concern at first of using Reason in case things changed, however from debugging perspective it also made sense to have it reflect the event hub codes as is.
* If the suggestion was to have a mapping b/w the event hub "reasons" and our custom "reasons", we'd miss out on updates that way when new reason codes are created.
* Else we can have a generic reason string but also capture the event hub reason some where although I don't think we have a spare field for that.

dustinburson: We can use the ILogger and log the original internally for us.  The main concern is these names will be exposed to customers that are using Azure Monitor integration.  So if we release this and a customer builds alerts off this information but then later there is a change in the supporting library that could change the name we will break their alerts.

pallar-ms: Got it, makes sense!

src/lib/Microsoft.Health.Events/EventHubProcessor/EventProcessor.cs
pallar-ms: The invalid host will be via SocketException and invalid event hub is through EventHubException.
I'm double checking on the auth one.

src/lib/Microsoft.Health.Events/Telemetry/Exceptions/EventHubExceptionTelemetryProcessor.cs
```diff
@@ -0,0 +1,70 @@
...
+
+        /// <summary>
+        /// Error code that categorizes exceptions of the type SocketException
+        /// </summary>
+        SocketErrors,
+
+        /// <summary>
+        /// Error code that categorizes all other generic Exceptions
+        /// </summary>
+        GeneralErrors,
```
dustinburson, line 35: nit: reason for plural in enum name?

src/lib/Microsoft.Health.Events/Telemetry/Exceptions/EventHubExceptionTelemetryProcessor.cs
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    public static class EventHubExceptionTelemetryProcessor
```
dustinburson, line 38: nit: Style should be in separate files.  Surprised this isn't being flag by code style rules.

dustinburson (approved):
Code looks good, thank you for the changes.  Minor comments on code style.

src/lib/Microsoft.Health.Events/EventHubProcessor/EventProcessor.cs
pallar-ms: Checked that we get UnauthorizedAccessException separately for authorization errors so handled that explicitly. In general though, given we don't know what the underlying types of exceptions we can expect, unfortunately it could be that most of it fall under "GeneralError" category.

src/lib/Microsoft.Health.Events/Telemetry/Exceptions/EventHubExceptionTelemetryProcessor.cs
pallar-ms: No reason, updated it to be singular.

src/lib/Microsoft.Health.Events/Telemetry/Exceptions/EventHubExceptionTelemetryProcessor.cs
pallar-ms: Moved it, although by default enums are allowed - https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1402.md

dustinburson (approved):

PR #127, opened 2021-08-02
Personal/rogordon/fix data mapper build

rogordon01:
Changes in how the CollectionContentTemplateFactory is initialized were not reflected inside of the DataMapper. This caused the project to fail. This PR correctly supplies the template factory as well as fixing additional bugs.

### DIFF
--- a/tools/data-mapper/Microsoft.Health.Tools.DataMapper/TelemetryLoggerFacade.cs
+++ b/tools/data-mapper/Microsoft.Health.Tools.DataMapper/TelemetryLoggerFacade.cs
@@ -12,7 +12,7 @@ using Microsoft.Health.Logging.Telemetry;
 namespace Microsoft.Health.Tools.DataMapper
 {
     /// <summary>
-    /// A facage which routes ITelemetryLogger operations to the underlying ILogger
+    /// A facade which routes ITelemetryLogger operations to the underlying ILogger
     /// </summary>
     public class TelemetryLoggerFacade : ITelemetryLogger
     {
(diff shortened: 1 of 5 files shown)


tools/data-mapper/Microsoft.Health.Tools.DataMapper/TelemetryLoggerFacade.cs
```diff
@@ -0,0 +1,43 @@
...
+using Microsoft.Extensions.Logging;
+using Microsoft.Health.Common.Telemetry;
+using Microsoft.Health.Logging.Telemetry;
+
+namespace Microsoft.Health.Tools.DataMapper
+{
+    /// <summary>
+    /// A facage which routes ITelemetryLogger operations to the underlying ILogger
+    /// </summary>
+    public class TelemetryLoggerFacade : ITelemetryLogger
```
rogordon01, line 17: A future task will remove the use of `ITelemetryLogger` within the Iomt-Fhir project. For now we simply create a facade to route all requests to the underlying `ILogger`

dustinburson: Another option would be to implement ILogger as part of the ITelemetryLogger so implementations could be used interchangeably.  I am fine if you want to keep this but it feels like you have functionality done this already.

dustinburson (approved):

dustinburson:
@kyclai FYI

tools/data-mapper/Microsoft.Health.Tools.DataMapper/TelemetryLoggerFacade.cs
wi-y: iirc there is some weirdness with ILogger and LogMetric

Technically you can call LogMetric with ILogger but because it is not strongly defined you have to intercept the LogMetric call when "Action = Metric" and then log the metric using the application insights sdk.

The Azure Functions/Webjobs team defined an extension for LogMetric which is useful if we are using functions/webjobs: https://github.com/Azure/azure-webjobs-sdk/blob/93834ece3dec6f80fe23a9533bbde3bee49ea22f/src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/LoggerExtensions.cs#L25

I don't think this defined LogMetric extension is available in the non-webjobs sdk though. Things might have changed since we last took a look and it is worth another pass. I am all for replacing the ITelemetryLogger interface if we have a better way of doing things.

PR #128, opened 2021-08-06
Updating the header and file name per OSS CELA recommendation.

msjasteppe:
Updating the file name and header per OSS CELA recommendation.


rbhaiya:
Looks like standard license.

NOTICE.md
```diff
@@ -1,541 +1,539 @@
...
-> 
-> Exhibit B - "Incompatible With Secondary Licenses" Notice
-> 
->       This Source Code Form is "Incompatible
->       With Secondary Licenses", as defined by
+# NOTICES
+
+This repository incorporates material as listed below or described in the code.
+
+## Ensure.That 9.2.0
```
rogordon01, line 5: Do we still need to list all of these individual software licenses?

msjasteppe: Yes. This is just a file name change and an adjustment to the header per this documentation - https://docs.opensource.microsoft.com/using/required-notice-template/

msjasteppe: @rogordon01 @rbhaiya Based on our conversations yesterday, it appears that we no longer need this file. Do we want to shutdown this PR, archive a copy of the file and then remove it completely?

rogordon01: The PR can be abandoned for now. I looked yesterday and all of the software listed here _seems_ to be pulled in as a reference. We can confirm with @dustinburson and @namalu  when they come back. At which point we can issue a new PR to delete the file.
The file will always be in a our Git history, so no need to manual archive it.

msjasteppe:
Closing PR per Rob's recommendation. We will pick this up once Dustin and Nate are back in the office.

PR #131, opened 2021-09-02
Data Mapper: Editing type name

kyclai:
Updated the IoMT Connector Data Mapper as follows:
- Enabled renaming a mapping's type name on the home page (containing the list of mappings) and on the mapping page, as shown in the following screenshots.
  - "Rename" option on home page:
    Home (list) page
  - "Rename" option on mapping page:
    Mapping page
  - Pressing "Rename" from the home page or mapping page opens this dialog, which is pre-filled with the current type name:
    Rename dialog
    - Note: Empty type names are not allowed anymore:
      Rename dialog - empty type name
- Also updated reference to Heart Rate (from Blood Pressure) in tutorial:
  Heart Rate

### DIFF
--- a/tools/data-mapper/Microsoft.Health.Tools.DataMapper/ClientApp/src/components/NavMenu.tsx
+++ b/tools/data-mapper/Microsoft.Health.Tools.DataMapper/ClientApp/src/components/NavMenu.tsx
@@ -8,7 +8,7 @@ import { Navbar, NavbarBrand, NavbarText } from 'reactstrap';
 import { Link, useLocation } from 'react-router-dom';
 import PersistService from '../services/PersistService';
 import { Mapping } from '../store/Mapping';
-import { default as MappingCreationModal, Action } from './mapping/List.Modals.Creation';
+import { default as MappingNameModal, Action } from './mapping/List.Modals.Name';
 import './NavMenu.css';
 
 const NavMenu = (props: {}) => {
@@ -48,7 +48,7 @@ const NavMenu = (props: {}) => {
                         <span className="page-heading" style={{ margin: "0px 10px 0px 50px" }}>
                             {mappingTypeName}
                         </span>
-                        <MappingCreationModal
+                        <MappingNameModal
                             onSave={(typename: string, errorHandler: Function) => renameMapping(mappingId, typename, errorHandler)}
                             action={Action.Rename}
                             inputDefaultValue={mappingTypeName}
--- a/tools/data-mapper/Microsoft.Health.Tools.DataMapper/ClientApp/src/components/NavMenu.tsx
+++ b/tools/data-mapper/Microsoft.Health.Tools.DataMapper/ClientApp/src/components/NavMenu.tsx
@@ -4,17 +4,57 @@
 // -------------------------------------------------------------------------------------------------
 
 import * as React from 'react';
-import { Navbar, NavbarBrand } from 'reactstrap';
-import { Link } from 'react-router-dom';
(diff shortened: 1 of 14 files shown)


tools/data-mapper/Microsoft.Health.Tools.DataMapper/ClientApp/src/components/NavMenu.tsx
```diff
@@ -4,17 +4,60 @@
...
 const NavMenu = (props: {}) => {
+    const [mappingId, setMappingId] = React.useState('');
+    const [mappingTypeName, setMappingTypeName] = React.useState('');
+
+    const location = useLocation();
+
+    React.useEffect(() => {
+        const pathname = location.pathname;
+        const id = pathname.substr(pathname.lastIndexOf('/') + 1);
+        const typeName = PersistService.getMapping(id)?.typeName;
```
pallar-ms, line 23: Just curious - do we expect this (`PersistService.getMapping(id)`) to be empty and if so what is the impact on user experience if we continue with empty typename?

dustinburson: Type name is required

kyclai: The root/home page, which contains a list of mappings, does not have a mapping ID in the URL. Since I'm getting the mapping ID from the URL with Line 22, there may be no mapping ID and therefore no mapping or typename, so I use the `?.` operator on Line 23.

Also, with the changes to _PersistService.ts_ in this PR, it's not possible to add a mapping with an empty typename anymore.

ms-teli:
would that be possible to paste a few screenshots for the UX we're adding? thanks!

kyclai:
> would that be possible to paste a few screenshots for the UX we're adding? thanks!

Thanks, @ms-teli, good suggestion. I've updated the original PR comment with screenshots.

tools/data-mapper/Microsoft.Health.Tools.DataMapper/ClientApp/src/components/NavMenu.tsx
```diff
@@ -4,17 +4,60 @@
...
 
 const NavMenu = (props: {}) => {
+    const [mappingId, setMappingId] = React.useState('');
+    const [mappingTypeName, setMappingTypeName] = React.useState('');
+
+    const location = useLocation();
+
+    React.useEffect(() => {
+        const pathname = location.pathname;
+        const id = pathname.substr(pathname.lastIndexOf('/') + 1);
```
ms-teli, line 22: Nit: please try a few invalid url path and see whether this can handle those situations properly.

ms-teli (approved):
LGTM

tools/data-mapper/Microsoft.Health.Tools.DataMapper/ClientApp/src/components/NavMenu.tsx
kyclai: Thanks, @ms-teli. With your comment I made another commit to get the mapping ID from the URL path more reliably - especially if the path changes in the future.

ms-teli (approved):

tools/data-mapper/Microsoft.Health.Tools.DataMapper/ClientApp/src/components/NavMenu.tsx
ms-teli: thanks for making the change.

PR #137, opened 2021-10-19
Personal/rogordon/batch measurements to eventhub

rogordon01:
This PR updates the normalization pipeline in order to send measurements to the internal eventhub in batches.

### DIFF
--- a/src/func/Microsoft.Health.Fhir.Ingest.Host/IomtConnectorFunctions.cs
+++ b/src/func/Microsoft.Health.Fhir.Ingest.Host/IomtConnectorFunctions.cs
@@ -88,7 +88,7 @@ namespace Microsoft.Health.Fhir.Ingest.Service
                     events.Length);
 
                 IDataNormalizationService<EventData, IMeasurement> dataNormalizationService = new MeasurementEventNormalizationService(_logger, template);
-                await dataNormalizationService.ProcessAsync(events, new EnumerableAsyncCollectorFacade<IMeasurement>(output)).ConfigureAwait(false);
+                await dataNormalizationService.ProcessAsync(events, output).ConfigureAwait(false);
 
                 if (normalizationSettings.Value.LogDeviceIngressSizeBytes)
                 {
--- a/src/func/Microsoft.Health.Fhir.Ingest.Host/IomtConnectorFunctions.cs
+++ b/src/func/Microsoft.Health.Fhir.Ingest.Host/IomtConnectorFunctions.cs
@@ -88,7 +88,7 @@ namespace Microsoft.Health.Fhir.Ingest.Service
                     events.Length);
 
                 IDataNormalizationService<EventData, IMeasurement> dataNormalizationService = new MeasurementEventNormalizationService(_logger, template);
-                await dataNormalizationService.ProcessAsync(events, output).ConfigureAwait(false);
+                await dataNormalizationService.ProcessAsync(events, new EnumerableAsyncCollectorFacade<IMeasurement>(output)).ConfigureAwait(false);
 
                 if (normalizationSettings.Value.LogDeviceIngressSizeBytes)
                 {
--- a/src/lib/Microsoft.Health.Fhir.Ingest/Data/DeterministicHashCodeGenerator.cs
+++ b/src/lib/Microsoft.Health.Fhir.Ingest/Data/DeterministicHashCodeGenerator.cs
@@ -3,6 +3,8 @@
 // Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // -------------------------------------------------------------------------------------------------
 
+using System;
+using System.Globalization;
 using EnsureThat;
 
 namespace Microsoft.Health.Fhir.Ingest.Data
@@ -13,7 +15,7 @@ namespace Microsoft.Health.Fhir.Ingest.Data
         {
         }
 
-        public int GenerateHashCode(string valueToHash)
+        public string GenerateHashCode(string valueToHash)
         {
             EnsureArg.IsNotNull(valueToHash, nameof(valueToHash));
 
@@ -24,7 +26,7 @@ namespace Microsoft.Health.Fhir.Ingest.Data
                 h = unchecked((31 * h) + c);
             }
 
-            return h;
+            return ((byte)(Math.Abs(h) % 256)).ToString(CultureInfo.InvariantCulture);
         }
     }
 }
--- a/src/lib/Microsoft.Health.Fhir.Ingest/Data/DeterministicHashCodeGenerator.cs
+++ b/src/lib/Microsoft.Health.Fhir.Ingest/Data/DeterministicHashCodeGenerator.cs
@@ -15,8 +15,9 @@ namespace Microsoft.Health.Fhir.Ingest.Data
         {
         }
 
-        public string GenerateHashCode(string valueToHash)
+        public string GenerateHashCode(string valueToHash, int range = 256)
         {
+            EnsureArg.IsGt(range, 0, nameof(range));
             EnsureArg.IsNotNull(valueToHash, nameof(valueToHash));
 
             int h = 0;
@@ -26,7 +27,7 @@ namespace Microsoft.Health.Fhir.Ingest.Data
                 h = unchecked((31 * h) + c);
             }
 
-            return ((byte)(Math.Abs(h) % 256)).ToString(CultureInfo.InvariantCulture);
+            return (Math.Abs(h) % range).ToString(CultureInfo.InvariantCulture);
         }
     }
 }
--- a/src/lib/Microsoft.Health.Fhir.Ingest/Data/DeterministicHashCodeGenerator.cs
+++ b/src/lib/Microsoft.Health.Fhir.Ingest/Data/DeterministicHashCodeGenerator.cs
@@ -1,0 +1,30 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using EnsureThat;
+
+namespace Microsoft.Health.Fhir.Ingest.Data
+{
+    public class DeterministicHashCodeGenerator : IHashCodeGenerator
+    {
+        public void Dispose()
+        {
+        }
+
+        public int GenerateHashCode(string valueToHash)
+        {
+            EnsureArg.IsNotNull(valueToHash, nameof(valueToHash));
+
+            int h = 0;
+
+            foreach (char c in valueToHash.ToCharArray())
+            {
+                h = unchecked((31 * h) + c);
+            }
+
+            return h;
+        }
+    }
+}
--- a/src/lib/Microsoft.Health.Fhir.Ingest/Data/MeasurementToEventMessageAsyncCollector.cs
+++ b/src/lib/Microsoft.Health.Fhir.Ingest/Data/MeasurementToEventMessageAsyncCollector.cs
@@ -5,7 +5,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Globalization;
 using System.Linq;
 using System.Text;
 using System.Threading;
@@ -60,9 +59,7 @@ namespace Microsoft.Health.Fhir.Ingest.Data
                 {
                     // cast as byte to restrict to 256 possible values. This will lead to a greater change of measurements ending up in the same bucket,
                     // while providing partition keys with enough entropy for EventHub to better distribute them across partitions.
-                    var partitionKey = (byte)hasher.GenerateHashCode(m.DeviceId.ToLower());
-                    var partitionKeyAsString = partitionKey.ToString(CultureInfo.InvariantCulture);
-                    return partitionKeyAsString;
+                    return hasher.GenerateHashCode(m.DeviceId.ToLower());
                 })
                 .Select(async grp =>
                 {
--- a/src/lib/Microsoft.Health.Fhir.Ingest/Data/MeasurementToEventMessageAsyncCollector.cs
+++ b/src/lib/Microsoft.Health.Fhir.Ingest/Data/MeasurementToEventMessageAsyncCollector.cs
@@ -23,7 +23,6 @@ namespace Microsoft.Health.Fhir.Ingest.Data
         IEnumerableAsyncCollector<IMeasurement>
     {
         private readonly IEventHubMessageService _eventHubService;
-        private readonly int _partitionCount;
         private readonly IHashCodeFactory _hashCodeFactory;
         private readonly ITelemetryLogger _telemetryLogger;
 
@@ -60,7 +59,9 @@ namespace Microsoft.Health.Fhir.Ingest.Data
                 var submissionTasks = items
                 .GroupBy(m =>
                 {
-                    var partitionKey = hasher.GenerateHashCode(m.DeviceId.ToLower()) % _partitionCount;
+                    // cast as byte to restrict to 256 possible values. This will lead to a greater change of measurements ending up in the same bucket,
+                    // while providing partition keys with enough entropy for EventHub to better distribute them across partitions.
+                    var partitionKey = (byte)hasher.GenerateHashCode(m.DeviceId.ToLower());
                     var partitionKeyAsString = partitionKey.ToString(CultureInfo.InvariantCulture);
                     return partitionKeyAsString;
                 })
@@ -68,7 +69,7 @@ namespace Microsoft.Health.Fhir.Ingest.Data
                 {
                     var partitionKey = grp.Key;
                     Stack<EventDataBatch> eventDataBatches = new Stack<EventDataBatch>();
-                    eventDataBatches.Push(await _eventHubService.CreateEventDataBatchAsync());
+                    eventDataBatches.Push(await _eventHubService.CreateEventDataBatchAsync(partitionKey));
 
                     foreach (var m in grp)
                     {
@@ -79,12 +80,13 @@ namespace Microsoft.Health.Fhir.Ingest.Data
                         if (!eventDataBatches.Peek().TryAdd(eventData))
                         {
                             // The current EventDataBatch cannot hold any more events. Create a new EventDataBatch and add this new message to it.
-                            var newEventDataBatch = await _eventHubService.CreateEventDataBatchAsync();
+                            var newEventDataBatch = await _eventHubService.CreateEventDataBatchAsync(partitionKey);
+
                             if (!newEventDataBatch.TryAdd(eventData))
                             {
                                 // The measurement event is greater than the size allowed by EventHub. Log and discard.
                                 // TODO in this case we should send this to a dead letter queue. We'd need to see how we can send it, as it is too big for EventHub...
-                                _telemetryLogger.LogError(new ArgumentOutOfRangeException("A measurement event exceeded the maximum message size for the event hub. It will be skipped."));
+                                _telemetryLogger.LogError(new ArgumentOutOfRangeException($"A measurement event exceeded the maximum message batch size of {newEventDataBatch.MaximumSizeInBytes} bytes. It will be skipped."));
                             }
                             else
                             {
--- a/src/lib/Microsoft.Health.Fhir.Ingest/Data/MeasurementToEventMessageAsyncCollector.cs
+++ b/src/lib/Microsoft.Health.Fhir.Ingest/Data/MeasurementToEventMessageAsyncCollector.cs
@@ -4,25 +4,33 @@
 // -------------------------------------------------------------------------------------------------
 
 using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using Azure.Messaging.EventHubs;
 using EnsureThat;
-using Microsoft.Azure.WebJobs;
 using Microsoft.Health.Fhir.Ingest.Service;
 using Newtonsoft.Json;
 
 namespace Microsoft.Health.Fhir.Ingest.Data
 {
     public class MeasurementToEventMessageAsyncCollector :
-        IAsyncCollector<IMeasurement>
+        IEnumerableAsyncCollector<IMeasurement>
     {
         private readonly IEventHubMessageService _eventHubService;
+        private readonly int _partitionCount;
+        private readonly IHashCodeFactory _hashCodeFactory;
 
-        public MeasurementToEventMessageAsyncCollector(IEventHubMessageService eventHubService)
+        public MeasurementToEventMessageAsyncCollector(
+            IEventHubMessageService eventHubService,
+            IHashCodeFactory hashCodeFactory,
+            int partitionCount = 16)
         {
             _eventHubService = EnsureArg.IsNotNull(eventHubService, nameof(eventHubService));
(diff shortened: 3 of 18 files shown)


src/lib/Microsoft.Health.Fhir.Ingest/Data/MeasurementToEventMessageAsyncCollector.cs
```diff
@@ -39,6 +44,29 @@ public async Task AddAsync(IMeasurement item, CancellationToken cancellationToke
             await _eventHubService.SendAsync(eventList, partitionKey, cancellationToken).ConfigureAwait(false);
         }
 
+        public async Task AddAsync(IEnumerable<IMeasurement> items, CancellationToken cancellationToken = default)
+        {
+            EnsureArg.IsNotNull(items, nameof(items));
+            var submissionTasks = items
+                .GroupBy(m => m.DeviceId == null ? 0 : m.DeviceId.ToLower().GetHashCode() % _partitionCount)
```
dustinburson, line 51: We shouldn't use the internal GetHashCode, it isn't guaranteed to generate the same hash code across app domains/processes.  The issue is if a different process (either due to scale out or app restart) encounters a device id, the hashcode could be different and result for data for the same device landing in different partitions defeating the purpose of partitioning by device id.

I think this is fine for a quick prototype and validation of the idea but we would need a different hashing solution before committing.  

For more information please see https://docs.microsoft.com/en-us/dotnet/api/system.object.gethashcode?view=net-5.0#remarks specifically the warning section.

src/lib/Microsoft.Health.Fhir.Ingest/Microsoft.Health.Fhir.Ingest.csproj
```diff
@@ -6,6 +6,7 @@
     <AssemblyName>Microsoft.Health.Fhir.Ingest</AssemblyName>
     <RootNamespace>Microsoft.Health.Fhir.Ingest</RootNamespace>
     <langversion>7.3</langversion>
+    <Configurations>Release;Debug</Configurations>
```
dustinburson, line 9: Curious why this was added.  Can this be reverted?

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementEventNormalizationService.cs
```diff
@@ -53,30 +54,31 @@ private static ISourceBlock<EventData> StartProducer(IEnumerable<EventData> data
...
         }
 
-        private async Task StartConsumer(ISourceBlock<EventData> producer, IAsyncCollector<IMeasurement> collector, Func<Exception, EventData, Task<bool>> errorConsumer)
+        private async Task StartConsumer(ISourceBlock<EventData> producer, IEnumerableAsyncCollector<IMeasurement> collector, Func<Exception, EventData, Task<bool>> errorConsumer)
         {
             // Collect non operation canceled exceptions as they occur to ensure the entire data stream is processed
             var exceptions = new ConcurrentBag<Exception>();
             var cts = new CancellationTokenSource();
-            var consumer = new ActionBlock<EventData>(
+            var transformingConsumer = new TransformManyBlock<EventData, IMeasurement>(
```
dustinburson, line 78: It isn't immediately clear to me the usage of the TransformManyBlock here.  Can you describe conceptually what you are doing here please?

rogordon01: It's a TPL block that can take an input X and transforms it into an enumerable of Y. Here, I receive a single `EventData` object and produce an enumerable of `IMeasurement` from it. The collection of `IMeasurement` then gets output into the data processing pipeline to later get batched and output to EventHub.

src/lib/Microsoft.Health.Fhir.Ingest/Microsoft.Health.Fhir.Ingest.csproj
rogordon01: Hmmm must have been some Visual Studio weirdness. I'll revert it

build/azure-pipelines-ci.yml
```diff
@@ -13,7 +13,7 @@ jobs:
 
 - job: Linux
   pool:
-    vmImage: 'ubuntu-16.04'
+    vmImage: 'ubuntu-20.04'
```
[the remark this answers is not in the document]
rogordon01: Yup I updated both :)

src/lib/Microsoft.Health.Fhir.Ingest/Data/DeterministicHashCodeGenerator.cs
```diff
@@ -0,0 +1,47 @@
...
+            EnsureArg.IsNotNullOrEmpty(valueToHash, nameof(valueToHash));
+
+            return _hashCodeCache.GetOrCreate(
+                valueToHash,
+                e =>
+                {
+                    e.SetSlidingExpiration(_cachedItemRetention);
+                    e.Size = 1;
+                    var hash = _hasher.ComputeHash(Encoding.Unicode.GetBytes(valueToHash));
+                    return BitConverter.ToInt32(hash, 0);
```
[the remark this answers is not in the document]
dustinburson: Sounds good thanks Rob! I was thinking of this and I wonder if we can do something more targeted for our use case.  We could go very simple and just look at the first character of the value (device id). The main goal of these code is have a fast and simple bucketing of a device to a  partition in a detrerministic way.  I am trying to think of a simple scheme that would allow almost a round robin distribution if device ids are assigned in a sequential manner.  Potentially over optimzing and fine keeping this but still wanted to discuss.

src/func/Microsoft.Health.Fhir.Ingest.Host/IomtConnectorFunctions.cs
```diff
@@ -88,7 +88,7 @@ public IomtConnectorFunctions(ITelemetryLogger logger)
                     events.Length);
 
                 IDataNormalizationService<EventData, IMeasurement> dataNormalizationService = new MeasurementEventNormalizationService(_logger, template);
-                await dataNormalizationService.ProcessAsync(events, output).ConfigureAwait(false);
+                await dataNormalizationService.ProcessAsync(events, new EnumerableAsyncCollectorFacade<IMeasurement>(output)).ConfigureAwait(false);
```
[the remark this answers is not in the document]
dustinburson: Have you verified the default OSS config with Azure Functions and SA behaves as expected with this change?  It seems safe but I would either prefer to verify or keep this code as is and only introduce this change for the non-Azure functions.

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementEventNormalizationService.cs
dustinburson: Okay the one part that isn't clear to me is at what point does the transform block submit its output to the action block that sends the data to EventHub.  Is everything thing from the producer exhausted and buffered in the transform block and everything flushed at the end once the source is exhausted or is it flushed for every input event.

For example if I have 2 events to process, the first one emits 3 measurements  and the second emits 2 measurements.  Does final action block get called twice (once with 3 measurements and another with two) or does it get called once with 5?

If it is the former then this batching really only helps if we are doing projections.  If have I have two distinct messages and each nly output 1 measurement we would still have 2 different requests to the destination event hub (one each).  This is how it appears to work based on my current reading of the code and research.  Does that match your understanding?

If so I am okay with the change, seems relatively safe.  We potentially could make additional modifications to batch even more but it might add additional latency which may not be desirable.

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementEventNormalizationService.cs
```diff
@@ -92,33 +94,53 @@ private async Task StartConsumer(ISourceBlock<EventData> producer, IAsyncCollect
...
+                    {
+                        cts.Cancel();
+                        throw;
+                    }
+                    catch (Exception ex)
+                    {
+                        exceptions.Add(ex);
+                    }
                 },
                 new ExecutionDataflowBlockOptions { MaxDegreeOfParallelism = _maxParallelism, SingleProducerConstrained = true, CancellationToken = cts.Token });
```
dustinburson, line 131: >_maxParallelism [](http://example.com/codeflow?start=77&length=15)

Do you know what max Parallelism you tested with?  I suspect if we do more batching of events parallelism will be counter productive.  If we are only batching measurements  projected from the same event keeping the parallelism should be fine but if my understanding isn't correct and we are batching a different way then we may need to dig into this more.

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementEventNormalizationService.cs
rogordon01: Based on my understanding, in your example it would get called once with 5 measurements.

The BatchBlock is invoked when either the amount of message batched equals `_asyncCollectorBatchSize` or the TransformBlock completes. In your example, the 5 message wouldn't hit the batch size but once the producer finishes the completion will propogate up the pipeline, causing the TransformBlock to finish, which forces the BatchBlock to finish. At this point it will flush out its batched 5 messages to the ActionBlock so that the 5 get written out as batch to EventHub.

rogordon01: FYI, the full TPL pipeline is:
producer -> TransformBlock -> BatchBlock -> ActionBlock. Its in the BatchBlock that message get batched before they are sent to the ActionBlock (i.e. EventHub)

src/lib/Microsoft.Health.Fhir.Ingest/Data/MeasurementToEventMessageAsyncCollector.cs
```diff
@@ -4,25 +4,33 @@
...
 using Azure.Messaging.EventHubs;
 using EnsureThat;
-using Microsoft.Azure.WebJobs;
 using Microsoft.Health.Fhir.Ingest.Service;
 using Newtonsoft.Json;
 
 namespace Microsoft.Health.Fhir.Ingest.Data
 {
     public class MeasurementToEventMessageAsyncCollector :
-        IAsyncCollector<IMeasurement>
```
dustinburson, line 19: >IAsyncCollector<IMeasurement> [](http://example.com/codeflow?start=8&length=29)

Can we keep the implemenation for IAsyncCollector as well?  That will allow you use the old signature and logic for Azure Function implemenation based on my prior comment there with out needing to retest.  Basically add this enchancement for the new version but keep the old impl for backwards compat.

dustinburson: I actually think this breaks the Azure Function.  The function has the binding   [EventHubMeasurementCollector("output", Connection = "OutputEventHub")] IAsyncCollector<IMeasurement> output and this change breaks it since it no longer conforms to this interface.

src/lib/Microsoft.Health.Fhir.Ingest/Data/DeterministicHashCodeGenerator.cs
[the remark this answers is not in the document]
rogordon01: I'm thinking its safer to hash over the entirety of the device id. What you're proposing could work if we could guarantee entropy in the first character of the device id. For example, if the customer used simple Guids as thier device id. But I'm concerned that that could lead to all measurements landing on a single partition if they don't use Guids. For example, if the customer creates device ids in the format `deviceIdXXXXX` where XXXXX is some incrementing number. If we simply take the first character it will always be `d`, which will force all measurements onto the same partition.

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementEventNormalizationService.cs
rogordon01: I think that this is current set to 4. Its my understanding that this will only apply to the TransformBlock. So we can be processing 4 device events in parallel. The outputs of the Measurement should all fan-in to the same BatchBlock/Action block.

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementEventNormalizationService.cs
dustinburson: Gothca I missed the batch block sorry.  So my understanding would be correct with out hte batch block

src/lib/Microsoft.Health.Fhir.Ingest/Data/MeasurementToEventMessageAsyncCollector.cs
rogordon01: `IEnumerableAsyncCollector ` implements `IAsyncColloector. But i'll look into this more to confirm.

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementEventNormalizationService.cs
dustinburson: but theMaxDegreeOfParallelism is being set on the consumer.  Based on the change is there any reason for this to be anything other than 1?  The only reason I can think we would want this to be >1 is if we submit a batch and before it has returned we have already collected enough events to send our next batch.  I guess that is possible.

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementEventNormalizationService.cs
```diff
@@ -92,33 +94,53 @@ private async Task StartConsumer(ISourceBlock<EventData> producer, IAsyncCollect
...
                     }
-                    catch (OperationCanceledException)
-                    {
-                        cts.Cancel();
-                        throw;
-                    }
-#pragma warning disable CA1031
                     catch (Exception ex)
                     {
                         if (await errorConsumer(ex, evt).ConfigureAwait(false))
```
dustinburson, line 102: >errorConsumer [](http://example.com/codeflow?start=34&length=13)

If we are going to retry the exception should we try to preempt execution now?

src/lib/Microsoft.Health.Fhir.Ingest/Data/MeasurementToEventMessageAsyncCollector.cs
dustinburson: Nevermind, it looks like this is a very similar class but different from MeasurementToEventAsyncCollector so the Azure Function will still compile but I still feel it would be better to leave the Azure Function implementation unchanged.

src/lib/Microsoft.Health.Fhir.Ingest/Data/MeasurementToEventMessageAsyncCollector.cs
```diff
@@ -39,6 +47,38 @@ public async Task AddAsync(IMeasurement item, CancellationToken cancellationToke
...
+        public async Task AddAsync(IEnumerable<IMeasurement> items, CancellationToken cancellationToken = default)
+        {
+            EnsureArg.IsNotNull(items, nameof(items));
+
+            using (var hasher = _hashCodeFactory.CreateDeterministicHashCodeGenerator())
+            {
+                var submissionTasks = items
+                .GroupBy(m =>
+                {
+                    var partitionKey = hasher.GenerateHashCode(m.DeviceId.ToLower()) % _partitionCount;
```
dustinburson, line 59: >hasher.GenerateHashCode(m.DeviceId.ToLower()) [](http://example.com/codeflow?start=39&length=45)

It looks like your hash value can be negative, do we need to take the absolute value if the hash?  Not sure how event hub handles a negative partition.

src/lib/Microsoft.Health.Fhir.Ingest/Data/MeasurementToEventMessageAsyncCollector.cs
dustinburson: Ya I just realized they are different, but named very similar and were implementing the same interface hence my confusion.

src/lib/Microsoft.Health.Fhir.Ingest/Data/MeasurementToEventMessageAsyncCollector.cs
rogordon01: The hash value gets converted to a String on the next line. It's the string version that EventHub receives, and they'll determine what partition to place that value one. So we aren't actually telling them what partition to place it on, we are just giving the a key in order to have them make the determination.
Since they should be able to handle any string, the fact that it will contain a negative sign shouldn't matter.

src/lib/Microsoft.Health.Fhir.Ingest/Data/MeasurementToEventMessageAsyncCollector.cs
```diff
@@ -39,6 +47,38 @@ public async Task AddAsync(IMeasurement item, CancellationToken cancellationToke
...
+                .Select(async grp =>
+                {
+                    var partitionKey = grp.Key;
+                    var eventList = new List<EventData>();
+                    foreach (var m in grp)
+                    {
+                        var measurementContent = JsonConvert.SerializeObject(m, Formatting.None);
+                        var contentBytes = Encoding.UTF8.GetBytes(measurementContent);
+                        var eventData = new EventData(contentBytes);
+                        eventList.Add(eventData);
```
dustinburson, line 72: >Add [](http://example.com/codeflow?start=34&length=3)

We should use EventDataBatch instead of a collection of events.  The EventDataBatch checks the total batch size client side and will error if the message added exceeds it.

We will want to modify the logic to catch this exception.  If we hit it we should flush immediately a then proceed with a new batch adding the event that triggered the error in the first place.

The underlying IEventHubMessageService already supports EventDataBatch with the signature SendAsync(EventDataBatch eventData, CancellationToken token)

rogordon01: Good catch. Will do.

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementEventNormalizationService.cs
```diff
@@ -92,33 +94,53 @@ private async Task StartConsumer(ISourceBlock<EventData> producer, IAsyncCollect
...
+                    return createdMeasurements;
+                });
+
+            var asyncCollectorConsumer = new ActionBlock<IMeasurement[]>(
+                async measurements =>
+                {
+                    try
+                    {
+                        await collector.AddAsync(measurements, cts.Token).ConfigureAwait(false);
+                        _log.LogMetric(
```
dustinburson, line 117: >LogMetric [](http://example.com/codeflow?start=29&length=9)

This is optional but I was thinking if we could have the destination partition in scope here it would be helpful in diagnosising if we have a hot partition.  This might be better for another PR/work item if we want to do this because I feel the refactor would be extensive but at least wanted to discuss.

rogordon01: I actually really like that idea, and logging it as a metric. At least we'd able to keep track of how we are distributing partition keys in our logic. We'd still need to leverage the work that Pallavi's has done to know how the EventHub  is actually distributing them across the EventHub but have the information here would help too.

pallar-ms: I actually did include the partitionId for this event already. We may just need to pull latest - https://github.com/microsoft/iomt-fhir/blob/master/src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementEventNormalizationService.cs#:~:text=IomtMetrics.NormalizedEvent(partitionId)%2C

dustinburson: Thanks Pallavi.  I think with your changes this will be the source partition id and not the destination partition but after discussing with Rob Event Hub still hashes the partition value we assign to get the true underlying partition so we don't know the actual destintion partiton at this point.

While knowing the destination partition the normalized event is going to would be helpful I don't know if we could realistically do that at this point.  We could track this on the FHIR conversion side but we would only know we have a hot partition if we have been able to convert enough data to FHIR.  Still useful for analytics but it might not help diagnosis a hot partition in real time.  Open to further thoughts.

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementEventNormalizationService.cs
rogordon01: I ended up setting this to 1 from 3.

src/func/Microsoft.Health.Fhir.Ingest.Host/IomtConnectorFunctions.cs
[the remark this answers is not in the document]
rogordon01: I reverted this change and moved the use of the facade into NormalizationService.ProcessAsync.

src/lib/Microsoft.Health.Fhir.Ingest/Data/MeasurementToEventMessageAsyncCollector.cs
```diff
@@ -39,6 +50,61 @@ public async Task AddAsync(IMeasurement item, CancellationToken cancellationToke
...
+            EnsureArg.IsNotNull(items, nameof(items));
+
+            using (var hasher = _hashCodeFactory.CreateDeterministicHashCodeGenerator())
+            {
+                var submissionTasks = items
+                .GroupBy(m =>
+                {
+                    // cast as byte to restrict to 256 possible values. This will lead to a greater change of measurements ending up in the same bucket,
+                    // while providing partition keys with enough entropy for EventHub to better distribute them across partitions.
+                    var partitionKey = (byte)hasher.GenerateHashCode(m.DeviceId.ToLower());
```
rogordon01, line 64: Here is where we restrict the hashcode to the values within the range of a `byte`

dustinburson: I don't think we should use a cast here.  I just tested, if the byte value is < 0 it will become zero and if it is > 255 it will become 255 with this method.  It means we will be artifically sending most hashes to partition key 0 or 255.

I would suggest one of 3 things:
* Update Hash interface to a byte
* Update the Hash interface to a generic and this class uses  the of byte one
* Or update the hash interface to return string to match what event hub wants for a partition key.  This would have the ultimate flexibility, allowing us to freely switch between the old int based version, the new byte based version, or another yet to be determined method.

I prefer the last one personally.

dustinburson: After some more digging it looks like there is no + operand defined for bytes which makes things tricker.  Adding two bytes together results in an integer.

To avoid use biasing towards the low or high end (0 or 255) I think we would need to do %256 and take the absolute value based on your current hasing algo

dustinburson: Something like this should do the trick if you go with the 3rd option, replace return h with return  ((byte)(Math.Abs(h) % 256)).ToString();

dustinburson (approved):
:shipit:

PR #142, opened 2021-11-01
Customized Event Hub configuration exceptions

kyclai:
Added custom exceptions for the following scenarios:
- Event Hub not configured (e.g. Namespace/FQDN, name, or consumer group is incorrect)
- Event Hub access not authorized (e.g. Azure Event Hubs Data Receiver role is missing)

### DIFF
--- a/src/lib/Microsoft.Health.Common/Telemetry/Metrics/Dimensions/DimensionNames.cs
+++ b/src/lib/Microsoft.Health.Common/Telemetry/Metrics/Dimensions/DimensionNames.cs
@@ -41,5 +41,10 @@ namespace Microsoft.Health.Common.Telemetry
         /// A metric dimension that represents the reason that caused the metric to be emitted.
         /// </summary>
         public static string Reason => nameof(Reason);
+
+        /// <summary>
+        /// A metric dimension to identify the error source, e.g. the user or system.
+        /// </summary>
+        public static string ErrorSource => nameof(ErrorSource);
     }
 }
--- a/src/lib/Microsoft.Health.Common/Telemetry/Metrics/Dimensions/DimensionNames.cs
+++ b/src/lib/Microsoft.Health.Common/Telemetry/Metrics/Dimensions/DimensionNames.cs
@@ -38,13 +38,13 @@ namespace Microsoft.Health.Common.Telemetry
         public static string ErrorSeverity => nameof(ErrorSeverity);
 
         /// <summary>
+        /// A metric dimension to identify the error source, e.g. the user or service.
+        /// </summary>
+        public static string ErrorSource => nameof(ErrorSource);
+
+        /// <summary>
         /// A metric dimension that represents the reason that caused the metric to be emitted.
         /// </summary>
         public static string Reason => nameof(Reason);
-
-        /// <summary>
-        /// A metric dimension to identify the error source, e.g. the user or system.
-        /// </summary>
-        public static string ErrorSource => nameof(ErrorSource);
     }
 }
--- a/src/lib/Microsoft.Health.Common/Telemetry/Metrics/Dimensions/ErrorSource.cs
+++ b/src/lib/Microsoft.Health.Common/Telemetry/Metrics/Dimensions/ErrorSource.cs
@@ -16,5 +16,10 @@ namespace Microsoft.Health.Common.Telemetry
         /// A user error.
         /// </summary>
         User,
+
+        /// <summary>
+        /// An error with an undefined source.
+        /// </summary>
+        Undefined,
     }
 }
--- a/src/lib/Microsoft.Health.Common/Telemetry/Metrics/Dimensions/ErrorSource.cs
+++ b/src/lib/Microsoft.Health.Common/Telemetry/Metrics/Dimensions/ErrorSource.cs
@@ -8,6 +8,11 @@ namespace Microsoft.Health.Common.Telemetry
     public enum ErrorSource
     {
         /// <summary>
+        /// An error with an undefined source.
+        /// </summary>
+        Undefined,
+
+        /// <summary>
         /// A service error.
         /// </summary>
         Service,
@@ -16,10 +21,5 @@ namespace Microsoft.Health.Common.Telemetry
         /// A user error.
         /// </summary>
         User,
-
-        /// <summary>
-        /// An error with an undefined source.
-        /// </summary>
-        Undefined,
     }
 }
--- a/src/lib/Microsoft.Health.Common/Telemetry/Metrics/Dimensions/ErrorSource.cs
+++ b/src/lib/Microsoft.Health.Common/Telemetry/Metrics/Dimensions/ErrorSource.cs
@@ -1,0 +1,20 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Common.Telemetry
+{
+    public static class ErrorSource
+    {
+        /// <summary>
+        /// A system error.
+        /// </summary>
+        public static string System => nameof(System);
+
+        /// <summary>
+        /// A user error.
+        /// </summary>
+        public static string User => nameof(User);
+    }
+}
--- a/src/lib/Microsoft.Health.Common/Telemetry/Metrics/Dimensions/ErrorSource.cs
+++ b/src/lib/Microsoft.Health.Common/Telemetry/Metrics/Dimensions/ErrorSource.cs
@@ -5,16 +5,16 @@
 
 namespace Microsoft.Health.Common.Telemetry
 {
-    public static class ErrorSource
+    public enum ErrorSource
     {
         /// <summary>
-        /// A system error.
+        /// A service error.
         /// </summary>
-        public static string System => nameof(System);
+        Service,
 
         /// <summary>
         /// A user error.
         /// </summary>
-        public static string User => nameof(User);
+        User,
     }
 }
--- a/src/lib/Microsoft.Health.Events/Telemetry/Exceptions/EventHubConfigurationExceptionTelemetryProcessor.cs
+++ b/src/lib/Microsoft.Health.Events/Telemetry/Exceptions/EventHubConfigurationExceptionTelemetryProcessor.cs
@@ -13,7 +13,7 @@ namespace Microsoft.Health.Events.Telemetry.Exceptions
 {
     public class EventHubConfigurationExceptionTelemetryProcessor
     {
-        private readonly HashSet<Type> _handledExceptions;
+        private readonly ISet<Type> _handledExceptions;
 
         public EventHubConfigurationExceptionTelemetryProcessor()
             : this(
--- a/src/lib/Microsoft.Health.Events/Telemetry/Exceptions/EventHubConfigurationExceptionTelemetryProcessor.cs
+++ b/src/lib/Microsoft.Health.Events/Telemetry/Exceptions/EventHubConfigurationExceptionTelemetryProcessor.cs
@@ -31,6 +31,7 @@ namespace Microsoft.Health.Events.Telemetry.Exceptions
         {
             EnsureArg.IsNotNull(ex, nameof(ex));
             EnsureArg.IsNotNull(logger, nameof(logger));
+            EnsureArg.IsNotNullOrWhiteSpace(connectorStage, nameof(connectorStage));
 
             var exType = ex.GetType();
 
--- a/src/lib/Microsoft.Health.Events/Telemetry/Exceptions/EventHubConfigurationExceptionTelemetryProcessor.cs
+++ b/src/lib/Microsoft.Health.Events/Telemetry/Exceptions/EventHubConfigurationExceptionTelemetryProcessor.cs
@@ -1,0 +1,63 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using EnsureThat;
+using Microsoft.Health.Common.Telemetry;
+using Microsoft.Health.Logging.Telemetry;
+
+namespace Microsoft.Health.Events.Telemetry.Exceptions
+{
+    public class EventHubConfigurationExceptionTelemetryProcessor
+    {
+        private readonly HashSet<Type> _handledExceptions;
+
+        public EventHubConfigurationExceptionTelemetryProcessor()
+            : this(
+                typeof(InvalidEventHubException),
+                typeof(UnauthorizedAccessEventHubException))
+        {
+        }
+
+        public EventHubConfigurationExceptionTelemetryProcessor(params Type[] handledExceptionTypes)
+        {
+            _handledExceptions = new HashSet<Type>(handledExceptionTypes);
+        }
+
+        public bool HandleException(Exception ex, ITelemetryLogger logger, string connectorStage)
+        {
+            EnsureArg.IsNotNull(ex, nameof(ex));
+            EnsureArg.IsNotNull(logger, nameof(logger));
+
+            var exType = ex.GetType();
+
+            var lookupType = exType.IsGenericType ? exType.GetGenericTypeDefinition() : exType;
+
+            if (_handledExceptions.Contains(lookupType))
+            {
+                if (ex is ITelemetryFormattable tel)
+                {
+                    logger.LogMetric(
+                        metric: tel.ToMetric,
+                        metricValue: 1);
+                }
+                else
+                {
+                    var metric = EventMetrics.HandledException(
+                        exType.Name,
+                        connectorStage);
+                    logger.LogMetric(
+                        metric: metric,
+                        metricValue: 1);
+                }
+
+                return true;
+            }
+
+            return false;
+        }
+    }
+}
--- a/src/lib/Microsoft.Health.Events/Telemetry/Exceptions/EventHubErrorCode.cs
+++ b/src/lib/Microsoft.Health.Events/Telemetry/Exceptions/EventHubErrorCode.cs
@@ -8,19 +8,19 @@ namespace Microsoft.Health.Events.Telemetry
     public enum EventHubErrorCode
     {
         /// <summary>
-        /// Error code that categorizes exceptions of the type EventHubsException
+        /// Error code for an issue relating to an Event Hub (e.g. invalid name or consumer group)
         /// </summary>
-        OperationError,
+        InstanceError,
+
+        /// <summary>
+        /// Error code for an issue relating to an Event Hubs Namespace (e.g. invalid FQDN)
+        /// </summary>
+        NamespaceError,
 
         /// <summary>
         /// Error code that indicates failures in initializing event hub partition
         /// </summary>
         EventHubPartitionInitFailed,
-
-        /// <summary>
-        /// Error code that categorizes exceptions of the type SocketException
-        /// </summary>
-        SocketError,
 
         /// <summary>
         /// Error code that categorizes authentication errors (eg: exceptions of the type UnauthorizedAccessException)
--- a/src/lib/Microsoft.Health.Events/Telemetry/Exceptions/EventHubErrorCode.cs
+++ b/src/lib/Microsoft.Health.Events/Telemetry/Exceptions/EventHubErrorCode.cs
@@ -8,19 +8,14 @@ namespace Microsoft.Health.Events.Telemetry
     public enum EventHubErrorCode
     {
         /// <summary>
-        /// Error code for an issue relating to an Event Hub (e.g. invalid name or consumer group)
-        /// </summary>
-        InstanceError,
-
-        /// <summary>
-        /// Error code for an issue relating to an Event Hubs Namespace (e.g. invalid FQDN)
-        /// </summary>
-        NamespaceError,
-
-        /// <summary>
         /// Error code that indicates failures in initializing event hub partition
         /// </summary>
         EventHubPartitionInitFailed,
+
+        /// <summary>
+        /// Error code that categorizes invalid configurations (e.g. invalid namespace/FQDN, event hub name, or consumer group)
+        /// </summary>
+        ConfigurationError,
 
         /// <summary>
         /// Error code that categorizes authentication errors (eg: exceptions of the type UnauthorizedAccessException)
--- a/src/lib/Microsoft.Health.Events/Telemetry/Exceptions/EventHubExceptionTelemetryProcessor.cs
+++ b/src/lib/Microsoft.Health.Events/Telemetry/Exceptions/EventHubExceptionTelemetryProcessor.cs
@@ -45,7 +45,6 @@ namespace Microsoft.Health.Events.Telemetry.Exceptions
         public static Exception CustomizeException(Exception exception)
         {
             string message = exception.Message;
-            string helpLink = exception.HelpLink;
 
             switch (exception)
             {
@@ -66,11 +65,11 @@ namespace Microsoft.Health.Events.Telemetry.Exceptions
                             return exception;
                     }
 
-                    return new InvalidEventHubException(message, exception, helpLink, EventHubErrorCode.ConfigurationError.ToString());
+                    return new InvalidEventHubException(message, exception, EventHubErrorCode.ConfigurationError.ToString());
 
                 case InvalidOperationException _:
                     message = "Verify that the provided Event Hub contains the provided consumer group.";
-                    return new InvalidEventHubException(message, exception, helpLink, EventHubErrorCode.ConfigurationError.ToString());
+                    return new InvalidEventHubException(message, exception, EventHubErrorCode.ConfigurationError.ToString());
 
                 case SocketException _:
                     SocketException socketException = (SocketException)exception;
@@ -78,14 +77,14 @@ namespace Microsoft.Health.Events.Telemetry.Exceptions
                     {
                         case SocketError.HostNotFound:
                             message = "Verify that the provided Event Hubs Namespace exists.";
-                            return new InvalidEventHubException(message, exception, helpLink, EventHubErrorCode.ConfigurationError.ToString());
+                            return new InvalidEventHubException(message, exception, EventHubErrorCode.ConfigurationError.ToString());
                         default:
                             return exception;
                     }
 
                 case UnauthorizedAccessException _:
                     message = "Verify that the provided Event Hub's 'Azure Event Hubs Data Receiver' role has been assigned to the applicable Azure Active Directory security principal or managed identity.";
-                    helpLink = "https://docs.microsoft.com/azure/event-hubs/authenticate-application";
+                    string helpLink = "https://docs.microsoft.com/azure/event-hubs/authenticate-application";
                     return new UnauthorizedAccessEventHubException(message, exception, helpLink, EventHubErrorCode.AuthorizationError.ToString());
 
                 default:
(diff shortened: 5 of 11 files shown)


src/lib/Microsoft.Health.Common/Telemetry/Metrics/Dimensions/DimensionNames.cs
```diff
@@ -41,5 +41,10 @@ public static class DimensionNames
         /// A metric dimension that represents the reason that caused the metric to be emitted.
         /// </summary>
         public static string Reason => nameof(Reason);
+
+        /// <summary>
+        /// A metric dimension to identify the error source, e.g. the user or system.
+        /// </summary>
+        public static string ErrorSource => nameof(ErrorSource);
```
kyclai, line 48: Open to making this be a flag/boolean if we don't expect other types of errors!

src/lib/Microsoft.Health.Events/Telemetry/Exceptions/EventHubConfigurationExceptionTelemetryProcessor.cs
```diff
@@ -0,0 +1,63 @@
...
+
+using System;
+using System.Collections.Generic;
+using EnsureThat;
+using Microsoft.Health.Common.Telemetry;
+using Microsoft.Health.Logging.Telemetry;
+
+namespace Microsoft.Health.Events.Telemetry.Exceptions
+{
+    public class EventHubConfigurationExceptionTelemetryProcessor
```
kyclai, line 14: Note: @pallar-ms has an upcoming change that will better process handled vs. unhandled exceptions and allow us to derive this class (`EventHubConfigurationExceptionTelemetryProcessor`) from a base class.

src/lib/Microsoft.Health.Events/Telemetry/Exceptions/InvalidEventHubException.cs
```diff
@@ -0,0 +1,54 @@
...
+        {
+        }
+
+        public InvalidEventHubException(string message, Exception innerException, string helpLink, string errorName)
+            : base(message, innerException)
+        {
+            EnsureArg.IsNotNullOrWhiteSpace(errorName, nameof(errorName));
+
+            HelpLink = helpLink;
+            _metricName = $"{ErrorType.EventHubError}{errorName}";
```
rogordon01, line 39: If any of the other constructors are used the `HelpLink` and `_metricName` look to be `null`. How will that impact the creation of the `Metric` in the `ToMetric` method? Will a `NullPointerException` be thrown?

src/lib/Microsoft.Health.Events/Telemetry/Exceptions/InvalidEventHubException.cs
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+        public Metric ToMetric => new Metric(
```
rogordon01, line 42: This code is identical between the exceptions. I'd move it into a base class that they both extend

src/lib/Microsoft.Health.Common/Telemetry/Metrics/Dimensions/ErrorSource.cs
```diff
@@ -0,0 +1,20 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Common.Telemetry
+{
+    public static class ErrorSource
```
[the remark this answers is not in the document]
pallar-ms: Yeah I'd prefer enum too for type safety.

src/lib/Microsoft.Health.Common/Telemetry/Metrics/Dimensions/ErrorSource.cs
```diff
@@ -0,0 +1,20 @@
...
+    {
+        /// <summary>
+        /// A system error.
+        /// </summary>
+        public static string System => nameof(System);
```
pallar-ms, line 13: nit: Service instead?

src/lib/Microsoft.Health.Events/Telemetry/Exceptions/EventHubConfigurationExceptionTelemetryProcessor.cs
```diff
@@ -0,0 +1,63 @@
...
+    {
+        private readonly HashSet<Type> _handledExceptions;
+
+        public EventHubConfigurationExceptionTelemetryProcessor()
+            : this(
+                typeof(InvalidEventHubException),
```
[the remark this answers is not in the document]
pallar-ms: I think having such an interface makes sense. 

The side note to this that I want to point out about "handled" exceptions is that there could be two interpretations with these being introduced now - 
1. Set of exceptions that we will expect during the processing of an event batch, that we will just catch, log and then continue the processing of next events and not stall . Refer usage of https://github.com/microsoft/iomt-fhir/blob/master/src/lib/Microsoft.Health.Fhir.Ingest/Telemetry/ExceptionTelemetryProcessor.cs 
In the above cases "handled" means we will not throw on these or stop processing.
2. Set of exceptions that we will expect during the processing of an event batch, that we will catch, convert to custom exceptions and log these custom exceptions, but unlike # 1 we will throw as we cannot just ignore these as this will keep occuring until the root cause is fixed, so we cannot move on to the next events.

Just calling out the above in case it influences how we want to model the IHandleable interface.

src/lib/Microsoft.Health.Events/Telemetry/Exceptions/EventHubErrorCode.cs
```diff
@@ -8,19 +8,19 @@ namespace Microsoft.Health.Events.Telemetry
...
         /// </summary>
-        OperationError,
+        InstanceError,
 
         /// <summary>
-        /// Error code that indicates failures in initializing event hub partition
+        /// Error code for an issue relating to an Event Hubs Namespace (e.g. invalid FQDN)
         /// </summary>
-        EventHubPartitionInitFailed,
+        NamespaceError,
```
pallar-ms, line 18: I would prefer combining these into a single "InvalidConfigError" or something like that to generally indicate that name/consumergroup/namespace is incorrect. Otherwise we will be exposing a shoebox metric for each configuration. 
But happy to hear your thoughts on this.

src/lib/Microsoft.Health.Events/Telemetry/Exceptions/EventHubExceptionTelemetryProcessor.cs
```diff
@@ -23,26 +23,69 @@ public static class EventHubExceptionTelemetryProcessor
...
+                {
+                    logger.LogMetric(
+                        EventMetrics.HandledException(
+                            errorMetricName ?? $"{ErrorType.EventHubError}{EventHubErrorCode.GeneralError}",
+                            ConnectorOperation.Setup),
+                        1);
+                }
+                else
+                {
+                    var processor = new EventHubConfigurationExceptionTelemetryProcessor();
```
pallar-ms, line 42: Can consider saving this instead of creating a new instance each time.

src/lib/Microsoft.Health.Events/Telemetry/Exceptions/EventHubExceptionTelemetryProcessor.cs
```diff
@@ -23,26 +23,69 @@ public static class EventHubExceptionTelemetryProcessor
             EnsureArg.IsNotNull(exception, nameof(exception));
             EnsureArg.IsNotNull(logger, nameof(logger));
 
-            logger.LogError(exception);
+            var ex = CustomizeException(exception);
+
+            logger.LogError(ex);
 
             if (shouldLogMetric)
```
pallar-ms, line 30: I think we can just get rid of `shouldLogMetric` as an input since we are processing all event hub exceptions within this class and can decide in here if certain exceptions should not be logged (although for now I don't see why we shouldn't log a metric for these cases)

src/lib/Microsoft.Health.Events/Telemetry/Exceptions/EventHubExceptionTelemetryProcessor.cs
```diff
@@ -23,26 +23,69 @@ public static class EventHubExceptionTelemetryProcessor
...
+                            return exception;
+                    }
+
+                    return new InvalidEventHubException(message, exception, helpLink, EventHubErrorCode.InstanceError.ToString());
+
+                case InvalidOperationException _:
+                    message = "Verify that the provided Event Hub contains the provided consumer group.";
+                    return new InvalidEventHubException(message, exception, helpLink, EventHubErrorCode.InstanceError.ToString());
+
                 case SocketException _:
```
pallar-ms, line 78: Is there a SocketErrorCode we can also check against for this case. 
This might not be the only scenario when we will get a socket exception.

src/lib/Microsoft.Health.Events/Telemetry/Exceptions/InvalidEventHubException.cs
```diff
@@ -0,0 +1,54 @@
...
+
+using System;
+using System.Collections.Generic;
+using EnsureThat;
+using Microsoft.Health.Common.Telemetry;
+
+namespace Microsoft.Health.Events.Telemetry.Exceptions
+{
+    public class InvalidEventHubException :
+        Exception,
```
pallar-ms, line 14: I remember us discussing a base class to capture some common fields (IsClientError, IsRetryable etc.) as a general base exception type (i.e., not specific to EventHub exceptions). Are you planning to send that as a subsequent change, totally fine if so, was just curious where we landed with that.

pallar-ms (changes requested):

src/lib/Microsoft.Health.Common/Telemetry/Metrics/Dimensions/ErrorSource.cs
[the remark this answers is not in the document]
kyclai: No good reason. I just thought of keeping it consistent with the other metric dimension types.
Will update.

src/lib/Microsoft.Health.Events/Telemetry/Exceptions/EventHubConfigurationExceptionTelemetryProcessor.cs
```diff
@@ -0,0 +1,63 @@
...
+using System.Collections.Generic;
+using EnsureThat;
+using Microsoft.Health.Common.Telemetry;
+using Microsoft.Health.Logging.Telemetry;
+
+namespace Microsoft.Health.Events.Telemetry.Exceptions
+{
+    public class EventHubConfigurationExceptionTelemetryProcessor
+    {
+        private readonly HashSet<Type> _handledExceptions;
```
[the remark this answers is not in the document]
kyclai: TIL that `HashSet` implements other interfaces, thanks!
Will do.

src/lib/Microsoft.Health.Events/Telemetry/Exceptions/EventHubConfigurationExceptionTelemetryProcessor.cs
```diff
@@ -0,0 +1,63 @@
...
+        public EventHubConfigurationExceptionTelemetryProcessor(params Type[] handledExceptionTypes)
+        {
+            _handledExceptions = new HashSet<Type>(handledExceptionTypes);
+        }
+
+        public bool HandleException(Exception ex, ITelemetryLogger logger, string connectorStage)
+        {
+            EnsureArg.IsNotNull(ex, nameof(ex));
+            EnsureArg.IsNotNull(logger, nameof(logger));
+
```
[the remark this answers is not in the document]
kyclai: I didn't think it'd be an issue if Operation ends up empty in the logs.

@pallar-ms - thoughts? I assumed it'd be unlikely for Operation to be filtered on.

src/lib/Microsoft.Health.Events/Telemetry/Exceptions/EventHubExceptionTelemetryProcessor.cs
```diff
@@ -23,26 +23,69 @@ public static class EventHubExceptionTelemetryProcessor
...
+
             switch (exception)
             {
                 case EventHubsException _:
-                    return $"{ErrorType.EventHubError}{EventHubErrorCode.OperationError}";
+                    EventHubsException eventHubsException = (EventHubsException)exception;
+                    switch (eventHubsException.Reason)
+                    {
+                        case EventHubsException.FailureReason.ConsumerDisconnected:
+                            message = "Verify that the provided Event Hub is not already receiving data for another IoT connector or Azure resource.";
```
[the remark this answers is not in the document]
kyclai: **Background information:**

The `ConsumerDisconnected` reason means that "A consumer client was disconnected by the Event Hub service from the Event Hub instance. It typically occurs when a consumer with a higher owner level asserts ownership over a partition and consumer group pairing."

I encountered the `EventHubsException` exception and `ConsumerDisconnected` reason only when testing locally with a ConnectionString. To trigger this exception and reason, I pointed to another Event Hub with the following changes to the InputEventHub:ConnectionString:
- Set `Endpoint` to a new Namespace's FQDN in a new Resource Group (which was deployed via Sandbox)
- Ensured this new Namespace I'm pointing at contains an Event Hub with the same SAS policy as the previous Event Hub's claims (it was the Manage claim which includes the Listen claim), and then updated `SharedAccessKeyName` and `SharedAccessKey` to this new SAS policy

The exception message was `Receiver '<an id>' with a higher epoch '1' already exists. Receiver 'nil' with epoch 0 cannot be created. Make sure you are creating receiver with increasing epoch value to ensure connectivity, or ensure all old epoch receivers are closed or disconnected.`

**Thoughts:**

Given this exception message and that the new Event Hub I tested with was deployed via Sandbox, I thought that the exception was triggered if the Event Hub is already receiving data for another resource -- which, now that you point it out and I re-think it, doesn't make sense since there can be data from multiple devices (my test here was for Normalization).
Would this custom exception message be better?: _"Verify that the provided Event Hub's **consumer group** is not already receiving data for another IoT connector or Azure resource."_
Please correct me if I'm wrong, but it's technically the "partition" that shouldn't already be receiving data, but the user can't check/update the partition? So we can generalize with "consumer group" instead?

src/lib/Microsoft.Health.Events/Telemetry/Exceptions/EventHubConfigurationExceptionTelemetryProcessor.cs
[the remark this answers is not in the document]
kyclai: That makes sense and we should consider it when making the interface.
For now I'll defer this change to when we implement @pallar-ms's later change that will better process handled vs. unhandled exceptions.

src/lib/Microsoft.Health.Events/Telemetry/Exceptions/EventHubExceptionTelemetryProcessor.cs
kyclai: Good point, thanks.

src/lib/Microsoft.Health.Events/Telemetry/Exceptions/EventHubExceptionTelemetryProcessor.cs
kyclai: Sounds good. I was wondering the same but forgot to ask.

src/lib/Microsoft.Health.Events/Telemetry/Exceptions/InvalidEventHubException.cs
kyclai: You're right, a `NullReferenceException` will be thrown in this case. Thanks for catching this, will handle.

src/lib/Microsoft.Health.Common/Telemetry/Metrics/Dimensions/ErrorSource.cs
kyclai: Ok, will do.

src/lib/Microsoft.Health.Events/Telemetry/Exceptions/EventHubExceptionTelemetryProcessor.cs
kyclai: Thanks for noticing this. Will only handle for `SocketError.HostNotFound`.

src/lib/Microsoft.Health.Events/Telemetry/Exceptions/EventHubExceptionTelemetryProcessor.cs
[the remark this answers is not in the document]
rogordon01: Just to clarify, when you encountered this error during testing, did you have multiple consumers attached to the EventHub? The exception _sounds_ as though it's saying a new consumer has been added to the consumer group. If that is the case I'm wondering if this exception is expected; the EventHub client has logic to balance the load among multiple consumers.

src/lib/Microsoft.Health.Events/Telemetry/Exceptions/EventHubErrorCode.cs
kyclai: As discussed offline, only 1 metric (TotalErrors) is exposed in shoebox and the user can filter on the "Name" dimension to see that there's an event hub configuration error. A specific error for each event hub configuration error is unneeded.
Therefore, I've combined `NamespaceError` and `InstanceError` into a single `ConfigurationError` (naming matches that of the other error codes; feel free to suggest another one).

src/lib/Microsoft.Health.Events/Telemetry/Exceptions/InvalidEventHubException.cs
kyclai: I was deferring it but I've added a base exception `IomtException` now.

src/lib/Microsoft.Health.Events/Telemetry/Exceptions/EventHubConfigurationExceptionTelemetryProcessor.cs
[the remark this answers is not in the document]
rogordon01: Agreed. Perhaps two types of interfaces would be needed here: `IHandledAndContinue` and `IHandledAndThrow`. Or a flag which indicates if you can continue or need to throw. Something to that effect.

src/lib/Microsoft.Health.Common/Telemetry/Metrics/Dimensions/ErrorSource.cs
```diff
@@ -0,0 +1,25 @@
...
+
+        /// <summary>
+        /// A user error.
+        /// </summary>
+        User,
+
+        /// <summary>
+        /// An error with an undefined source.
+        /// </summary>
+        Undefined,
```
pallar-ms, line 23: nit: I would place this at the top since that would be taken as the default if not set.

src/lib/Microsoft.Health.Events/Telemetry/Exceptions/EventHubConfigurationExceptionTelemetryProcessor.cs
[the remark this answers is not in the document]
pallar-ms: We should not ideally allow null/empty Operation as we will want to filter on these when debugging, etc.

src/lib/Microsoft.Health.Events/Telemetry/Exceptions/EventHubExceptionTelemetryProcessor.cs
```diff
@@ -14,35 +14,81 @@ namespace Microsoft.Health.Events.Telemetry.Exceptions
...
+
             switch (exception)
             {
                 case EventHubsException _:
-                    return $"{ErrorType.EventHubError}{EventHubErrorCode.OperationError}";
+                    EventHubsException eventHubsException = (EventHubsException)exception;
+                    switch (eventHubsException.Reason)
+                    {
+                        case EventHubsException.FailureReason.ConsumerDisconnected:
+                            message = "Verify that the provided Event Hub's consumer group is not already receiving data for another IoT connector or Azure resource.";
```
pallar-ms, line 56: Maybe something we can consider as a backlog item, but ideally these strings should be localized.

src/lib/Microsoft.Health.Events/Telemetry/Exceptions/InvalidEventHubException.cs
pallar-ms: Thanks for making the change, this is much cleaner IMO!

src/lib/Microsoft.Health.Events/Telemetry/Exceptions/InvalidEventHubException.cs
```diff
@@ -0,0 +1,29 @@
...
+    public sealed class InvalidEventHubException : IomtTelemetryFormattableException
+    {
+        public InvalidEventHubException(
+            string message,
+            Exception innerException,
+            string errorName)
+            : base(
+                  message,
+                  innerException,
+                  name: $"{ErrorType.EventHubError}{errorName}",
```
pallar-ms, line 20: nit: Can just use ErrType now so that if that changes for whatever reason, we won't have to remember to change here too.

src/lib/Microsoft.Health.Events/Telemetry/Exceptions/EventHubExceptionTelemetryProcessor.cs
```diff
@@ -14,35 +14,81 @@ namespace Microsoft.Health.Events.Telemetry.Exceptions
...
+                            break;
+                        case EventHubsException.FailureReason.ResourceNotFound:
+                            message = "Verify that the provided Event Hubs Namespace contains the provided Event Hub and that the provided Event Hub contains the provided consumer group.";
+                            break;
+                        case EventHubsException.FailureReason.ServiceCommunicationProblem:
+                            message = "Verify that the provided Event Hub Namespace, Event Hub name, and consumer group are correct and that access permissions to the provided Event Hub have been granted.";
+                            break;
+                        default:
+                            return exception;
```
pallar-ms, line 65: Sorry didn't ask this in 1st iteration - for these exceptions that we are processing, if we should use the FailureReason/SocketErrorCode in the default case too. 
I think for exceptions that we are not processing having a metric `EventHubErrorGeneralError` is fine, since we don't know what the exception type is, so in the default case at line 91 this metric is fine.
However for the defaults within each of these exceptions it would be nice to still have the FailureReason/SocketErrorCode captured in the metric (instead of just GeneralError) too so we can get visibility into what the various types of errors we run into.

src/lib/Microsoft.Health.Events/Telemetry/Exceptions/EventHubConfigurationExceptionTelemetryProcessor.cs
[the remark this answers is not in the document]
kyclai: Added check since pallar-ms confirmed that Operation would be used as a filter.

src/lib/Microsoft.Health.Events/Telemetry/Exceptions/EventHubConfigurationExceptionTelemetryProcessor.cs
```diff
@@ -0,0 +1,63 @@
...
+            var exType = ex.GetType();
+
+            var lookupType = exType.IsGenericType ? exType.GetGenericTypeDefinition() : exType;
+
+            if (_handledExceptions.Contains(lookupType))
```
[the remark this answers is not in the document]
kyclai: Since we need to spend some more time thinking about handled exceptions and applying the `IHandle...` interface, I'm going to hold off applying your suggestion for now and I'll revisit it when we make these code changes.
(Also, since we don't expect subclasses of `InvalideEventHubException` and `UnauthorizedAccessEventHubExceptions`, since your comment I've marked these two classes with `sealed`.)

src/lib/Microsoft.Health.Events/Telemetry/Exceptions/EventHubExceptionTelemetryProcessor.cs
kyclai: Good idea. I initially thought the inner exception would tell us the reason/socketError but having it shown in the Metric Name would be helpful in identifying, at a glance, errors we haven't come across yet.
Updated code accordingly.

src/lib/Microsoft.Health.Events/Telemetry/Exceptions/EventHubExceptionTelemetryProcessor.cs
```diff
@@ -14,36 +14,105 @@ namespace Microsoft.Health.Events.Telemetry.Exceptions
...
+                    string helpLink = "https://docs.microsoft.com/azure/event-hubs/authenticate-application";
+                    return new UnauthorizedAccessEventHubException(message, exception, helpLink, nameof(EventHubErrorCode.AuthorizationError));
+
                 default:
-                    return $"{ErrorType.EventHubError}{EventHubErrorCode.GeneralError}";
+                    return exception;
             }
         }
+
+        private static string GetErrorMetricName(Exception exception)
```
pallar-ms, line 96: As discussed, can avoid recasting the exceptions and have the previous one also return the errorname additionally.

src/lib/Microsoft.Health.Events/Telemetry/Exceptions/EventHubExceptionTelemetryProcessor.cs
pallar-ms: Would you be able to create this tracking backlog item?

pallar-ms (approved):

rogordon01 (approved):

src/lib/Microsoft.Health.Events/Telemetry/Exceptions/EventHubExceptionTelemetryProcessor.cs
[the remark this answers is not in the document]
kyclai: **Rephrasing background information:**
A receiver (consumer) hits an `EventHubsException` with `FailureReason.ConsumerDisconnected` if it can't read from a partition due to another receiver with a set higher "owner level" being the exclusive receiver/reader of the partition for the consumer group. (The owner level is a value to indicate a receiver's relative priority in being allowed to be the exclusive receiver.)

**Encountering `FailureReason.ConsumerDisconnected` when testing locally:** (Thanks, pallar-ms, for helping me understand this!)
I deployed a Gen1 test environment via [Sandbox](https://github.com/microsoft/iomt-fhir/blob/master/docs/Sandbox.md). Then, in my `appsettings.json`, I modified only the "InputEventHub:ConnectionString" to point at the devicedata Event Hub in the Gen1 test environment's Event Hubs Namespace (and updated the SharedAccessKey accordingly). However, since I ran my Console app locally, it acted as a receiver (a key point I missed before) and thus was competing with the Gen1 Function App as a receiver for the same devicedata Event Hub. I hit `FailureReason.ConsumerDisconnected` since my local app lost the owner level battle and couldn't draw events. Therefore, in contrast, once I "Stop" the Function App via Portal, I do not hit the `FailureReason.ConsumerDisconnected` anymore. (Initially I was also trying to understand why, if I specify a new consumer group in `appsettings.json`'s "InputEventHub:EventHubConsumerGroup", I was not hitting the `FailureReason.ConsumerDisconnected` - well, my local app was not competing with other consumers.)

**Testing with Gen2 test environment:**
pallar-ms and I tried to reproduce the `FailureReason.ConsumerDisconnected` in a somewhat similar scenario by pointing two IoT connectors deployed to read from the same devicedata Event Hub (by pointing the normalization pods to the same devicedata Event Hub), but we did not encounter it.

**Result:**
Since we have encountered the `FailureReason.ConsumerDisconnected` only with a local app and Gen1 Function App running at the same time and pointing to the same Event Hub, which is not a realistic scenario in production, I removed the `FailureReason.ConsumerDisconnected` case in my last commit of this PR. If we encounter it in the future we can observe how it was triggered and customize the exception then.

PR #144, opened 2021-11-09
Custom MI related exception

pallar-ms:
Changes include:
* Add new custom exceptions to summarize managed identity related errors
  - if the custom auth type is used and the credential bundle for the managed identity is not present in the key vault
  - if the underlying mi is no longer present
  - use localized strings for the exception message


src/lib/Microsoft.Health.Events/Resources/EventResources.resx
```diff
@@ -0,0 +1,144 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
rogordon01, line 1: Is this file needed?

src/lib/Microsoft.Health.Events/Resources/EventResources.Designer.cs
```diff
@@ -0,0 +1,135 @@
+﻿//------------------------------------------------------------------------------
```
rogordon01, line 1: Is this file needed?

src/lib/Microsoft.Health.Events/Microsoft.Health.Events.csproj
```diff
@@ -39,4 +39,17 @@
     <ProjectReference Include="..\Microsoft.Health.Common\Microsoft.Health.Common.csproj" />
     <ProjectReference Include="..\Microsoft.Health.Logger\Microsoft.Health.Logging.csproj" />
   </ItemGroup>
+  <ItemGroup>
```
rogordon01, line 42: Not sure if we needed these files. If not you can revert this change

src/lib/Microsoft.Health.Events/Telemetry/Exceptions/EventHubExceptionTelemetryProcessor.cs
```diff
@@ -68,24 +72,40 @@ public static (Exception customException, string errorName) CustomizeException(E
...
+                    var msalErrorCode = ((MsalServiceException)exception).ErrorCode;
+                    message = EventResources.ManagedIdentityAuthenticationError;
+                    return (new ManagedIdentityAuthenticationError(message, exception, msalErrorCode), msalErrorCode);
+
+                case RequestFailedException _:
+                    var errorCode = ((RequestFailedException)exception).ErrorCode;
+
+                    if (string.Equals(errorCode, EventResources.SecretNotFound, StringComparison.CurrentCultureIgnoreCase) || message.Contains(EventResources.SecretNotFound, StringComparison.CurrentCultureIgnoreCase))
+                    {
+                        message = EventResources.ManagedIdentityCredentialNotFound;
```
rogordon01, line 89: If we get here is it always guaranteed that we have an issue with Managed Identity? Looking at the code I think that we also support [ConnectionString](https://microsofthealth.visualstudio.com/Health/_git/workspace-platform?path=/iot-connector/core/iomt-fhir/Microsoft.Health.Fhir.Ingest.Console.PaaS/EventHubs/EventHubConsumerClientFactory.cs&version=GBmain&line=32&lineEnd=33&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents).

src/lib/Microsoft.Health.Events/Resources/EventResources.resx
pallar-ms: Yes, I added this resource file to localize the strings that we expose to the customers, in this case I've added the custom exception messages.

src/lib/Microsoft.Health.Events/Telemetry/Exceptions/EventHubExceptionTelemetryProcessor.cs
pallar-ms: If using the ConnectionString auth type, since there is no secret involved that we fetch from the key vault, I think only in the custom auth type, we use the credential bundle and can expect to get this error if it doesn't exist in the keyvault.

rogordon01 (approved):

namalu (approved):

src/lib/Microsoft.Health.Events/Microsoft.Health.Events.csproj
pallar-ms: Yes, I added this resource file to localize the strings that we expose to the customers, in this case I've added the custom exception messages.

PR #145, opened 2021-11-12
refactor: Exception telemetry processor

pallar-ms:
Changes include:
* Creating a base exception telemetry processor, that can be extended based on fhir exceptions, normalization, etc.
* Removing EventHubConfigurationTelemetryProcessor as it wasn't really using the "HandleException" but only logging the associated exception error metric.


src/lib/Microsoft.Health.Events/Telemetry/Exceptions/EventHubConfigurationExceptionTelemetryProcessor.cs
```diff
@@ -1,67 +0,0 @@
-﻿// -------------------------------------------------------------------------------------------------
```
pallar-ms, line 1: Removing this class as the list of handledexceptions are not really handleable based on the api's contract, in that it cannot be continued upon. In any case, the caller to this wasn't utilizing the return value of HandleException but was just logging the associated error metric, so updated that to just call the log api instead.
When we do identify exceptions that are handledable we should add those in here instead.

pallar-ms:
I also have some files to re-arrange, eg: moving the exceptions outside the telemetry folder since they aren't really telemetry exceptions, etc. but will do that in 2nd iteration so that the 1st iteration can show the actual diff in these files for review.

dustinburson (approved):

Issue #82, opened 2020-12-09
QuantityFhirValueProcessor fails with multiple measurment data


giacomodeliberali:
When multiple data inside the same measurment with a ValueQuantity FHIR mapping is proccessed, an exception is thrown.

*QuantityFhirValueProcessor* is assuming only a value is present. We would like to choose between some strategies (first, last or some aggregation).

https://github.com/microsoft/iomt-fhir/blob/4390667cc8e77216dc93e8b397efe5ed008911f4/src/lib/Microsoft.Health.Fhir.R4.Ingest/Template/QuantityFhirValueProcessor.cs#L15-L30

dustinburson:
The failure is by design but spinning is not.  Spinning was addressed in this commit https://github.com/microsoft/iomt-fhir/commit/a134481872bd85894f01df6b88843312e6030109

PR #146, opened 2021-11-15
fix: QuantityFhirValueProcessor format exceptions cause connector to spin on bad data

pallar-ms:
Bug: QuantityFhirValueProcessor format exceptions cause connector to spin on bad data
Issue: When an invalid data value such as "NaN" is passed as a QuantityFhirValue, we are unable to parse it to decimal and the exception thrown is not handled, causing the processor to continuously retry on this exception.
Fix: Add custom exception for invalid data quantity type values. Throw this custom exception for invalid data and add it to the list of handleable FHIR exceptions that will just log the associated error metric and continue processing subsequent event batches.

Note: This PR is currently based off of a refactor PR - https://github.com/microsoft/iomt-fhir/pull/145

### DIFF
--- a/src/lib/Microsoft.Health.Fhir.R4.Ingest/Template/QuantityFhirValueProcessor.cs
+++ b/src/lib/Microsoft.Health.Fhir.R4.Ingest/Template/QuantityFhirValueProcessor.cs
@@ -9,6 +9,7 @@ using System.Globalization;
 using System.Linq;
 using EnsureThat;
 using Hl7.Fhir.Model;
+using Microsoft.Health.Fhir.Ingest.Service;
 
 namespace Microsoft.Health.Fhir.Ingest.Template
 {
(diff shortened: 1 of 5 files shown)


src/lib/Microsoft.Health.Fhir.R4.Ingest/Template/QuantityFhirValueProcessor.cs
```diff
@@ -20,9 +21,21 @@ protected override DataType CreateValueImpl(QuantityFhirValueType template, IObs
             EnsureArg.IsNotNull(inValue, nameof(inValue));
             IEnumerable<(DateTime, string)> values = EnsureArg.IsNotNull(inValue.Data, nameof(IObservationData.Data));
 
+            decimal value;
+            try
+            {
+                value = decimal.Parse(values.Single().Item2, CultureInfo.InvariantCulture);
```
pallar-ms, line 27: Let me know if there are multiple values then if that should be considered a non handleable situation.

dustinburson: Currently we throw an exception I believe, so handling that scenario as part of this change seems appropriate. 

We do have an open issue https://github.com/microsoft/iomt-fhir/issues/82 to allow different strategies but outside the scope of this PR.

dustinburson (approved):

PR #147, opened 2021-11-17
[Bug 86283]: Add "Codes" Validation to FHIR Mapping

kenziedolish:
**Issue**: [Bug 86283](https://microsofthealth.visualstudio.com/Health/_workitems/edit/86283)

**Purpose of PR**: Add to template validation a check to make sure the code is defined on FHIR mappings

**Code Changes:**
- updated CodeValueFhirTemplate.cs and CodeValueMapping.cs by making "Codes" a required field
- added GivenInvalidTemplateJsonMissingCodeField_WhenFactoryCreate_ThenTemplateCreated_Test and input file to test that template validation exception is thrown when "Codes" is missing
- updated GivenValidTemplateJsonWithValueStringType test and input


src/lib/Microsoft.Health.Fhir.Ingest.Template/FhirCodeableConcept.cs
```diff
@@ -12,6 +13,7 @@ public class FhirCodeableConcept
         public string Text { get; set; }
 
 #pragma warning disable CA2227
+        [JsonProperty(Required = Required.Always)]
         public IList<FhirCode> Codes { get; set; }
```
dustinburson, line 17: I don't think this is correct.  The CodeableConcept in FHIR allows 0 to N codings and 0 to 1 text element.

https://www.hl7.org/fhir/datatypes.html#CodeableConcept so it is valid for someone to supply a text element but not supply any codes.

dustinburson: Same comment applies to CodeableConceptFhirValueType.cs above.

src/lib/Microsoft.Health.Fhir.Ingest.Template/CodeValueFhirTemplate.cs
```diff
@@ -14,6 +15,7 @@ public class CodeValueFhirTemplate : FhirTemplate
 #pragma warning restore CA2227
 
 #pragma warning disable CA2227
+        [JsonProperty(Required = Required.Always)]
         public virtual IList<FhirCode> Codes { get; set; }
```
dustinburson, line 19: Ideally we would replace this with the CodeableConcept type and make that required which would mean they user would either need to supply the Text or at least one code element with in the codeable concept.  Unfortunately we can't just make that change without factoring in backwards compatibility.  

No actual action item, I am fine proceeding with this as an intermediate solution till we can support the full codeable concept in a backwards compatible manner.

dustinburson (changes requested):
I believe we need to remove the changes to require codes on the codeable concept class (and the value type that inherits from it) to conform with the FHIR specification.

src/lib/Microsoft.Health.Fhir.Ingest.Template/FhirCodeableConcept.cs
kenziedolish: Thanks for your feedback. I've further explored the CodeableConcept docs and see that 0-N codings are allowed. I've removed the required attribute on Codes from FhirCodeableConcept.cs and CodeableConceptFhirValueType.cs

dustinburson (approved):

pallar-ms:
Sorry late to this PR - but do we want to also check if the code value is not null or empty.

PR #148, opened 2021-11-17
fix: Incompatible event data does not cause normalization processing to stall

pallar-ms:
Bug: If there is an incompatibility b/w the event data and the template, then we end up stalling the processing until the data/template is fixed.
Issue: When the above situation occurs, we throw an exception which is currently set to retry forever, thereby stalling the entire processing of the subsequent data. 
Fix: For such cases, we should log an error (which will be exposed to the customer), drop these events and continue processing the remaining events. 
(A longer-term solution that will prevent the customer from resending the rectified events is to have a dead letter queue where they can an action from for such dropped events.)
Changes include:
* Adding a custom exception to throw when there is incompatible data and we cannot evaluate the expression.
* Add a normalization exception telemetry processor that handles this exception (logs the associated error metric).
* Updated UTs

Note: This PR is currently based off of a refactor PR - #145

### DIFF
--- a/test/Microsoft.Health.Fhir.Ingest.UnitTests/Service/MeasurementEventNormalizationServiceTests.cs
+++ b/test/Microsoft.Health.Fhir.Ingest.UnitTests/Service/MeasurementEventNormalizationServiceTests.cs
@@ -10,6 +10,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Azure.EventHubs;
 using Microsoft.Health.Fhir.Ingest.Data;
+using Microsoft.Health.Fhir.Ingest.Telemetry;
 using Microsoft.Health.Fhir.Ingest.Template;
 using Microsoft.Health.Logging.Telemetry;
 using Newtonsoft.Json.Linq;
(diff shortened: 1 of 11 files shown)


test/Microsoft.Health.Fhir.Ingest.UnitTests/Service/MeasurementEventNormalizationServiceTests.cs
```diff
@@ -130,68 +123,72 @@ public async Task GivenEvents_WhenProcessAsync_ThenIngestionTimeUtcSet_Test()
...
-            converter.ReceivedWithAnyArgs(events.Length).Convert(null);
-            await consumer.Received(2).AddAsync(Arg.Is<IEnumerable<IMeasurement>>(l => l.Count() == 5), Arg.Any<CancellationToken>());
+            _template.ReceivedWithAnyArgs(events.Length).GetMeasurements(null);
+            _converter.ReceivedWithAnyArgs(events.Length).Convert(null);
+            await _consumer.Received(2).AddAsync(Arg.Is<IEnumerable<IMeasurement>>(l => l.Count() == 5), Arg.Any<CancellationToken>());
         }
 
         [Fact]
-        public async Task GivenEventsAndOperationCancellationException_WhenProcessAsync_ThenExecutionHalted_Test()
+        public async Task GivenEventsAndDefaultErrorConsumer_WhenProcessAsyncAndHandleableConsumerErrors_ThenExceptionNotThrown_Test()
```
pallar-ms, line 142: This is the new test added in this file.

dustinburson (approved):

dustinburson:
Looks like there are some unit test failures that need to be accounted for but overall the changes look good.

PR #150, opened 2021-12-08
Customized FHIR service configuration exceptions

kyclai:
Added custom exceptions for the following scenarios:
- Invalid FHIR service URL (e.g. not a URL for a FHIR server URL, not an absolute URL, etc.)
- Unauthorized access to create/update FHIR resources (i.e. FHIR Data Writer role is missing)

Also renamed "FhirServerError" to "FhirServiceError" for consistency with product name "FHIR service".


src/lib/Microsoft.Health.Extensions.Fhir.R4/FhirServiceValidator.cs
```diff
@@ -0,0 +1,32 @@
...
+using EnsureThat;
+using Hl7.Fhir.Rest;
+using Microsoft.Health.Extensions.Fhir.Telemetry.Exceptions;
+using Microsoft.Health.Logging.Telemetry;
+
+namespace Microsoft.Health.Extensions.Fhir
+{
+    public static class FhirServiceValidator
+    {
+        public static bool ValidateFhirService(FhirClient client, ITelemetryLogger logger)
```
pallar-ms, line 16: I don't see us utilizing this bool returned.

src/lib/Microsoft.Health.Extensions.Fhir.R4/FhirClientFactory.cs
```diff
@@ -64,18 +65,29 @@ public FhirClient Create()
...
 
-            var client = new FhirClient(url, fhirClientSettings, new BearerTokenAuthorizationMessageHandler(uri, tokenCredential, logger));
+            FhirClient client;
+            try
+            {
+                client = new FhirClient(url, fhirClientSettings, new BearerTokenAuthorizationMessageHandler(uri, tokenCredential, logger));
+            }
+            catch (Exception ex)
+            {
+                client = null;
```
pallar-ms, line 86: Do we actually get back a certain value for `client` in case of such an exception?

src/lib/Microsoft.Health.Extensions.Fhir.R4/FhirClientFactory.cs
```diff
@@ -64,18 +65,29 @@ public FhirClient Create()
...
+                FhirServiceExceptionProcessor.ProcessException(ex, logger);
+            }
+
+            FhirServiceValidator.ValidateFhirService(client, logger);
```
pallar-ms, line 90: Should we check against the returned bool if it succeeded?

src/lib/Microsoft.Health.Extensions.Fhir.R4/FhirClientFactory.cs
```diff
@@ -117,15 +129,16 @@ protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage
 
                 if (Logger != null && !response.IsSuccessStatusCode)
                 {
+                    var errorType = ErrorType.FHIRServiceError;
                     var statusDescription = response.ReasonPhrase.Replace(" ", string.Empty);
 
                     if (response.StatusCode == System.Net.HttpStatusCode.TooManyRequests)
                     {
-                        Logger.LogMetric(FhirClientMetrics.HandledException($"FhirServerError{statusDescription}", ErrorSeverity.Informational, ConnectorOperation.FHIRConversion), 1);
+                        Logger.LogMetric(FhirClientMetrics.HandledException($"{errorType}{statusDescription}", ErrorSeverity.Informational), 1);
```
pallar-ms, line 137: nit: Sounds like only the severity is different in this if else clause so can simplify to :

var severity = response.StatusCode == System.Net.HttpStatusCode.TooManyRequests ? ErrorSeverity.Informational : ErrorSeverity.Critical;
Logger.LogMetric(...);

And probably then don't need that extra var for errorType.

src/lib/Microsoft.Health.Extensions.Fhir.R4/FhirClientFactory.cs
pallar-ms: If the above fails and client = null, can we avoid calling the ValidateFhirService()

src/lib/Microsoft.Health.Extensions.Fhir.R4/Telemetry/Exceptions/FhirServiceExceptionProcessor.cs
```diff
@@ -0,0 +1,82 @@
...
+            {
+                case FhirOperationException _:
+                    var status = ((FhirOperationException)exception).Status;
+                    switch (status)
+                    {
+                        case HttpStatusCode.Forbidden:
+                            message = FhirResources.FhirServiceAccessForbidden;
+                            string helpLink = "https://docs.microsoft.com/azure/healthcare-apis/iot/deploy-iot-connector-in-azure#accessing-the-iot-connector-from-the-fhir-service";
+                            errorName = nameof(FhirServiceErrorCode.AuthorizationError);
+                            return (new UnauthorizedAccessFhirServiceException(message, exception, helpLink, errorName), errorName);
```
pallar-ms, line 52: Does the original exception also list the name/url of the fhir service?

src/lib/Microsoft.Health.Extensions.Fhir.R4/Telemetry/Exceptions/FhirServiceExceptionProcessor.cs
```diff
@@ -0,0 +1,82 @@
...
+
+                case UriFormatException _:
+                    message = FhirResources.FhirServiceUriFormatInvalid;
+                    errorName = nameof(FhirServiceErrorCode.ConfigurationError);
+                    return (new InvalidFhirServiceException(message, exception, errorName), errorName);
+
+                case HttpRequestException _:
+                    message = FhirResources.FhirServiceHttpRequestError;
+                    errorName = nameof(FhirServiceErrorCode.ConfigurationError);
+                    return (new InvalidFhirServiceException(message, exception, errorName), errorName);
```
pallar-ms, line 69: Should we look for a specific error code in HttpRequestException? I wonder if it is too broad to assume that it is a config error, as even intermittent 503s etc could throw this?

src/lib/Microsoft.Health.Extensions.Fhir.R4/Telemetry/Exceptions/FhirServiceExceptionProcessor.cs
```diff
@@ -0,0 +1,82 @@
...
+
+                case MsalServiceException _:
+                    string errorCode = ((MsalServiceException)exception).ErrorCode;
+                    message = FhirResources.FhirServiceMsalServiceError;
+                    errorName = $"{nameof(FhirServiceErrorCode.ConfigurationError)}_{errorCode}";
```
pallar-ms, line 74: Should this be auth error instead?

src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4FhirImportService.cs
```diff
@@ -63,17 +64,29 @@ public virtual async Task<string> SaveObservationAsync(ILookupTemplate<IFhirTemp
...
+                }
+
                 _logger.LogMetric(IomtMetrics.FhirResourceSaved(ResourceType.Observation, ResourceOperation.Created), 1);
             }
             else
             {
                 var policyResult = await Policy<Model.Observation>
-                     .Handle<FhirOperationException>(ex => ex.Status == System.Net.HttpStatusCode.Conflict || ex.Status == System.Net.HttpStatusCode.PreconditionFailed)
+                     .Handle<FhirOperationException>(ex =>
+                         ex.Status == System.Net.HttpStatusCode.Forbidden
```
pallar-ms, line 87: I'm not sure I understand why the Forbidden case was added here, I don't expect us to retry for this error?

pallar-ms (changes requested):

src/lib/Microsoft.Health.Extensions.Fhir.R4/FhirClientFactory.cs
kyclai: `BaseFhirClient` throws `ArgumentNullException` or `ArgumentException` if the endpoint is invalid.

In commit 9ab3837, I've initialized `client` to null.

src/lib/Microsoft.Health.Extensions.Fhir.R4/FhirClientFactory.cs
kyclai: In this scenario, I don't think checking the returned bool is needed for any reason.

src/lib/Microsoft.Health.Extensions.Fhir.R4/FhirServiceValidator.cs
kyclai: The bool's currently used only in the unit test for this function. I'm also thinking that this function can be used elsewhere in the future.

src/lib/Microsoft.Health.Extensions.Fhir.R4/Telemetry/Exceptions/FhirServiceExceptionProcessor.cs
kyclai: No, `FhirOperationException` currently has properties for only `Outcome` (operation outcome) and `Status` (HTTP status code).
We discussed offline potentially adding the URL to the custom exception message, but I have not implemented this update.

src/lib/Microsoft.Health.Extensions.Fhir.R4/Telemetry/Exceptions/FhirServiceExceptionProcessor.cs
kyclai: `HttpRequestException`'s error code property, `StatusCode`, applies to only .NET 5 and later.

In commit 9ab3837, I've scoped `HttpRequestException` down by checking the `Message` property for now and added a TODO to check for `StatusCode` in .NET 5 and later.

src/lib/Microsoft.Health.Extensions.Fhir.R4/Telemetry/Exceptions/FhirServiceExceptionProcessor.cs
kyclai: From my testing, no, because `MsalServiceException` was given for invalid FHIR service URLs such as https://clai1015.azurehealthcareapis.com/metadata, https://clai1015.azurehealthcareapis.com///, https://sdfsldfkwr389lksnldsf.com

In commit 9ab3837, I've scoped `MsalServiceException` down by checking the `ErrorCode` property.

src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4FhirImportService.cs
kyclai: I was trying to catch the FHIR service authorization failure here, but I was wrong.

In commit 4810905, I've reverted this change and updated where to catch the FHIR service authorization failure.

src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4FhirImportService.cs
```diff
@@ -42,14 +43,21 @@ public R4FhirImportService(IResourceIdentityService resourceIdentityService, Fhi
...
-            foreach (var grp in grps)
+                foreach (var grp in grps)
+                {
+                    _ = await SaveObservationAsync(config, grp, ids).ConfigureAwait(false);
+                }
+            }
+            catch (Exception ex)
             {
-                _ = await SaveObservationAsync(config, grp, ids).ConfigureAwait(false);
+                FhirServiceExceptionProcessor.ProcessException(ex, _logger);
```
pallar-ms, line 60: I think we should retain the previous behaviour and not eat the exception?

test/Microsoft.Health.Extensions.Fhir.R4.UnitTests/FhirServiceExceptionProcessorTests.cs
```diff
@@ -0,0 +1,102 @@
...
+
+            logger.ReceivedWithAnyArgs(1).LogError(ex);
+            logger.Received(1).LogMetric(
+                Arg.Is<Metric>(m =>
+                    m.Name.Equals(expectedErrorMetricName) &&
+                    m.Dimensions[DimensionNames.Name].Equals(expectedErrorMetricName) &&
+                    m.Dimensions[DimensionNames.Operation].Equals(ConnectorOperation.FHIRConversion) &&
+                    m.Dimensions[DimensionNames.Category].Equals(Category.Errors) &&
+                    m.Dimensions[DimensionNames.ErrorType].Equals(ErrorType.FHIRServiceError) &&
+                    m.Dimensions[DimensionNames.ErrorSeverity].Equals(ErrorSeverity.Critical) &&
```
pallar-ms, line 88: nit: Maybe this can go into a separate method like ValidateFHIRServiceErrorProperteies() similar to a recent change you made in the other test file.

pallar-ms (approved):

src/lib/Microsoft.Health.Extensions.Fhir.R4/FhirClientFactory.cs
```diff
@@ -64,18 +65,27 @@ public FhirClient Create()
...
             {
                 PreferredFormat = ResourceFormat.Json,
             };
 
-            var client = new FhirClient(url, fhirClientSettings, new BearerTokenAuthorizationMessageHandler(uri, tokenCredential, logger));
+            FhirClient client = null;
+            try
+            {
+                client = new FhirClient(url, fhirClientSettings, new BearerTokenAuthorizationMessageHandler(uri, tokenCredential, logger));
+                FhirServiceValidator.ValidateFhirService(client, logger);
```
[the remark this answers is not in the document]
kyclai: In this case (when `ValidateFhirService` catches an exception, gets the exception logged, and returns false), `FhirClient` is returned (Line 90), as before these code changes.

src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4FhirImportService.cs
kyclai: Just clarifying - do you mean to wrap only the line with `ResolveResourceIdentitiesAsync` in the try-block, so that the other lines are still executed if we do get an exception thrown (for an authorization error) from `ResolveResourceIdentitiesAsync`?

src/lib/Microsoft.Health.Extensions.Fhir.R4/FhirClientFactory.cs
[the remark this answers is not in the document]
pallar-ms: Sure but I was trying to understand that if this validation failed and we still returned the client, then what happens in the ingestion flow? Do we encounter other exceptions that block the ingestion?

src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4FhirImportService.cs
pallar-ms: To retain the previous behaviour where this function would throw if we encountered an exception, we can just throw it back after processing the exception. Unless we feel like throwing it back doesn't make sense anymore? I'm worried that if we were previously retrying on this exception, then by eating the exception now, we will not be and if that is desired.

kyclai: Oh I see what you mean, thanks for the details. I think you bring up a good point; I'll re-throw while keeping the stack trace.

src/lib/Microsoft.Health.Extensions.Fhir.R4/FhirClientFactory.cs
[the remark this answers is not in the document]
kyclai: As discussed offline, what happens when an invalid client is returned is something to further investigate later. For now, the previous behaviour is still retained: a client, whether valid or not, gets returned.

PR #151, opened 2021-12-14
Updated custom Event Hub exceptions

kyclai:
Updated custom Event Hub configuration exceptions:
- Checked `InvalidOperationException`'s exception message when processing Event Hub exceptions, since `InvalidOperationException` 'can be thrown in a wide variety of circumstances' (unfortunately `InvalidOperationException` currently does not have a property such as error code that can be checked instead)
- Specified connector operation/stage in Event Hub configuration exceptions
- Added the three common constructors for custom exceptions (as per https://docs.microsoft.com/en-us/dotnet/standard/exceptions/best-practices-for-exceptions#include-three-constructors-in-custom-exception-classes)

### DIFF
--- a/src/lib/Microsoft.Health.Events/Telemetry/Exceptions/InvalidEventHubException.cs
+++ b/src/lib/Microsoft.Health.Events/Telemetry/Exceptions/InvalidEventHubException.cs
@@ -16,12 +16,13 @@ namespace Microsoft.Health.Events.Telemetry.Exceptions
         public InvalidEventHubException(
             string message,
             Exception innerException,
-            string errorName)
+            string errorName,
+            string operation)
             : base(
                   message,
                   innerException,
                   name: $"{_errorType}{errorName}",
-                  operation: ConnectorOperation.Setup)
+                  operation: operation)
         {
         }
 
--- a/src/lib/Microsoft.Health.Events/Telemetry/Exceptions/InvalidEventHubException.cs
+++ b/src/lib/Microsoft.Health.Events/Telemetry/Exceptions/InvalidEventHubException.cs
@@ -30,13 +30,12 @@ namespace Microsoft.Health.Events.Telemetry.Exceptions
         public InvalidEventHubException(
             string message,
             Exception innerException,
-            string errorName,
-            string operation)
+            string errorName)
             : base(
                   message,
                   innerException,
                   name: $"{_errorType}{errorName}",
-                  operation: operation)
+                  operation: ConnectorOperation.Setup)
         {
         }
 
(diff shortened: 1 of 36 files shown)


namalu (approved):

src/lib/Microsoft.Health.Events/Telemetry/Metrics/EventMetrics.cs
```diff
@@ -21,16 +21,21 @@ public static class EventMetrics
         private static readonly string _operationDimension = DimensionNames.Operation;
         private static readonly string _reasonDimension = DimensionNames.Reason;
 
-        private static string _connectorOperation = ConnectorOperation.Unknown;
+        private static string _connectorOperation = Microsoft.Health.Common.Telemetry.ConnectorOperation.Unknown;
 
-        public static void SetConnectorOperation(string connectorOperation)
+        public static string ConnectorOperation
```
pallar-ms, line 26: nit: Just curious - any reason to change to property? 
Personally speaking, its not intuitive that properties are one time set so its easy to mistake this as an updateable one until we hit a runtime exception. But that's just my thinking. 
Or maybe consider adding a comment indicating this is a one time set property so that it shows in intellisense, etc.

src/lib/Microsoft.Health.Events/Telemetry/Exceptions/InvalidEventHubException.cs
```diff
@@ -13,15 +13,30 @@ public sealed class InvalidEventHubException : IomtTelemetryFormattableException
...
             string message,
             Exception innerException,
-            string errorName)
+            string errorName,
+            string operation)
             : base(
                   message,
                   innerException,
                   name: $"{_errorType}{errorName}",
-                  operation: ConnectorOperation.Setup)
```
pallar-ms, line 24: I actually preferred these exceptions being categorized as Setup as opposed to Normalization/FhirTransformation to differentiate from actual operation issues post setup. Just curious on why we want to change this?
Adding @wi-y to add any thoughts as some of the earlier related metrics were also intentionally set with Setup.

pallar-ms (changes requested):

src/lib/Microsoft.Health.Events/Telemetry/Metrics/EventMetrics.cs
kyclai: In commit 65a6c34, since I needed a `get` accessor for the operation to specify the operation as Normalization/FhirTransformation, I thought I'd convert `SetConnectorOperation` to a `set` accessor to pair the accessors up under one `ConnectorOperation` property.

I reverted commit 65a6c34 in commit 2a6520f (see other comment), so this change got reverted, but I'll keep your point in mind in the future, thanks!

pallar-ms (approved):

PR #155, opened 2022-01-07
Data Mapper: Bug fixes

kyclai:
Fixed Data Mapper bugs:

- Specified a mapping's current device and FHIR mappings when renaming the mapping; otherwise undefined device and FHIR mappings get saved over the current ones. Fixes a bug where renaming a mapping's type name (either from the mapping page or from the home page) cleared all the Device Mapping and FHIR Mapping fields; when the user refreshed or navigated to the mapping page, the fields were shown empty.

   bug_rename

- Corrected FhirValue property for Quantity. Fixes a bug where, when ValueType is "Quantity", adding or deleting a component cleared the Code, System, and/or Unit field(s).

   bug_quantity


ms-teli (approved):

ms-teli:
Looks good to me, thanks for making the change!

PR #156, opened 2022-01-11
Throwing an exception if supplied field is null and is marked required

rogordon01:
We currently throw an exception if a required field is not present within the device event. This PR throws an exception when the field is supplied,  is `null` and is marked as required.

### DIFF
--- a/src/lib/Microsoft.Health.Fhir.Ingest.Template/MeasurementExtractor.cs
+++ b/src/lib/Microsoft.Health.Fhir.Ingest.Template/MeasurementExtractor.cs
@@ -64,11 +64,11 @@ namespace Microsoft.Health.Fhir.Ingest.Template
                         try
                         {
                             var value = evaluatedToken.Value<T>();
-                            var isNull = value is string s ? string.IsNullOrEmpty(s) : value == null;
+                            var isNull = value is string s ? string.IsNullOrWhiteSpace(s) : value == null;
 
                             if (isRequired && isNull)
                             {
-                                exceptions.Add(new IncompatibleDataException($"A null value was supplied"));
+                                exceptions.Add(new IncompatibleDataException($"A null value was supplied for [{name}]"));
                             }
                             else
                             {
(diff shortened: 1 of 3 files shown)


src/lib/Microsoft.Health.Fhir.Ingest.Template/MeasurementExtractor.cs
```diff
@@ -63,7 +63,16 @@ protected T EvalExpression<T>(JToken token, string name, bool isRequired = false
                     {
                         try
                         {
-                            return evaluatedToken.Value<T>();
+                            var value = evaluatedToken.Value<T>();
+
+                            if (isRequired && value == null)
```
rogordon01, line 68: I stopped short of checking for blank/empty strings here. The reason being is that a customer would have to explicitly set that value. And if they went to the length to do that they may want that value to flow through the system. I'm not sure of a scenario where the actually makes sense, though. Thoughts?

dustinburson (approved):

src/lib/Microsoft.Health.Fhir.Ingest.Template/MeasurementExtractor.cs
namalu: *"Will require the value to be present in the payload. If not found a measurement will not be generated and an InvalidOperationException will be thrown."*

I think our definition of "required" is a little ambiguous, and we should decide if empty satisfies the requirement and update appropriately. Because empty doesn't cleanly map to an int or decimal, I think that we should count empty as violating the "required" flag. We could have a different behavior for strings, but I'm just not sure it makes sense to create an observation with no value.

namalu (approved):

src/lib/Microsoft.Health.Fhir.Ingest.Template/MeasurementExtractor.cs
rogordon01: Based on @namalu comments, perhaps we can also 1) Check if the value is a `string` and 2) check if it is null or empty/blank? We can also update our documentation to clearly state as such:

"Will require the value to be present in the payload. If not found a measurement will not be generated and an InvalidOperationException will be thrown. A value is considered present when it is not `null`. If the value is a String it must also not be empty or all whitespace".

rogordon01: Done

PR #161, opened 2022-01-24
Improve retry logic to handle invalid cache and no changes to Observation

namalu:
This PR includes updates to the R4FhirImportService SaveObservationAsync logic.

## Invalid cache entry:
If an Observation has been saved to the in memory cache and deleted from the FHIR server, it may take an hour for the cache to be updated. During this time, when new data is sent, PUT operations on the deleted Observation will fail. These code updates will handle the failure by removing the invalid Observation from the cache, creating a new Observation in the FHIR server, and saving the new Observation to the cache.

## Observations updated repeatedly:
If a batch of data contains one or more messages that can successfully be transformed into Observations, but also contains one or more messages that will cause an "unhandled" exception, the entire batch will be retried indefinitely. The retry will include updating the successful Observations in the FHIR Server repeatedly. These code updates will check the new Observation with the existing Observation and only perform a PUT if there is a change. This will prevent unnecessary  updates to Observations, and reduce requests to the FHIR Server.


src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4FhirImportService.cs
```diff
@@ -72,26 +72,66 @@ public virtual async Task<string> SaveObservationAsync(ILookupTemplate<IFhirTemp
...
-                    {
-                        existingObservation = await GetObservationFromServerAsync(identifier).ConfigureAwait(false);
-                    })
-                    .ExecuteAndCaptureAsync(async () =>
-                    {
-                        var mergedObservation = MergeObservation(config, existingObservation, observationGroup);
-                        return await _client.UpdateAsync(mergedObservation, versionAware: true).ConfigureAwait(false);
-                    }).ConfigureAwait(false);
+                     .Handle<FhirOperationException>(ex => ex.Status == System.Net.HttpStatusCode.Conflict || ex.Status == System.Net.HttpStatusCode.PreconditionFailed)
+                     .RetryAsync(2, async (polyRes, attempt) =>
```
rogordon01, line 90: Any particular reason for splitting the logic here between the `RetryAsync` and `ExecuteAndCaptureAsync`? I believe that `ExecuteAndCaptureAsync` runs inside of the policy created by `RetryAsync`, so I'm curious what the benefit is for the split. Personally, it makes the logic a bit hard to follow.

src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4FhirImportService.cs
```diff
@@ -72,26 +72,66 @@ public virtual async Task<string> SaveObservationAsync(ILookupTemplate<IFhirTemp
...
+                         existingObservation = await GetObservationFromServerAsync(identifier).ConfigureAwait(false);
+
+                         // If the Observation no longer exists on the FHIR Server, it was most likely deleted.
+                         if (existingObservation == null)
+                         {
+                             // Remove the Observation from the cache (this version no longer exists on the FHIR Server.
+                             _observationCache.Remove(cacheKey);
+
+                             // Create a new Observation.
+                             result = await _client.CreateAsync(mergedObservation).ConfigureAwait(false);
```
rogordon01, line 108: Is `mergedObservation` `null` at this point? If so, won't that result in a blank Observation? Since you know that there is no Observation inside of the Fhir Server could we simply create a new Observation using the data passed into this method?

```c#
// Create a new Observation.
var newObservation = GenerateObservation(config, observationGroup, identifier, ids);
result = await _client.CreateAsync(newObservation).ConfigureAwait(false);
```

src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4FhirImportService.cs
```diff
@@ -72,26 +72,66 @@ public virtual async Task<string> SaveObservationAsync(ILookupTemplate<IFhirTemp
...
+                     {
+                         if (result != null)
+                         {
+                             // If result is not null, this means that a cached Observation was deleted from the FHIR Server.
+                             // The RetryAsync has handled this condition, so no further action is required, return the result.
+                             operationType = ResourceOperation.Created;
+                             return result;
+                         }
+
+                         // Merge the new data with the existing Observation.
```
rogordon01, line 121: Could we combine this logic with what is inside of the `RetryPolicy` block above? It would make the logic easier to read.

```c#
if (existingObservation == null)
{
   /// create new Observation
  return <createdObservation>
}
else
{
// Merge the new data with the existing Observation.
 mergedObservation = MergeObservation(config, existingObservation, observationGroup);

 // Check to see if there are any changes after merging.
 if (mergedObservation.IsExactly(existingObservation))
 {
     // There are no changes to the Observation - Do not update.
     return existingObservation;
 }

 // Update the Observation. Some failures will be handled in the RetryAsync block above.
 operationType = ResourceOperation.Updated;
 return await _client.UpdateAsync(mergedObservation, versionAware: true).ConfigureAwait(false);
}

```

rogordon01 (changes requested):

src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4FhirImportService.cs
dustinburson: RetryAsync only executes on failure.  ExecuteAndCapture runs in every every attempt.

So flow is as such.  First attempt is Execute block.  If there is no failure we are done.   If there is a failure the retry block executes to refresh the observation and then we execute the ExecuteAndCapture step again.  Personally I think this makes the logic easier to follow.  We have code we always execute in an attempt and then we have specific code on an old e-tag that refreshes our observation.  The other benefit is in the future if there are other errors codes that require different retry steps they are easier integrated into the current logic.

src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4FhirImportService.cs
namalu: mergedObservation here is not null because the ExecuteAndCaptureAsync block has already run one time, which performed the merge operation. I see your point though, where if an external actor deleted the Observation, we should not create a new Observation with the data that was previously deleted. I will update this to just create a new Observation.

src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4FhirImportService.cs
```diff
@@ -72,26 +72,66 @@ public virtual async Task<string> SaveObservationAsync(ILookupTemplate<IFhirTemp
...
+                             // Remove the Observation from the cache (this version no longer exists on the FHIR Server.
+                             _observationCache.Remove(cacheKey);
+
+                             // Create a new Observation.
+                             result = await _client.CreateAsync(mergedObservation).ConfigureAwait(false);
+                         }
+                     })
+                     .ExecuteAndCaptureAsync(async () =>
+                     {
+                         if (result != null)
```
dustinburson, line 113: >result [](http://example.com/codeflow?start=29&length=6)

Since we have to account for the observation going missing and triggering the create flow it seems like we can move the newObservation logic inside the ExecuteAndCapture and remove the above if statement since we are replicating the same logic.  The code works but would be nice if we reduce to 1 create observation path.  Thoughts?

src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4FhirImportService.cs
namalu: This code has been here for a while, but I believe this is a pattern outlined by [Polly](https://github.com/App-vNext/Polly#post-execution-capturing-the-result-or-any-final-exception).

The ExecuteAndCapture block will always run first. If there is an exception the RetryAsync block runs (to try and recover from the issue)  before the ExecuteAndCapture block runs again. I personally like the syntax, maybe I can make the code easier to read.

src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4FhirImportService.cs
```diff
@@ -72,26 +72,66 @@ public virtual async Task<string> SaveObservationAsync(ILookupTemplate<IFhirTemp
...
             if (existingObservation == null)
             {
                 var newObservation = GenerateObservation(config, observationGroup, identifier, ids);
                 result = await _client.CreateAsync(newObservation).ConfigureAwait(false);
-                _logger.LogMetric(IomtMetrics.FhirResourceSaved(ResourceType.Observation, ResourceOperation.Created), 1);
+                operationType = ResourceOperation.Created;
             }
             else
             {
+                Model.Observation mergedObservation = null;
```
dustinburson, line 86: >mergedObservation [](http://example.com/codeflow?start=34&length=17)

This change + the addition of the operationType is causing the state of the entire of the outer method to be captured in an implicit class by the compiler and passed into the lambdas.

Not the end of the world but I believe we can do something similar below and avoid it.

See:
[https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/lambda-expressions#capture-of-outer-variables-and-variable-scope-in-lambda-expressions](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/lambda-expressions#capture-of-outer-variables-and-variable-scope-in-lambda-expressions)

dustinburson: It looks like I forgot about the existingObservation which was outside scope and being implicitly captured already.  If we can eliminate all of them great.  Otherwise should be fine.

src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4FhirImportService.cs
dustinburson: Agreed.  This should work pretty well if you follow the other suggestion to move the core update logic into the ExecuteAndCapture  block.

src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4FhirImportService.cs
```diff
@@ -72,26 +72,66 @@ public virtual async Task<string> SaveObservationAsync(ILookupTemplate<IFhirTemp
...
+                     .RetryAsync(2, async (polyRes, attempt) =>
+                     {
+                         // 409 Conflict or 412 Precondition Failed can occur if the Observation.meta.versionId does not match the update request.
+                         // This can happen if 2 independent processes are updating the same Observation simultaneously.
+                         // or
+                         // The update operation failed because the Observation no longer exists.
+                         // This can happen if a cached Observation was deleted from the FHIR Server.
+
+                         // Attempt to get the most recent version of the Observation.
+                         existingObservation = await GetObservationFromServerAsync(identifier).ConfigureAwait(false);
```
dustinburson, line 99: >GetObservationFromServerAsync [](http://example.com/codeflow?start=53&length=29)

I originally was wondering if we can rely on this returning null and not a 404 but after looking at the code stack we should be good.  Since we are performing a search and not a resource get the search result should always return a bundle and the underlying code checks the count and will return null if nothing is found.  If we ever switch to using a get instead of a search we will need to ensure the get method handles 404 to null.

src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4FhirImportService.cs
```diff
@@ -72,26 +72,66 @@ public virtual async Task<string> SaveObservationAsync(ILookupTemplate<IFhirTemp
...
+                             // The RetryAsync has handled this condition, so no further action is required, return the result.
+                             operationType = ResourceOperation.Created;
+                             return result;
+                         }
+
+                         // Merge the new data with the existing Observation.
+                         mergedObservation = MergeObservation(config, existingObservation, observationGroup);
+
+                         // Check to see if there are any changes after merging.
+                         if (mergedObservation.IsExactly(existingObservation))
```
dustinburson, line 125: >IsExactly [](http://example.com/codeflow?start=47&length=9)

This is part of the HL7.Fhir.Core library.  It looks like it does a deep compare on each element.  Fine using this for now but wonder if a alternative where we generate the raw json and hash it and compare might be better in the long run.

dustinburson (approved):
Added some suggestions to optimize but overall the changes look good.

src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4FhirImportService.cs
dustinburson: I like it too Nate but I am biased an wrote it originally.  The key thing about this pattern too is it extends nicely if you have different retry blocks for different exception conditions in the future.

rogordon01: Thanks for the clarification. The split of logic between the two methods + the order of execution wasn't clear but now I follow.

rogordon01 (approved):

src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4FhirImportService.cs
namalu: Thanks for the suggestion. This simplified the code quite a bit.

src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4FhirImportService.cs
namalu: I was able to easily make some changes to avoid variable capture in most cases, but the "existingObservation" issue is one that is a bit harder. I'm going to punt this one down the road.

src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4FhirImportService.cs
namalu: Interesting. We could definitely improve this code by eliminating a search and just performing a GET on the Observation directly.

dustinburson: To do that we would need to switch from using identifers to using the observation id, which we could do.  It would improve our efficiency on FHIR calls but we would need to make it optional to avoid breaking against installs that use the identifier to look up.

PR #167, opened 2022-02-07
Updated FhirClient

kenziedolish:
We currently create a FhirClient singleton and use it across threads. According to the [documentation](https://docs.fire.ly/projects/Firely-NET-SDK/client/setup.html#creating-a-fhirclient) this is incorrect, as the FhirClient is not threadsafe. It is also discussed inside of this [Github thread](https://github.com/FirelyTeam/firely-net-sdk/issues/1110).

In this PR, a thread-safe FhirClient from the Microsoft Health OSS project has been leveraged to mitigate the above issue. 

Code Changes:
- leveraged Microsoft.Health.Fhir.Client FhirClient
- removed references to HL7 FhirClient
- imported Microsoft.Health.Fhir.R4.Client package
- updated projects to c# 8.0
- removed FhirClientFactory.cs
- added FhirServiceRepository.cs
- removed FhirOperationException
- updated unit tests
- added ServiceCollectionExtensions and HttpClientBuilderExtensions
- removed Common.R4.Ingest.UnitTests
- updated build pipeline


src/lib/Microsoft.Health.Extensions.Fhir.R4/HttpClientBuilderRegistrationExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+
+using System;
+using EnsureThat;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Health.Extensions.Host.Auth;
+using Microsoft.Health.Logging.Telemetry;
+
+namespace Microsoft.Health.Extensions.Fhir
+{
+    public static class HttpClientBuilderRegistrationExtensions
```
dustinburson, line 14: >HttpClientBuilderRegistrationExtensions [](http://example.com/codeflow?start=24&length=39)

Is there anything in this class that has a dependency on FHIR R4 libraries?  If not this should be in the Microsoft.Health.Extensions.Fhir project.  By placing it here we will need to duplicate this once R5 or other FHIR versions are released.  If there is a dependency that requires it be separate then this is fine.

kenziedolish: This class instantiates BearerTokenAuthorizationMessageHandler() which has dependencies on the FhirClientMetric class in Fhir R4 libraries. Because of these existing dependencies, I believe we should keep HttpClientBuilderRegistrationExtensions class separate

However if the above solution is not meeting our needs, a potential idea is to move Fhir.R4.Telemetry folder and HttpClientBuilderRegistrationExtensions class to Microsoft.Health.Extensions.Fhir project now that we've removed FhirOperationException.  Feel free to reopen this thread if you'd like to discuss this further

I moved the ServiceCollectionExtensions class to the Microsoft.Health.Extensions.Fhir project because it has no dependencies on FHIR R4 libraries

src/console/MeasurementCollectionToFhir/ProcessorStartup.cs
```diff
@@ -38,16 +41,31 @@ public void ConfigureServices(IServiceCollection services)
             services.Configure<ResourceIdentityOptions>(Configuration.GetSection("ResourceIdentity"));
             services.Configure<FhirClientFactoryOptions>(Configuration.GetSection("FhirClient"));
 
-            services.TryAddSingleton<IFactory<FhirClient>, FhirClientFactory>();
-            services.TryAddSingleton(sp => sp.GetRequiredService<IFactory<FhirClient>>().Create());
+            var url = new Uri(Environment.GetEnvironmentVariable("FhirService:Url"));
+            bool useManagedIdentity = Configuration.GetValue<bool>("FhirClient:UseManagedIdentity");
```
dustinburson, line 45: Suggestion:  It would be good to wrap these in an options class or incorporate in it's own extension for easier reuse.

src/lib/Microsoft.Health.Extensions.Fhir.R4/Repository/FhirServiceRepository.cs
```diff
@@ -0,0 +1,129 @@
...
+            CancellationToken token = cancellationToken.Equals(null) ? _cancellationTokenSource.Token : cancellationToken;
+
+            return await _fhirClient.UpdateAsync(resource, ifMatchVersion, provenanceHeader, token).ConfigureAwait(false);
+        }
+
+        public async IAsyncEnumerable<Bundle> IterateOverAdditionalBundlesAsync(
+            Bundle bundle,
+            [EnumeratorCancellation] CancellationToken cancellationToken = default)
+        {
+            if (bundle.Link.Count != 0)
```
dustinburson, line 116: >!= [](http://example.com/codeflow?start=34&length=2)

Unfortunately we can't short cut based on the count.  It is possible to get a continuation token with no entries but still have a next link. In that scenario we keep following the token till the link is empty.

src/lib/Microsoft.Health.Extensions.Fhir.R4/Repository/FhirServiceRepository.cs
```diff
@@ -0,0 +1,129 @@
...
+using System.Runtime.CompilerServices;
+using System.Threading;
+using System.Threading.Tasks;
+using EnsureThat;
+using Hl7.Fhir.Model;
+using IFhirClient = Microsoft.Health.Fhir.Client.IFhirClient;
+
+namespace Microsoft.Health.Extensions.Fhir.Repository
+{
+    public class FhirServiceRepository : IFhirServiceRepository
```
dustinburson, line 16: >Repository [](http://example.com/codeflow?start=28&length=10)

Not sure repository is the correct term.  Usually a repository is responsible for retrieving and persisting data from another service.  These are really extensions on top of the existing FHIR Client.  Suggestion to just do these as extension methods.

src/lib/Microsoft.Health.Extensions.Fhir.R4/Repository/FhirServiceRepository.cs
```diff
@@ -0,0 +1,129 @@
...
+    {
+        private readonly IFhirClient _fhirClient;
+
+        private readonly CancellationTokenSource _cancellationTokenSource;
```
dustinburson, line 20: >_cancellationTokenSource [](http://example.com/codeflow?start=49&length=24)

Is there are reason to have a static cts?  I don't think this will behave properly in multi threaded scenarios.  This really should be an instance for each method.

src/lib/Microsoft.Health.Extensions.Fhir.R4/Repository/FhirServiceRepository.cs
```diff
@@ -0,0 +1,129 @@
...
+            T resource,
+            string conditionalCreateCriteria = null,
+            string provenanceHeader = null,
+            CancellationToken cancellationToken = default)
+            where T : Resource
+        {
+            EnsureArg.IsNotNull(resource, nameof(resource));
+            _cancellationTokenSource.CancelAfter(60000);
+
+            CancellationToken token = cancellationToken.Equals(null) ? _cancellationTokenSource.Token : cancellationToken;
```
dustinburson, line 38: >cancellationToken.Equals(null) ? _cancellationTokenSource.Token : cancellationToken; [](http://example.com/codeflow?start=38&length=84)

You want create linked tokens.  You can chain they together so you can have the timeout or the original ct that was passed in.  Right now you only have one.

src/lib/Microsoft.Health.Extensions.Fhir.R4/Repository/FhirServiceRepository.cs
```diff
@@ -0,0 +1,129 @@
...
+
+        public async Task<T> CreateResourceAsync<T>(
+            T resource,
+            string conditionalCreateCriteria = null,
+            string provenanceHeader = null,
+            CancellationToken cancellationToken = default)
+            where T : Resource
+        {
+            EnsureArg.IsNotNull(resource, nameof(resource));
+            _cancellationTokenSource.CancelAfter(60000);
```
dustinburson, line 36: >60000 [](http://example.com/codeflow?start=49&length=5)

Would be good to make the timeout a constant.  Will make it easier to update and make it configurable in the future if we want change the timeout.

src/lib/Microsoft.Health.Fhir.Ingest/Service/CachedResourceIdentityService.cs
```diff
@@ -70,5 +70,7 @@ protected virtual void Dispose(bool disposing)
         protected abstract Task<IDictionary<ResourceType, string>> ResolveResourceIdentitiesInternalAsync(IMeasurementGroup input);
 
         protected abstract string GetCacheKey(IMeasurementGroup input);
+
+        public abstract ResourceIdentityServiceType GetResourceIdentityServiceType();
```
dustinburson, line 74: >GetResourceIdentityServiceType [](http://example.com/codeflow?start=52&length=30)

Can you add some details on how this is used and why it is needed?

dustinburson (changes requested):
:clock1:

src/lib/Microsoft.Health.Extensions.Fhir.R4/BundleExtensions.cs
```diff
@@ -96,30 +103,16 @@ public static async Task<IEnumerable<TResource>> ReadFromBundleWithContinuationA
                         count--;
                     }
                 }
+            };
+
+            storeResources.Invoke(bundle);
 
-                bundle = await fhirClient.ContinueAsync(bundle).ConfigureAwait(false);
+            await foreach (var currentBundle in fhirClient.IterateOverAdditionalBundlesAsync(bundle))
```
rogordon01, line 110: I think that we should be passing in non-default Cancellation tokens into the fhir client. I've seen cases where calls to the Fhir Server using the original client hung. By enforcing the use of non-default cancellation tokens we can help mitigate that.

src/lib/Microsoft.Health.Extensions.Fhir.R4/FhirServiceValidator.cs
```diff
@@ -4,22 +4,27 @@
...
+            string url,
+            ITelemetryLogger logger)
         {
             EnsureArg.IsNotNull(client, nameof(client));
+            EnsureArg.IsNotNullOrWhiteSpace(url, nameof(url));
 
             try
             {
-                client.CapabilityStatement(SummaryType.True);
+                await client.ReadResourceAsync<Hl7.Fhir.Model.CapabilityStatement>(url).ConfigureAwait(false);
```
rogordon01, line 27: Supply a cancellation token to the fhir client

src/lib/Microsoft.Health.Extensions.Fhir.R4/Repository/FhirServiceRepository.cs
rogordon01: With this PR, we would no longer pass around the FhirClient, so there would be no object to have extension methods against. The original idea was to provide a layer of abstraction between our code and how we communicate between the FhirServer, and to remove hard dependencies to the HL7 Fhir Client. Our code would now retrieve and persist data via this class and hide the fact that we use the FhirClient. 

Do you feel that this level of abstraction is not needed? I do see that the new FhirClient makes use of an interface `IFhirClient`. Should we make use of that everywhere and then make this class extension methods?

src/lib/Microsoft.Health.Fhir.R4.Ingest/Host/FhirHealthCheckExtensions.cs
```diff
@@ -33,8 +36,17 @@ internal static IWebJobsBuilder AddFhirHealthCheck(this IWebJobsBuilder builder)
             builder.Services.Configure<FhirClientFactoryOptions>(config.GetSection("FhirClient"));
 
             // Register services
-            builder.Services.TryAddSingleton<IFactory<FhirClient>, FhirClientFactory>();
-            builder.Services.TryAddSingleton(sp => sp.GetRequiredService<IFactory<FhirClient>>().Create());
+            var url = new Uri(Environment.GetEnvironmentVariable("FhirService:Url"));
```
rogordon01, line 39: You can use the Configuration object versus calling Environment.GetEnviornmentVariable. This allows you to read configuration values from any/all registered providers (env, from a config file, etc)

src/lib/Microsoft.Health.Extensions.Fhir.R4/Telemetry/Exceptions/FhirServiceExceptionProcessor.cs
```diff
@@ -51,11 +50,13 @@ public static (Exception customException, string errorName) CustomizeException(E
...
                             return (new UnauthorizedAccessFhirServiceException(message, exception, helpLink, errorName), errorName);
                         case HttpStatusCode.NotFound:
-                            message = FhirResources.FhirServiceNotFound;
+                            message = DetermineNotFoundErrorMessage(exception);
                             errorName = nameof(FhirServiceErrorCode.ConfigurationError);
                             return (new InvalidFhirServiceException(message, exception, errorName), errorName);
-                        default:
+                        case HttpStatusCode.BadRequest:
                             return (exception, status.ToString());
+                        default:
```
rogordon01, line 58: Previously, it looks like if the status code didn't match any `case` statement, we'd return the status code as the error name. Now we always return `FhirServiceErrorCode.HttpRequestError`. Can you ask @kyclai or @pallar-ms  if that is desired? I know that they did a lot of work with the exceptions and I don't want to mishandle telemetry.

rogordon01 (changes requested):

src/console/MeasurementCollectionToFhir/ProcessorStartup.cs
kenziedolish: Incorporated into ServiceCollectionExtensions.cs

src/lib/Microsoft.Health.Extensions.Fhir.R4/Repository/FhirServiceRepository.cs
kenziedolish: Updated logic to follow token until token.NextLink equals null

src/lib/Microsoft.Health.Extensions.Fhir.R4/Repository/FhirServiceRepository.cs
kenziedolish: Made timeout a constant

src/lib/Microsoft.Health.Extensions.Fhir.R4/Repository/FhirServiceRepository.cs
kenziedolish: Updated to create an instance of new CancellationTokenSource() in each method

src/lib/Microsoft.Health.Extensions.Fhir.R4/Repository/FhirServiceRepository.cs
kenziedolish: Thanks for suggesting this. Did some further reading and chained the internal and external tokens for each method together

src/lib/Microsoft.Health.Extensions.Fhir.R4/Repository/FhirServiceRepository.cs
kenziedolish: Agreed that repository may not be the best term. I think a central point of interaction with the FhirClient could provide further extensibility and organization though. An example off the top of my head is if we want to refactor the FhirClient again in the future, we'd only have to update the Repository class

Just my perspective. Looking forward to hearing both your thoughts

src/lib/Microsoft.Health.Extensions.Fhir.R4/BearerTokenAuthorizationMessageHandler.cs
```diff
@@ -0,0 +1,54 @@
...
+using Microsoft.Health.Logging.Telemetry;
+
+namespace Microsoft.Health.Extensions.Fhir
+{
+    public class BearerTokenAuthorizationMessageHandler : DelegatingHandler
+    {
+        public BearerTokenAuthorizationMessageHandler(Uri uri, TokenCredential tokenCredentialProvider, ITelemetryLogger logger)
+        {
+            TokenCredential = EnsureArg.IsNotNull(tokenCredentialProvider, nameof(tokenCredentialProvider));
+            Uri = EnsureArg.IsNotNull(uri);
```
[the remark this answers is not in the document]
kenziedolish: Updated to use EnsureArg on Uri and Logger

src/lib/Microsoft.Health.Extensions.Fhir.R4/BearerTokenAuthorizationMessageHandler.cs
```diff
@@ -0,0 +1,54 @@
...
+            Scopes = new string[] { Uri + ".default" };
```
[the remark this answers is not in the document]
kenziedolish: No, in my testing the string concatenation was not adding a forward slash to the URI if it wasn't provided. I've updated the code to append a forward slash when the URI doesn't have one. 

I chose to go with the .EndsWith() method for strings instead of System.TryCreateUri because Scopes is a string array

src/lib/Microsoft.Health.Extensions.Fhir.R4/BearerTokenAuthorizationMessageHandler.cs
```diff
@@ -0,0 +1,54 @@
...
+        private TokenCredential TokenCredential { get; }
+
+        private Uri Uri { get; }
+
+        private string[] Scopes { get; }
+
+        protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
+        {
+            var requestContext = new TokenRequestContext(Scopes);
+            var accessToken = await TokenCredential.GetTokenAsync(requestContext, CancellationToken.None);
```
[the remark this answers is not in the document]
kenziedolish: Updated to pass in cancellationToken

src/lib/Microsoft.Health.Fhir.R4.Ingest/Host/FhirHealthCheckExtensions.cs
kenziedolish: Updated to use the configuration object. Also moved code out to extension method

src/lib/Microsoft.Health.Extensions.Fhir.R4/Telemetry/Exceptions/FhirServiceExceptionProcessor.cs
kenziedolish: Absolutely, I connected with Carrie offline yesterday about the changes made to the FhirServiceExceptionProcessor and FhirServiceValidator. We're coordinating our changes in one PR

@kyclai or @pallar-ms I'll wait for your feedback on the changes made in the FhirServiceExceptionProcessor before resolving

src/lib/Microsoft.Health.Extensions.Fhir.R4/Telemetry/Exceptions/FhirServiceExceptionProcessor.cs
```diff
@@ -51,11 +50,13 @@ public static (Exception customException, string errorName) CustomizeException(E
                             errorName = nameof(FhirServiceErrorCode.AuthorizationError);
                             return (new UnauthorizedAccessFhirServiceException(message, exception, helpLink, errorName), errorName);
                         case HttpStatusCode.NotFound:
-                            message = FhirResources.FhirServiceNotFound;
+                            message = DetermineNotFoundErrorMessage(exception);
                             errorName = nameof(FhirServiceErrorCode.ConfigurationError);
                             return (new InvalidFhirServiceException(message, exception, errorName), errorName);
-                        default:
+                        case HttpStatusCode.BadRequest:
```
kyclai, line 56: What scenario gives a `BadRequest`? I'm wondering if it needs a custom exception like `InvalidFhirServiceException` to indicate it's a user error, but based on these changes I'm thinking it's more likely that this case can be removed and handled as part of the `default` case. (Please see my next comment.)

src/lib/Microsoft.Health.Extensions.Fhir.R4/Telemetry/Exceptions/FhirServiceExceptionProcessor.cs
kyclai: Thanks for bringing this up; please see my other comments.

src/lib/Microsoft.Health.Extensions.Fhir.R4/Telemetry/Exceptions/FhirServiceExceptionProcessor.cs
```diff
@@ -74,17 +75,6 @@ public static (Exception customException, string errorName) CustomizeException(E
                     errorName = nameof(FhirServiceErrorCode.ConfigurationError);
                     return (new InvalidFhirServiceException(message, exception, errorName), errorName);
 
-                case HttpRequestException _:
-                    // TODO: In .NET 5 and later, check HttpRequestException's StatusCode property instead of the Message property
-                    if (exception.Message.Contains(FhirResources.HttpRequestErrorNotKnown, StringComparison.CurrentCultureIgnoreCase))
```
kyclai, line 79: In my testing with the previous FhirClient, the `HttpRequestException` given when the FHIR service in the FHIR service URL does not exist, e.g. _https://[non-existent FHIR service name].azurehealthcareapis.com_, has a null `StatusCode`. (This is why I still checked the `Message` in the closed PR #168.)

So if this scenario with a non-existent FHIR service still produces a null `StatusCode` with the new FhirClient, then
- checking the `Message` shouldn't be handled as part of the `NotFound` `StatusCode` (via `DetermineNotFoundErrorMessage`), and
- the `NotFound` case should continue producing the exception `message` of `FhirResources.FhirServiceNotFound`.

test/Microsoft.Health.Extensions.Fhir.R4.UnitTests/FhirServiceExceptionProcessorTests.cs
```diff
@@ -19,14 +18,14 @@ namespace Microsoft.Health.Extensions.Fhir.R4.UnitTests
...
+        private static readonly Exception _httpRequestForbiddenEx = new HttpRequestException("test", inner: null, HttpStatusCode.Forbidden);
+        private static readonly Exception _httpRequestNotFoundEx = new HttpRequestException("test", inner: null, HttpStatusCode.NotFound);
+        private static readonly Exception _httpRequestBadRequestEx = new HttpRequestException("test", inner: null, HttpStatusCode.BadRequest);
         private static readonly Exception _argEndpointNullEx = new ArgumentNullException("endpoint");
         private static readonly Exception _argEndpointEx = new ArgumentException("endpoint", "Endpoint must be absolute");
         private static readonly Exception _argEx = new ArgumentException("test_message", "test_param");
         private static readonly Exception _uriEx = new UriFormatException();
-        private static readonly Exception _httpNotKnownEx = new HttpRequestException("Name or service not known");
+        private static readonly Exception _httpNotKnownEx = new HttpRequestException("Name or service not known", inner: null, HttpStatusCode.NotFound);
```
kyclai, line 28: Please refer to my previous comment about the `HttpRequestException` with the `Message` of "Name or service not known" not having a `StatusCode`. Not sure if this is still true with the new FhirClient.

kyclai (changes requested):

src/lib/Microsoft.Health.Extensions.Fhir.R4/Telemetry/Exceptions/FhirServiceExceptionProcessor.cs
```diff
@@ -41,8 +40,8 @@ public static (Exception customException, string errorName) CustomizeException(E
 
             switch (exception)
             {
-                case FhirOperationException _:
-                    var status = ((FhirOperationException)exception).Status;
+                case HttpRequestException _:
```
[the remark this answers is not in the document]
kyclai: TIL, thanks!

src/lib/Microsoft.Health.Extensions.Fhir.R4/BearerTokenAuthorizationMessageHandler.cs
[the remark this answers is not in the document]
kenziedolish: Thanks for calling this out. We discussed this offline and I've created a [WI](https://microsofthealth.visualstudio.com/Health/_workitems/edit/88540) that can be taken up in a subsequent PR to cache the results of AquireServiceTokenAsync(). This will enable us to leverage the token for its full lifetime rather than acquire new tokens for each request

src/lib/Microsoft.Health.Extensions.Fhir.R4/BundleExtensions.cs
kenziedolish: Thanks for calling this out. We discussed this offline, and I've created a [WI](https://microsofthealth.visualstudio.com/Health/_workitems/edit/88541) that can be taken up in a subsequent PR to refactor upstream methods to pass non-default tokens. Currently, linking cancellation tokens in the FhirServiceRepository will prevent the FhirService from hanging

src/lib/Microsoft.Health.Extensions.Fhir.R4/BundleExtensions.cs
```diff
@@ -41,7 +41,10 @@ public static IEnumerable<TResource> ReadFromBundle<TResource>(this Bundle bundl
             }
         }
 
-        public static async Task<TResource> ReadOneFromBundleWithContinuationAsync<TResource>(this Bundle bundle, FhirClient fhirClient, bool throwOnMultipleFound = true)
+        public static async Task<TResource> ReadOneFromBundleWithContinuationAsync<TResource>(
+            this Bundle bundle,
+            IFhirServiceRepository fhirClient,
+            bool throwOnMultipleFound = true)
```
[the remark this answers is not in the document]
kenziedolish: Thanks for calling this out. We discussed this offline, and I've created a [WI](https://microsofthealth.visualstudio.com/Health/_workitems/edit/88541) that can be taken up in a subsequent PR to refactor upstream methods to pass non-default tokens

src/lib/Microsoft.Health.Extensions.Fhir.R4/FhirServiceValidator.cs
kenziedolish: Thanks for calling this out. We discussed this offline, and I've created a [WI](https://microsofthealth.visualstudio.com/Health/_workitems/edit/88541) that can be taken up in a subsequent PR to refactor upstream methods to pass non-default tokens

src/lib/Microsoft.Health.Extensions.Fhir.R4/Repository/FhirServiceRepository.cs
kenziedolish: Sounds good. Thanks Rob and Dustin for your feedback on this. I'm resolving this conversation thread but feel free to reopen if there any further questions/comments

src/lib/Microsoft.Health.Fhir.Ingest/Service/CachedResourceIdentityService.cs
kenziedolish: Thanks for your feedback on this. My intention was to improve the code by injecting the IdentityServices in ProcessorStartup. However I agree that with these changes, we're creating all possible services (Create, Lookup, and LookupWithEncounter) and then selecting a match. I've gone ahead and reverted the changes here

src/lib/Microsoft.Health.Extensions.Fhir.R4/Telemetry/Exceptions/FhirServiceExceptionProcessor.cs
kenziedolish: Taking a look on comments. Closing this thread

src/lib/Microsoft.Health.Fhir.Ingest/Service/ResourceIdentityServiceFactory.cs
```diff
@@ -9,83 +9,33 @@
 using EnsureThat;
 using Microsoft.Health.Common;
 using Microsoft.Health.Fhir.Ingest.Config;
-using Microsoft.Health.Fhir.Ingest.Host;
 
 namespace Microsoft.Health.Fhir.Ingest.Service
 {
-    public class ResourceIdentityServiceFactory : IFactory<IResourceIdentityService, ResourceIdentityOptions>
```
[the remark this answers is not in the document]
kenziedolish: Reverted these changes

src/lib/Microsoft.Health.Extensions.Fhir.R4/Repository/FhirServiceRepository.cs
```diff
@@ -0,0 +1,139 @@
...
+using System.Threading.Tasks;
+using EnsureThat;
+using Hl7.Fhir.Model;
+using IFhirClient = Microsoft.Health.Fhir.Client.IFhirClient;
+
+namespace Microsoft.Health.Extensions.Fhir.Repository
+{
+    public class FhirServiceRepository : IFhirServiceRepository
+    {
+        private const double TimeOut = 60;
```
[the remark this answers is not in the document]
kenziedolish: Thanks for pointing this out! I updated AddFhirClient() in the ServiceCollectionExtensions class to set a timeout on the HttpClient that's passed to the FhirClient.

src/lib/Microsoft.Health.Fhir.Ingest/Microsoft.Health.Fhir.Ingest.csproj
```diff
@@ -5,7 +5,7 @@
     <HighEntropyVA>true</HighEntropyVA>
     <AssemblyName>Microsoft.Health.Fhir.Ingest</AssemblyName>
     <RootNamespace>Microsoft.Health.Fhir.Ingest</RootNamespace>
-    <langversion>7.3</langversion>
+    <langversion>8.0</langversion>
```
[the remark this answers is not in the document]
kenziedolish: Bumped all 3 projects to 10.0

src/lib/Microsoft.Health.Extensions.Fhir.R4/Telemetry/Exceptions/FhirServiceExceptionProcessor.cs
kenziedolish: Thanks for reviewing Carrie! 

Yesterday, I ran through the test cases in the document you sent over and found that the FhirClient is throwing a FhirException which also inherits from the base class System.Exception - similar to the FhirOperationException we recognized previously. We can leverage FhirException in place of FhirOperationException and maintain all of the current logic in the FhirServiceExceptionProcessor. I also incorporated the changes from your PR for the HttpRequestException cases and updated the FhirServiceExceptionProcessor tests with your changes.

Please look on the latest changes and provide your feedback

test/Microsoft.Health.Extensions.Fhir.R4.UnitTests/FhirServiceExceptionProcessorTests.cs
kenziedolish: Pushed new changes to FhirServiceExceptionProcessorTests. Please refer to comment above

src/lib/Microsoft.Health.Extensions.Fhir.R4/Telemetry/Exceptions/FhirServiceExceptionProcessor.cs
kenziedolish: Thanks for pointing this out. I pushed new changes to the FhirServiceExceptionProcessor. Please see my comment below

src/lib/Microsoft.Health.Extensions.Fhir.R4/Telemetry/Exceptions/FhirServiceExceptionProcessor.cs
```diff
@@ -51,11 +50,13 @@ public static (Exception customException, string errorName) CustomizeException(E
...
                         case HttpStatusCode.NotFound:
-                            message = FhirResources.FhirServiceNotFound;
+                            message = DetermineNotFoundErrorMessage(exception);
                             errorName = nameof(FhirServiceErrorCode.ConfigurationError);
                             return (new InvalidFhirServiceException(message, exception, errorName), errorName);
-                        default:
+                        case HttpStatusCode.BadRequest:
                             return (exception, status.ToString());
+                        default:
+                            return (exception, nameof(FhirServiceErrorCode.HttpRequestError));
```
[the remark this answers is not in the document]
kenziedolish: Thanks for calling this out. I updated the FhirServiceExceptionProcessor. Please see my comment below

src/lib/Microsoft.Health.Extensions.Fhir.R4/Telemetry/Exceptions/FhirServiceExceptionProcessor.cs
```diff
@@ -101,5 +91,17 @@ public static (Exception customException, string errorName) CustomizeException(E
                     return (exception, nameof(FhirServiceErrorCode.GeneralError));
             }
         }
+
+        private static string DetermineNotFoundErrorMessage(Exception exception)
```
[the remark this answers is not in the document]
kenziedolish: Latest changes removed this method

src/lib/Microsoft.Health.Extensions.Fhir.R4/Telemetry/Exceptions/FhirServiceExceptionProcessor.cs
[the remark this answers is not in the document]
kenziedolish: Thanks for calling this out! Please refer to my [comment](https://github.com/microsoft/iomt-fhir/pull/167#discussion_r803212593) below providing context on the latest changes for this class.

src/lib/Microsoft.Health.Extensions.Fhir.R4/Telemetry/Exceptions/FhirServiceExceptionProcessor.cs
kyclai: The changes to FhirServiceExceptionProcessor and corresponding unit tests look good. Thanks for testing!

src/lib/Microsoft.Health.Extensions.Fhir.R4/Service/FhirService.cs
```diff
@@ -0,0 +1,97 @@
...
+
+            return await _fhirClient.UpdateAsync(resource, ifMatchVersion, provenanceHeader, cancellationToken).ConfigureAwait(false);
+        }
+
+        public async IAsyncEnumerable<Bundle> IterateOverAdditionalBundlesAsync(
+            Bundle bundle,
+            [EnumeratorCancellation] CancellationToken cancellationToken = default)
+        {
+            Bundle nextBundle = bundle;
+            while (nextBundle.NextLink != null)
```
rogordon01, line 87: You need to check for `null` here.  The caller may pass in a `null` bundle and `fhirClient.SearchAsync` may return a `null` bundle.

```c#
while (nextBundle?.NextLink != null) {
...
}
```

src/lib/Microsoft.Health.Extensions.Fhir.R4/FhirClientValidator.cs
```diff
@@ -4,22 +4,28 @@
...
+            this HttpClient client,
+            ITelemetryLogger logger)
         {
             EnsureArg.IsNotNull(client, nameof(client));
+            EnsureArg.IsNotNull(logger, nameof(logger));
 
             try
             {
-                client.CapabilityStatement(SummaryType.True);
+                var fhirClient = new FhirClient(client);
```
rogordon01, line 27: We are building a new FhirClient here, and we don't really need to. You can adjust the original method signature and make this an extension method for `IFhirClient`. And invoke it as follows on the Singleton FhirClient that we have available via DI:

```c#
var serviceProvider = serviceCollection.BuildServiceProvider();
var fhirClient = serviceProvider.GetRequiredService<IFhirClient>();
var logger = serviceProvider.GetRequiredService<ITelemetryLogger>();

await fhirClient.ValidateFhirClientAsync(logger);
```

We just need a place to execute the above. Perhaps at the end of `ProcessorStartup#ConfigureServices` method?

rogordon01 (changes requested):

src/lib/Microsoft.Health.Extensions.Fhir.R4/BearerTokenAuthorizationMessageHandler.cs
```diff
@@ -0,0 +1,54 @@
...
+        private TokenCredential TokenCredential { get; }
+
+        private Uri Uri { get; }
+
+        private string[] Scopes { get; }
+
+        protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
+        {
+            var requestContext = new TokenRequestContext(Scopes);
+            var accessToken = await TokenCredential.GetTokenAsync(requestContext, cancellationToken);
```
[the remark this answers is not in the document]
kenziedolish: Yes this is the [WI](https://microsofthealth.visualstudio.com/Health/_workitems/edit/88540)

src/lib/Microsoft.Health.Extensions.Fhir.R4/FhirClientValidator.cs
kenziedolish: Thanks for reviewing Rob. This definitely works for the console project. However ProcessorStartup isn't called when running the functions (Microsoft.Health.Fhir.Ingest.Host project). If we're to place fhirClient.ValidateFhirClientAsync() at the top of both the Normalization and FhirTransformation functions, the validation method would run each time they're triggered.

We could consider calling await fhirClient.ValidateFhirClientAsync(logger) at the end of AddMeasurementFhirImport() and in ProcessorStartup.ConfigureServices. My thought is that it'd best to call the validation method when we configure the client though, so we don't have to call it in multiple locations.

Happy to discuss further

kenziedolish: Synced offline and pushed new changes

src/lib/Microsoft.Health.Extensions.Fhir.R4/Service/FhirService.cs
kenziedolish: Good callout. Added the null check

rogordon01 (approved):

src/lib/Microsoft.Health.Extensions.Fhir.R4/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,59 @@
...
+            var logger = serviceProvider.GetRequiredService<ITelemetryLogger>();
+
+            serviceCollection.AddHttpClient<IFhirClient, FhirClient>(client =>
+            {
+                client.BaseAddress = url;
+                client.Timeout = TimeSpan.FromSeconds(60);
+
+                // Using discard because we don't need result
+                var fhirClient = new FhirClient(client);
+                _ = fhirClient.ValidateFhirClientAsync(logger);
```
dustinburson, line 39: >_ = fhirClient.ValidateFhirClientAsync(logger); [](http://example.com/codeflow?start=16&length=47)

If we aren't waiting, is there value to doing this call when the class is created?

src/lib/Microsoft.Health.Extensions.Fhir.R4/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,59 @@
...
+            var url = new Uri(configuration.GetValue<string>("FhirService:Url"));
+            bool useManagedIdentity = configuration.GetValue<bool>("FhirClient:UseManagedIdentity");
+
+            serviceCollection.AddSingleton(typeof(ITelemetryLogger), typeof(IomtTelemetryLogger));
+            var serviceProvider = serviceCollection.BuildServiceProvider();
+            var logger = serviceProvider.GetRequiredService<ITelemetryLogger>();
+
+            serviceCollection.AddHttpClient<IFhirClient, FhirClient>(client =>
+            {
+                client.BaseAddress = url;
```
dustinburson, line 34: >url [](http://example.com/codeflow?start=37&length=3)

More of a FYI, but since the url and logger are variables from an outside scope this means the entire contents of this method outside this lamda will be implicitly captured in a class in the compiled code.  

For more details and understanding the implications see here [https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/lambda-expressions#capture-of-outer-variables-and-variable-scope-in-lambda-expressions](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/lambda-expressions#capture-of-outer-variables-and-variable-scope-in-lambda-expressions)

src/lib/Microsoft.Health.Extensions.Fhir.R4/Telemetry/Exceptions/FhirServiceExceptionProcessor.cs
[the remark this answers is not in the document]
dustinburson: Isn't clear to me how comment address the cast when it isn't needed.  

Instead of:
case FhirException _:
                    var status = ((FhirException)exception).StatusCode;

you can just do
case FhirException fhirEx:
                    var status = fhirEx.StatusCode;

src/lib/Microsoft.Health.Extensions.Fhir.R4/Telemetry/Exceptions/FhirServiceExceptionProcessor.cs
```diff
@@ -41,8 +41,8 @@ public static (Exception customException, string errorName) CustomizeException(E
 
             switch (exception)
```
kyclai, line 42: Nit: I think the case for `ArgumentException` isn't needed anymore, since 1) it was used to capture the `ArgumentNullException` and `ArgumentException` that can be thrown in the previous SDK's `FhirClient` constructor and 2) now we're not wrapping a `try`-`catch` around creation of the `FhirClient` to process/log any exceptions.

Correspondingly, the following can probably be removed:
- `ArgumentNullException` and `ArgumentException` test cases in `FhirServiceExceptionProcessorTests`
- `FhirServiceEndpointInvalid` in `FhirResources`

src/lib/Microsoft.Health.Extensions.Fhir.R4/Service/FhirService.cs
```diff
@@ -0,0 +1,97 @@
...
+    public class FhirService : IFhirService
+    {
+        private readonly IFhirClient _fhirClient;
+
+        public FhirService(IFhirClient fhirClient)
+        {
+            _fhirClient = EnsureArg.IsNotNull(fhirClient, nameof(fhirClient));
+        }
+
+        public async Task<T> CreateResourceAsync<T>(
```
dustinburson, line 25: It seems like most of these methods are just pass throughs to the same method on the IFhirClient.  Curious as to the overall purpose of the class rather than using the IFhirClient directly?

src/lib/Microsoft.Health.Extensions.Fhir.R4/IdentifierExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace Microsoft.Health.Extensions.Fhir
+{
+    public static class IdentifierExtensions
+    {
+        public static string ToSearchQueryParameter(this Identifier identifier)
+        {
+            EnsureArg.IsNotNull(identifier, nameof(identifier));
+
+            return $"identifier={identifier.System}|{identifier.Value}";
```
dustinburson, line 17: >identifier.System}|{identifier.Value} [](http://example.com/codeflow?start=33&length=37)

I think it would be better to place this in the existing SearchExtensions class.  Also should use the already ToSearchToken.  The will properly prepare the search term for the identifier where the system| should be dropped if the system isn't specified on the identifier, i.e. 
        public static string ToSearchToken(this Hl7.Fhir.Model.Identifier identifier)

src/lib/Microsoft.Health.Extensions.Fhir.R4/FhirClientValidator.cs
```diff
@@ -4,22 +4,26 @@
...
+            this IFhirClient client,
+            ITelemetryLogger logger)
         {
             EnsureArg.IsNotNull(client, nameof(client));
+            EnsureArg.IsNotNull(logger, nameof(logger));
 
             try
             {
-                client.CapabilityStatement(SummaryType.True);
+                await client.ReadAsync<Hl7.Fhir.Model.CapabilityStatement>("metadata").ConfigureAwait(false);
```
dustinburson, line 26: >CapabilityStatement [](http://example.com/codeflow?start=54&length=19)

Does this preserve the summary functionality from the old call?  I don't think so.  Here is the source from the FirelyClient. Would be good if we could preserve [https://github.com/FirelyTeam/firely-net-sdk/blob/471a5a1417355e4bbcfbff722b62d624c0f946e1/src/Hl7.Fhir.Core/Rest/TransactionBuilder.cs#L209](https://github.com/FirelyTeam/firely-net-sdk/blob/471a5a1417355e4bbcfbff722b62d624c0f946e1/src/Hl7.Fhir.Core/Rest/TransactionBuilder.cs#L209)

Capability statements can be quite large and we are just doing this call to verify we can connect to the FHIR service (metadata end point doesn't need auth) so a summary is fine.

src/lib/Microsoft.Health.Extensions.Fhir.R4/BearerTokenAuthorizationMessageHandler.cs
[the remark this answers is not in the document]
dustinburson: If we are requesting a token with every request that is significant override head.  Can we run the client through the perf testing framework?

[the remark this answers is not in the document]
dustinburson: Nevermind this should be fine, the tokens are cached in the implementation of DefaultAzureCredential

src/lib/Microsoft.Health.Extensions.Fhir.R4/BearerTokenAuthorizationMessageHandler.cs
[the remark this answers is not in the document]
dustinburson: I agree with Rob, using Uri.TryCreate should be cleaner and safer.

Example 
var baseUri = new Uri("[https://test.com](https://test.com)");
Uri.TryCreate(baseUri, ".default", out Uri newUri);

or in this code it would be 
Uri.TryCreate(Uri, ".default", out Uri scopeUri);
Scopes = new string [] { scopeUri.ToString() };

dustinburson (commented):
Added a few more comments, most aren't required changes but still interested in getting your thoughts on the new comments.

rogordon01:
Are we sure that caching is being done in DefaultAzureCredential? I had taken a look at its source code<https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/identity/Azure.Identity/src/DefaultAzureCredential.cs#L99> and didn’t notice any caching. Also the docs for the method signature seemed to indicate that clients were responsible for caching the returned token:

“This method is called automatically by Azure SDK client libraries. You may call this method directly, but you must also handle token caching and token refreshing. “

From: Dustin Burson ***@***.***>
Date: Thursday, February 10, 2022 at 5:03 PM
To: microsoft/iomt-fhir ***@***.***>
Cc: Rob Gordon ***@***.***>, Review requested ***@***.***>
Subject: Re: [microsoft/iomt-fhir] Updated FhirClient (PR #167)

@dustinburson commented on this pull request.

In src/lib/Microsoft.Health.Extensions.Fhir.R4/BearerTokenAuthorizationMessageHandler.cs<https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fmicrosoft%2Fiomt-fhir%2Fpull%2F167%23discussion_r804275862&data=04%7C01%7CRob.Gordon%40microsoft.com%7Cb7d355a3bcfc4377602108d9ecfa4e67%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C637801382078738534%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000&sdata=ytQ56FIhGZBga4ScnmkfJs9jCAmlJg3VWLTI922eLes%3D&reserved=0>:

> +            Scopes = new string[] { Uri + ".default" };

+            Logger = logger;

+        }

+

+        private ITelemetryLogger Logger { get; }

+

+        private TokenCredential TokenCredential { get; }

+

+        private Uri Uri { get; }

+

+        private string[] Scopes { get; }

+

+        protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)

+        {

+            var requestContext = new TokenRequestContext(Scopes);

+            var accessToken = await TokenCredential.GetTokenAsync(requestContext, CancellationToken.None);

Nevermind this should be fine, the tokens are cached in the implementation of DefaultAzureCredential

—
Reply to this email directly, view it on GitHub<https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fmicrosoft%2Fiomt-fhir%2Fpull%2F167%23discussion_r804275862&data=04%7C01%7CRob.Gordon%40microsoft.com%7Cb7d355a3bcfc4377602108d9ecfa4e67%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C637801382078738534%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000&sdata=ytQ56FIhGZBga4ScnmkfJs9jCAmlJg3VWLTI922eLes%3D&reserved=0>, or unsubscribe<https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fnotifications%2Funsubscribe-auth%2FARAAJ2XPHA3ENP54D4HYIIDU2RN5FANCNFSM5NYHPWDA&data=04%7C01%7CRob.Gordon%40microsoft.com%7Cb7d355a3bcfc4377602108d9ecfa4e67%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C637801382078738534%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000&sdata=mem%2BH70wFlJzSITk27odNZNqGY1bH4eOUPut4xUA%2FeI%3D&reserved=0>.
Triage notifications on the go with GitHub Mobile for iOS<https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Fapps.apple.com%2Fapp%2Fapple-store%2Fid1477376905%3Fct%3Dnotification-email%26mt%3D8%26pt%3D524675&data=04%7C01%7CRob.Gordon%40microsoft.com%7Cb7d355a3bcfc4377602108d9ecfa4e67%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C637801382078788515%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000&sdata=DfP8FplCpsO7MVSLIIp1IN5Keq%2FmcWB%2FX3ahfMnJgyg%3D&reserved=0> or Android<https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Fplay.google.com%2Fstore%2Fapps%2Fdetails%3Fid%3Dcom.github.android%26referrer%3Dutm_campaign%253Dnotification-email%2526utm_medium%253Demail%2526utm_source%253Dgithub&data=04%7C01%7CRob.Gordon%40microsoft.com%7Cb7d355a3bcfc4377602108d9ecfa4e67%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C637801382078788515%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000&sdata=2tMUgtj7hIae21%2Fp75lUiSwLzwDx3lKBxcFSgyzFH%2Bs%3D&reserved=0>.

dustinburson:
> @dustinburson Are we sure that caching is being done in DefaultAzureCredential? I had taken a look at its source code<https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/identity/Azure.Identity/src/DefaultAzureCredential.cs#L99> and didn’t notice any caching. Also the docs for the method signature seemed to indicate that clients were responsible for caching the returned token: “This method is called automatically by Azure SDK client libraries. You may call this method directly, but you must also handle token caching and token refreshing. “

Pretty sure but we can follow up to verify.  The key thing for this PR though is the underlying mechanism hasn't changed so there shouldn't be a regression.

src/lib/Microsoft.Health.Extensions.Fhir.R4/Service/FhirService.cs
kenziedolish: Sure, in a previous comment above we discussed this added layer of abstraction. I refactored the classname to be FhirService instead of FhirServiceRepository since the term "repository" wasn't the best fit. I believe this layer of abstraction provides advantages for us. The first is that if we want to update the FhirClient again in the future for example, the blast radius will be significantly smaller because we only need to make changes to this class. In addition, this removes hard dependencies on the underlying FhirClient.

Just my perspective. Do you think this layer of abstraction is unnecessary? We can pass around the IFhirClient and change this class to extension methods as well

src/lib/Microsoft.Health.Extensions.Fhir.R4/IdentifierExtensions.cs
kenziedolish: Thanks for calling this out. I've removed the IdentifierExtensions class and leveraged ToSearchToken() instead

src/lib/Microsoft.Health.Extensions.Fhir.R4/Telemetry/Exceptions/FhirServiceExceptionProcessor.cs
kenziedolish: Thanks for taking the time to review. In my testing, a System.ArgumentException is thrown from the try/catch block in FhirClientValidator.ValidateFhirClientAsync() if the user does not provide a necessary parameter (such as the Fhir Service authority for example). However, this may not necessarily be a production case since the user only provides the Fhir Service url

I think we should go ahead and add this as an item to the engineering backlog to further explore and take up in a subsequent PR. For this PR, I think it'd be best to keep our current exception handling logic as not introduce any regressions 

Happy to discuss further

test/Microsoft.Health.Extensions.Fhir.R4.UnitTests/FhirClientValidatorTests.cs
```diff
@@ -0,0 +1,41 @@
...
+            await ValidateFhirClientUrl(url, false);
+        }
+
+        private async Task ValidateFhirClientUrl(string url, bool expectedIsValid)
+        {
+            var fhirClient = Substitute.For<IFhirClient>();
+
+            string uri = url + "/metadata";
+
+            fhirClient.ReadAsync<Hl7.Fhir.Model.CapabilityStatement>(uri).ThrowsForAnyArgs(new Exception());
```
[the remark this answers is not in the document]
kenziedolish: Thanks for taking the time to review Carrie. You are correct that a valid test case for the ValidateFhirClientUrl helper function would fail with these changes since the code is throwing an exception for any string parameter. We could update the mocked ReadAsync to return different results based on the test case urls. However, I've reverted my changes here so that we can add future tests cases without mocking out each Read call based on the URL

In regard to the naming, the ValidateFhirClientUrl helper function tests the Microsoft.Health.Fhir.R4.Client FhirClient rather than the FhirService class. We are similarly validating the Fhir Client in ValidateFhirClientAsync(). I think this naming helps to differentiate between FhirService.cs class and the FhirClient 

Happy to discuss further

src/lib/Microsoft.Health.Extensions.Fhir.R4/FhirClientValidator.cs
kenziedolish: Thanks for calling this out. I've gone ahead and updated the GET call to return all supported elements that are marked as "summary" in the base definition of the resource

src/lib/Microsoft.Health.Extensions.Fhir.R4/ServiceCollectionExtensions.cs
kenziedolish: From my perspective, the value in doing this call when the class is created is that the ValidateFhirClientAsync() executes right after the services are configured and built. This surfaces any invalid FhirClient configurations to the user before ingesting events from the input event hub

In addition by calling the validation method here, we don't need to call it across various classes for the console project and function apps

src/lib/Microsoft.Health.Extensions.Fhir.R4/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,59 @@
...
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection.Extensions;
+using Microsoft.Health.Extensions.Host.Auth;
+using Microsoft.Health.Logging.Telemetry;
+using FhirClient = Microsoft.Health.Fhir.Client.FhirClient;
+using IFhirClient = Microsoft.Health.Fhir.Client.IFhirClient;
+
+namespace Microsoft.Health.Extensions.Fhir
+{
+    public static class ServiceCollectionExtensions
```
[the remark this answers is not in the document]
kenziedolish: Good point, these methods are being used for doing DI on the FhirClient. I think if we want to go ahead and refactor the FhirService.cs class to extension methods then we could refactor these methods into that class. Not sure if it's not necessary otherwise

Happy to discuss further

src/lib/Microsoft.Health.Extensions.Fhir.R4/Telemetry/Exceptions/FhirServiceExceptionProcessor.cs
kyclai: I hadn't come across an ArgumentException for a missing FHIR Service authority before, so it's good that you uncovered it in your testing and the exception processor caught it.

I agree that we don't need to update which exceptions are processed by the exception processor in this PR. As you suggested I'll create a work item to check which exceptions are processed given the updated FhirClient and to also update our internal doc about the original exceptions that are processed (e.g. FhirOperationException -> FhirException).

src/lib/Microsoft.Health.Extensions.Fhir.R4/HttpClientBuilderRegistrationExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+        {
+            EnsureArg.IsNotNull(httpClientBuilder, nameof(httpClientBuilder));
+            EnsureArg.IsNotNull(services, nameof(services));
+            EnsureArg.IsNotNull(uri, nameof(uri));
+
+            if (useManagedIdentity)
+            {
+                services.AddNamedManagedIdentityCredentialProvider();
+                httpClientBuilder.AddHttpMessageHandler(x =>
+                    ActivatorUtilities.CreateInstance<BearerTokenAuthorizationMessageHandler>(x, new object[] { uri, new ManagedIdentityAuthService() }));
```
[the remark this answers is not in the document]
kenziedolish: My initial thinking here was to instantiate the BearerTokenAuthorizationMessageHandler with the logger argument provided directly from the IServiceProvider. However, since we're now configuring and building the logger in AddFhirClient() it makes sense to pass it a new instance of BearerTokenAuthMessageHandler. I've updated the code to reflect that

src/lib/Microsoft.Health.Extensions.Fhir.R4/ServiceCollectionExtensions.cs
kenziedolish: Thanks for sharing this. My understanding here is that the implicitly captured contents of the entire method outside the lambda will not affect the instantiation of the FhirClient. However, I may be incorrect. Happy to discuss further

kyclai (approved):

kyclai (approved):

src/lib/Microsoft.Health.Extensions.Fhir.R4/ServiceCollectionExtensions.cs
dustinburson: You should be fine. Nothing looked problematic in this code but I wanted to make you aware of the concept in principal.  Captured variables like this can artificially extend the life time of objects.

src/lib/Microsoft.Health.Extensions.Fhir.R4/Service/FhirService.cs
dustinburson: Thanks for the context.  I am fine with this as is and we can re-evaluate later if needed based on some of the things we discussed offline.

src/lib/Microsoft.Health.Extensions.Fhir.R4/IdentifierExtensions.cs
dustinburson: Just wanted to confirm the ToSearchToken does everything you need.  I was originally just referring to using the ToSearchToken for composing the right hand side of the equal sign for identifier={identifier.System}|{identifier.Value}.

I believe with this change the ToSearchToken will only return the properly formatted identifier which will either be system|identifier or just identifer.  You may still need the identifer= portion.

I.E. the change would be to have your extension (moved to the common extension file) that did something like return $"identifier={identifier.ToSearchToken()}";

If the current code is working as expected no changes needed but would like to verify before completing the PR. Thanks!

dustinburson (approved):
Approved with one caveat.  Please double check the reply on old IdentityExtensions file thread.

src/lib/Microsoft.Health.Extensions.Fhir.R4/IdentifierExtensions.cs
kenziedolish: Thanks for pointing this out. I traced both ToSearchToken and ToSearchQueryParameter methods and their output is different as you suggested.

ToSearchQueryParameter
Input: { Value="android" System=null }
Output: identifier=|android

ToSearchToken
Input: { Value="android" System=null }
Output: android

My latest changes moved ToSearchQueryParameter to the common extension file and returns $"identifier={identifier.ToSearchToken()}" which leads to the following output for this test case

Latest changes
Input: { Value="android" System=null }
Output: identifier=android

Now the identifier is properly formatted for the search methods. Let me know if you have any further feedback
Thanks!

PR #175, opened 2022-03-11
Host R4 Templates inside of own project

rogordon01:
The R4 Templates are required inside of our provisioning service. This PR moves the R4 Templates inside of their own project which targets netstandard2.0.


src/lib/Microsoft.Health.Fhir.R4.Ingest.Templates/Extensions/DateTimeExtensions.cs
```diff
@@ -8,7 +8,7 @@
 using EnsureThat;
 using Hl7.Fhir.Model;
 
-namespace Microsoft.Health.Extensions.Fhir
+namespace Microsoft.Health.Fhir.R4.Ingest.Templates.Extensions
```
rogordon01, line 11: The project where this file was hosted currently targets net6.0. Since these methods are only used by the R4 templates I moved this class into the new `Health.Fhir.R4.Ingest.Tempates` project.

dustinburson (approved):

PR #176, opened 2022-03-14
Throw proper exception when TypeNameMatch expression is missing

rogordon01:
This PR ensure that we throw a better exception when the `typeNameMatch` expression is not provided. Now, the following exception will be thrown: `An expression must be set for [TypeNameMatchExpression]`

### DIFF
--- a/test/Microsoft.Health.Fhir.Ingest.Template.UnitTests/IotJsonPathContentTemplateTests.cs
+++ b/test/Microsoft.Health.Fhir.Ingest.Template.UnitTests/IotJsonPathContentTemplateTests.cs
@@ -5,7 +5,6 @@
 
 using System.Collections.Generic;
 using System.Linq;
-using Microsoft.Health.Fhir.Ingest.Template;
 using Microsoft.Health.Tests.Common;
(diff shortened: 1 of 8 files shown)


dustinburson (approved):

test/Microsoft.Health.Fhir.Ingest.Template.UnitTests/IotJsonPathContentTemplateTests.cs
```diff
@@ -36,6 +35,16 @@ public class IotJsonPathContentTemplateTests
             },
         });
 
+        private static readonly IContentTemplate SingleValueMissingTypeNameTemplate = BuildMeasurementExtractor(new IotJsonPathContentTemplate
```
kyclai, line 38: Nit: This field doesn't seem to be referenced in Line 73's test or anywhere else.

rogordon01: You are correct; I forgot to delete this. I'll remove it in my next PR

PR #179, opened 2022-03-21
Personal/rogordon/validation aggregration updates

rogordon01:
This PR updates the validation logic to support aggregation of device event errors.  Also, it adds a validation category to identify what part of the processing pipeline an error was encountered in.

### DIFF
--- a/src/lib/Microsoft.Health.Fhir.Ingest.Validation/MappingValidator.cs
+++ b/src/lib/Microsoft.Health.Fhir.Ingest.Validation/MappingValidator.cs
@@ -77,6 +77,20 @@ namespace Microsoft.Health.Fhir.Ingest.Validation
                 return validationResult;
             }
 
+            ValidateDeviceEvents(deviceEvents, contentTemplate, fhirTemplate, validationResult);
+
+            return validationResult;
+        }
+
+        /// <summary>
+        /// Validates device events. This method then enriches the passed in ValidationResult object with DeviceResults.
+        /// </summary>
+        /// <param name="deviceEvents">The device events to validate</param>
+        /// <param name="contentTemplate">The device mapping template</param>
+        /// <param name="fhirTemplate">The fhir mapping template</param>
+        /// <param name="validationResult">The ValidationResult</param>
+        protected virtual void ValidateDeviceEvents(IEnumerable<JToken> deviceEvents, IContentTemplate contentTemplate, ILookupTemplate<IFhirTemplate> fhirTemplate, ValidationResult validationResult)
+        {
             foreach (var payload in deviceEvents)
             {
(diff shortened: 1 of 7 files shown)


src/lib/Microsoft.Health.Fhir.Ingest.Validation/MappingValidator.cs
```diff
@@ -195,16 +253,16 @@ private void ProcessDeviceEvent(JToken deviceEvent, IContentTemplate contentTemp
...
             }
             catch (Exception e)
             {
-                validationResult.CaptureException(e);
+                validationResult.CaptureException(e, ValidationCategory.NORMALIZATION);
             }
         }
 
-        private void ProcessNormalizedeEvent(Measurement normalizedEvent, ILookupTemplate<IFhirTemplate> fhirTemplate, DeviceResult validationResult)
+        protected virtual void ProcessNormalizedEvent(Measurement normalizedEvent, ILookupTemplate<IFhirTemplate> fhirTemplate, DeviceResult validationResult)
```
kenziedolish, line 265: Good catch

src/lib/Microsoft.Health.Fhir.Ingest.Validation/MappingValidator.cs
```diff
@@ -168,23 +222,27 @@ private void CheckForTemplateCompatibility(IContentTemplate contentTemplate, ILo
...
                 catch (Exception e)
                 {
-                    validationResult.CaptureException(e);
+                    validationResult.CaptureException(e, ValidationCategory.FHIRTRANSFORMATION);
                 }
             }
         }
 
-        private void ProcessDeviceEvent(JToken deviceEvent, IContentTemplate contentTemplate, DeviceResult validationResult)
+        protected virtual void ProcessDeviceEvent(JToken deviceEvent, IContentTemplate contentTemplate, DeviceResult validationResult)
```
kenziedolish, line 245: Did you make this change so that a derived class can leverage ProcessDeviceEvent() and ProcessNormalizedEvent()? Just curious to learn more

kenziedolish (approved):

src/lib/Microsoft.Health.Fhir.Ingest.Validation/MappingValidator.cs
rogordon01: Correct. I was originally going to create a subclass of the MappingValidator which would perform aggregated validations. But after thinking more I simply kept that logic inside of this class.

kenziedolish: Makes sense, thanks. Feel free to leave it as is

PR #187, opened 2022-04-22
Improve Observation Value Merge Type Mismatch Errors

dustinburson:
* Add additional error information on existing type when merge value detect type mismatch.
* Add additional check to fail observation save if the observation found on the initial get doesn't contain the desired identifier.

### DIFF
--- a/src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4FhirImportService.cs
+++ b/src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4FhirImportService.cs
@@ -76,6 +76,12 @@ namespace Microsoft.Health.Fhir.Ingest.Service
             if (!_observationCache.TryGetValue(cacheKey, out Model.Observation existingObservation))
             {
                 existingObservation = await GetObservationFromServerAsync(identifier).ConfigureAwait(false);
+
+                // Discovered an issue where FHIR Service is only matching on first 128 characters of the identifier.  This is a temporary measure to prevent merging of different observations until a fix is available.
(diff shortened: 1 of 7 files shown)


src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4FhirImportService.cs
```diff
@@ -76,6 +76,12 @@ public virtual async Task<string> SaveObservationAsync(ILookupTemplate<IFhirTemp
             if (!_observationCache.TryGetValue(cacheKey, out Model.Observation existingObservation))
             {
                 existingObservation = await GetObservationFromServerAsync(identifier).ConfigureAwait(false);
+
+                // Discovered an issue where FHIR Service is only matching on first 128 characters of the identifier.  This is a temporary measure to prevent merging of different observations until a fix is available.
+                if (existingObservation != null && !existingObservation.Identifier.Exists(i => i.IsExactly(identifier)))
+                {
+                    throw new NotSupportedException("FHIR Service returned matching observation but expected identifier was not present.");
```
pallar-ms, line 83: nit: Just curious if we are allowed to include the identifier in the exception message?

pallar-ms (approved):

src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4FhirImportService.cs
dustinburson: It could be PII so no.

PR #189, opened 2022-05-02
Address dependabot alert

kenziedolish:
This PR addresses the security vulnerability in the OSS Data Mapper project

Alert: https://github.com/microsoft/iomt-fhir/security/dependabot/24

### DIFF
--- a/tools/data-mapper/Microsoft.Health.Tools.DataMapper/ClientApp/package.json
+++ b/tools/data-mapper/Microsoft.Health.Tools.DataMapper/ClientApp/package.json
@@ -20,6 +20,11 @@
     "redux": "4.0.4",
     "redux-thunk": "2.3.0"
   },
+  "overrides": {
+    "jake": {
+      "async": "3.2.2"
+    }
+  },
   "devDependencies": {
     "@types/codemirror": "^5.60.2",
     "@types/history": "4.7.3",


tools/data-mapper/Microsoft.Health.Tools.DataMapper/ClientApp/package.json
```diff
@@ -20,6 +20,11 @@
     "redux": "4.0.4",
     "redux-thunk": "2.3.0"
   },
+  "overrides": {
```
kyclai, line 23: I don't think overrides is needed. An `npm audit fix` should fix the vulnerability by upgrading the async and jake packages, which would modify package-lock.json.

kenziedolish: When I ran npm audit fix, my package manager wanted to degrade react-scripts to 2.3.1 which is a breaking change. I chose to override the async dependency on jake as not to regress any existing behavior

Maybe something is different with my package manager. Happy to discuss further

rogordon01: FYI: `overrides` only works with NPM version 8+. So hopefully dependabot is using that version or else this might not work. 
Also, Our docs for DataMapper specify using NPM version 6 or later. If we are introducing a NPM 8+ change, should we also update the doc? https://github.com/microsoft/iomt-fhir/tree/main/tools/data-mapper#client-app

kenziedolish: Makes sense, I have npm v8.5.0 on my machine which is likely why there is discrepancies between the output of npm audit fix for Carrie and I

In the future, we should consider upgrading this project to use NPM v8 because of the performance and development benefits. For now, I can either close this PR and if either of you have npm v6 and can upgrade it that'd be great. Otherwise I can degrade my version of npm and resubmit it

kyclai: Thanks for pointing out the docs, rogordon01. It looks like Dependabot is using npm 8 since it was updated on [Feb 25](https://github.com/dependabot/dependabot-core/blob/main/CHANGELOG.md#v01750-25-february-2022) and we haven't configured our own dependabot.yml in our repo.

@kenziedolish and I discussed the fix offline. Since I have npm v6.14.13 I can commit the quick npm 6 fix for now. I agree with kenziedolish's suggestion of upgrading to npm 8, so I'll create a backlog item for it and mention updating the doc.

PR #190, opened 2022-05-02
Data Mapper: Resolved package vulnerabilities

kyclai:
Updated packages with `npm audit fix` to resolve vulnerabilities and address a Dependabot alert.

Tested Data Mapper after updating the packages and ensured that it still runs as before: it is possible to add/edit/rename/delete/export/import mappings and run Normalization and FHIR Transformation tests.


rogordon01 (approved):
Is the only change in this PR to `package-lock.json`?

kyclai:
> Is the only change in this PR to `package-lock.json`?

Yep, and we commit package-lock.json because it ensures we all have the same dependencies installed.

PR #192, opened 2022-05-09
Add line numbers to template exceptions

rogordon01:
This PR lays the groundwork to store line numbers inside of POCO objects that are created from user mapping files. These line numbers will then be added to exceptions/telemetry. This allows us to associate errors with the position within a template where they occurred.

Several areas of Normalization are updated to take advantage of line numbers. Additional work for other areas within Normalization and Fhir Transformation will occur in subsequent PRs.

Also, this PR adjusts how errors are captured during deserialization of the initial JSON object. Line numbers are now stored when available. This information will be used in a later PR when sending details back to the Health Portal.

### DIFF
--- a/src/lib/Microsoft.Health.Fhir.Ingest.Template/CollectionContentTemplateFactory.cs
+++ b/src/lib/Microsoft.Health.Fhir.Ingest.Template/CollectionContentTemplateFactory.cs
@@ -3,6 +3,7 @@
 // Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // -------------------------------------------------------------------------------------------------
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using EnsureThat;
@@ -44,11 +45,27 @@ namespace Microsoft.Health.Fhir.Ingest.Template
                 }
                 catch (InvalidTemplateException ex)
                 {
-                    errors.Add(new TemplateError(ex.Message));
+                    errors.Add(new TemplateError(ex.Message, ex.GetLineInfo));
                 }
                 catch (JsonSerializationException ex)
                 {
+                    errors.Add(new TemplateError(ex.Message, new LineInfo() { LineNumber = ex.LineNumber, LinePosition = ex.LinePosition }));
+                }
+                catch (AggregateException ex)
+                {
                     errors.Add(new TemplateError(ex.Message));
+
+                    foreach (var innerException in ex.InnerExceptions)
+                    {
+                        if (innerException is InvalidTemplateException ite)
+                        {
+                            errors.Add(new TemplateError(ite.Message, ite.GetLineInfo));
+                        }
+                        else
+                        {
+                            errors.Add(new TemplateError(innerException.Message));
+                        }
+                    }
                 }
             }
 
--- a/src/lib/Microsoft.Health.Fhir.Ingest.Template/CollectionContentTemplateFactory.cs
+++ b/src/lib/Microsoft.Health.Fhir.Ingest.Template/CollectionContentTemplateFactory.cs
@@ -53,18 +53,9 @@ namespace Microsoft.Health.Fhir.Ingest.Template
                 }
                 catch (AggregateException ex)
                 {
-                    errors.Add(new TemplateError(ex.Message));
-
-                    foreach (var innerException in ex.InnerExceptions)
+                    foreach (var error in ex.ConvertExceptionToTemplateErrors())
                     {
-                        if (innerException is InvalidTemplateException ite)
-                        {
-                            errors.Add(new TemplateError(ite.Message, ite.GetLineInfo));
-                        }
-                        else
-                        {
-                            errors.Add(new TemplateError(innerException.Message));
-                        }
+                        errors.Add(error);
                     }
                 }
             }
--- a/src/lib/Microsoft.Health.Fhir.Ingest.Template/ILineInfo.cs
+++ b/src/lib/Microsoft.Health.Fhir.Ingest.Template/ILineInfo.cs
@@ -5,15 +5,12 @@
 
 namespace Microsoft.Health.Fhir.Ingest.Template
 {
-    public class LineInfo : ILineInfo
+    public interface ILineInfo
     {
-        public int LineNumber { get; set; } = -1;
+        public int LineNumber { get; set; }
 
-        public int LinePosition { get; set; } = -1;
+        public int LinePosition { get; set; }
 
-        public bool HasLineInfo()
-        {
-            return LineNumber >= 0 && LinePosition >= 0;
-        }
+        public bool HasLineInfo();
     }
 }
--- a/src/lib/Microsoft.Health.Fhir.Ingest.Template/JsonPathCalculatedFunctionContentTemplateAdapter.cs
+++ b/src/lib/Microsoft.Health.Fhir.Ingest.Template/JsonPathCalculatedFunctionContentTemplateAdapter.cs
@@ -3,6 +3,7 @@
 // Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // -------------------------------------------------------------------------------------------------
 
+using System.Collections.Generic;
 using System.Linq;
 using EnsureThat;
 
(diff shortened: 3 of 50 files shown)


src/lib/Microsoft.Health.Fhir.Ingest.Template/ILineInfo.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace Microsoft.Health.Fhir.Ingest.Template
+{
+    public interface ILineInfo
+    {
+        public int LineNumber { get; set; }
+
+        public int LinePosition { get; set; }
+
+        public bool HasLineInfo();
```
kenziedolish, line 14: If hasLineInfo = false, what are the values of LineNumber and LinePosition? Both zero?

rogordon01:
They would be -1

Get Outlook for iOS<https://aka.ms/o0ukef>

dustinburson:
@rogordon01 I haven't had a chance to do a full review yet but I l did a pass at a high level and the changes look good.  

I do have a few questions that are more conceptual that I want to get your thoughts on.

1. While I think the changes are overall a net positive, I am curious on the potential performance implications, especially around memory consumption with large mapping templates.  Any thoughts on how we might be able to measure the impact?
2. Related to above, I am wondering how much overhead this adds when parsing and creating the mapping templates in memory.  Right now each batch attempt regenerates the mappings fresh from blob to ensure we have the latest.  This has been fast and efficient enough to not warrant further optimization.  But if this change or others changes that we might need to consider further enchancements.

rogordon01:
> @rogordon01 I haven't had a chance to do a full review yet but I l did a pass at a high level and the changes look good.
> 
> I do have a few questions that are more conceptual that I want to get your thoughts on.
> 
> 1. While I think the changes are overall a net positive, I am curious on the potential performance implications, especially around memory consumption with large mapping templates.  Any thoughts on how we might be able to measure the impact?

My first thought is that the changes introduced don't dramatically alter the memory used during de-serialization. We currently deserialize the entire mapping file into [memory](https://github.com/microsoft/iomt-fhir/blob/main/src/lib/Microsoft.Health.Fhir.Ingest.Template/CollectionTemplateFactory.cs#L46). Transformation of that into a POCO should be reusing objects as needed as that is the [default behavior](https://www.newtonsoft.com/json/help/html/P_Newtonsoft_Json_JsonSerializerSettings_ObjectCreationHandling.htm). And the JsonConverters introduced should only be creating specific objects as needed and not recreating things like the entire ContentTemplate. Having said that, I'm for throwing this up into a test cluster and doing some addition perf testing.
> 2. Related to above, I am wondering how much overhead this adds when parsing and creating the mapping templates in memory.  Right now each batch attempt regenerates the mappings fresh from blob to ensure we have the latest.  This has been fast and efficient enough to not warrant further optimization.  But if this change or others changes that we might need to consider further enchancements.

Perf testing should help us understand the impact of this change. But if this change does incur cost, I'd welcome optimizations in our mapping generation code path. The reason is, regeneration of mappings in the manner that you described seems like overkill. For example, if we processed ~35k messages per minute, and our batch size is 100, that is 350 batches per minute = 350 mapping regenerations per minute. Mapping file do not change at a rate to take advantage of that frequency.

src/lib/Microsoft.Health.Fhir.Ingest.Template/CollectionContentTemplateFactory.cs
```diff
@@ -44,11 +45,27 @@ protected override IContentTemplate BuildCollectionTemplate(JArray templateColle
...
                 }
                 catch (JsonSerializationException ex)
+                {
+                    errors.Add(new TemplateError(ex.Message, new LineInfo() { LineNumber = ex.LineNumber, LinePosition = ex.LinePosition }));
+                }
+                catch (AggregateException ex)
                 {
                     errors.Add(new TemplateError(ex.Message));
+
+                    foreach (var innerException in ex.InnerExceptions)
```
dustinburson, line 58: Suggestion: Code is repeated a couple times.  Would be good to extract into an extension or a common path.

src/lib/Microsoft.Health.Fhir.Ingest.Template/JsonPathCalculatedFunctionContentTemplateAdapter.cs
```diff
@@ -16,32 +16,56 @@ public JsonPathCalculatedFunctionContentTemplateAdapter(TTemplate innerTemplate)
...
+                    return new TemplateExpression(value, TemplateExpressionLanguage.JsonPath);
+                }
+
+                var templateExpression = new TemplateExpression(value, TemplateExpressionLanguage.JsonPath)
+                {
+                    LineNumber = lineAwareJsonObject.LineNumber,
+                    LinePosition = lineAwareJsonObject.LinePosition,
+                };
+
+                templateExpression.LineInfoForProperties[nameof(TemplateExpression.Value)] = new LineInfo()
```
dustinburson, line 62: How do we ensure we don't conflict with a property named Value?  Referring to LineAwareJsonObjectExtensions line 45.

src/lib/Microsoft.Health.Fhir.Ingest.Template/MeasurementExtractor.cs
```diff
@@ -77,24 +78,26 @@ protected T EvalExpression<T>(JToken token, string name, bool isRequired = false
...
                         {
-                            exceptions.Add(new IncompatibleDataException($"Encounted an error while extracting value for [{name}] using expression {expression.Value}", e));
+                            exceptions.Add(new IncompatibleDataException($"Encounted an error while extracting value for [{name}] using expression {expression.Value}", e, valueLineInfo));
                         }
                     }
                 }
 
                 if (isRequired)
                 {
+                    var lineInfo = expressions.First(e => e != null);
```
dustinburson, line 88: First() will throw an exception if there aren't elements that satisfy the criteria.  I don't think this is what you want given the checks you have below.  You can use FirstOrDefault() and the lineInfo expression will be null if nothing is found.

If you do want the First() call to throw if nothing is found it will be an invalid operation exception, https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.first?view=net-6.0.  You will want to make sure the code up the stack is handling it properly.

dustinburson (commented):
Overall changes look good.  Added a few comments for things to double check.

src/lib/Microsoft.Health.Fhir.Ingest.Template/CollectionContentTemplateFactory.cs
rogordon01: I'll throw this in an extension method.

src/lib/Microsoft.Health.Fhir.Ingest.Template/JsonPathCalculatedFunctionContentTemplateAdapter.cs
rogordon01: This is a new `TemplateExpression` object (created on line 56). Its `LineInfoForProperties` object is empty so there would be nothing to conflict with. 

In general, a `TemplateExpression` only has 2 properties: `value` and `language`. Language is optional, so I don't see it its position. Its `value` can be specified 2 ways, shown below. In the first case, we set the position of `value` to that of the overall `TemplateExpression`. In the second case, we are setting its line position to that of the explicit object `value`.

```json
"deviceIdExpression": "$.device"
```

```json
"deviceIdExpression": {
      "value": "$.device"
    },
```
```json
"deviceIdExpression": {
      "value": "$.device",
      "language": "JsonPath"
    },
```

src/lib/Microsoft.Health.Fhir.Ingest.Template/MeasurementExtractor.cs
rogordon01: Good point. The check on line 48 ensures that I have at least one non-null `TemplateExpression`. So this would never throw. But looking at it more, I think I'll refactor this whole method to take in a single `TemplateExpression` versus an var-args list of expressions. That will simplify the logic here.

rogordon01: Actually nevermind; it looks like we do require the use of var-ags list.

rogordon01:
Deployed latest changes to dev AKS cluster. Ran performance tests. Noted that Normalization is still able to process ~45k messages per minute, which is the same before the changes in this PR.

image

src/lib/Microsoft.Health.Fhir.Ingest.Template/JsonPathCalculatedFunctionContentTemplateAdapter.cs
dustinburson: Thanks for the clarification.  No concern then!

dustinburson (approved):
Changes look good.  Thanks!

PR #196, opened 2022-06-03
Update how measurements are processed

rogordon01:
This PR makes several enhancements to how Measurements are produced from DeviceEvents:

- NormalizedEvents are emitted to the processing pipeline as soon as they are created. Previously, all NormalizedEvents for a single DeviceEvent were added to a collection before being emitted to the processing pipeline
- Reusing a single JObject clone during template evaluation 
- Added a metric to capture processing time per NormalizedEvent

### DIFF
--- a/src/lib/Microsoft.Health.Fhir.Ingest.Template/MeasurementExtractor.cs
+++ b/src/lib/Microsoft.Health.Fhir.Ingest.Template/MeasurementExtractor.cs
@@ -126,16 +126,12 @@ namespace Microsoft.Health.Fhir.Ingest.Template
             EnsureArg.IsNotNull(token, nameof(token));
             var evaluator = CreateRequiredExpressionEvaluator(Template.TypeMatchExpression, nameof(Template.TypeMatchExpression));
 
-            var counter = 0;
-
             JObject tokenClone = null;
 
             foreach (var extractedToken in evaluator.SelectTokens(token))
             {
                 // Add the extracted data as an element of the original data.
                 // This allows subsequent expressions access to data from the original event data
-                counter++;
-
                 if (tokenClone == null)
                 {
                     tokenClone = new JObject(token);
@@ -146,7 +142,7 @@ namespace Microsoft.Health.Fhir.Ingest.Template
                 tokenClone.Remove(MatchedToken);
             }
 
-            // return Enumerable.Empty<JToken>();
+            tokenClone = null;
         }
 
         protected IExpressionEvaluator CreateRequiredExpressionEvaluator(TemplateExpression expression, string expressionName)
--- a/src/lib/Microsoft.Health.Fhir.Ingest.Template/MeasurementExtractor.cs
+++ b/src/lib/Microsoft.Health.Fhir.Ingest.Template/MeasurementExtractor.cs
@@ -131,7 +131,7 @@ namespace Microsoft.Health.Fhir.Ingest.Template
                 // Add the extracted data as an element of the original data.
                 // This allows subsequent expressions access to data from the original event data
 
-                var tokenClone = token.DeepClone() as JObject;
+                var tokenClone = new JObject(token);
                 tokenClone.Add(MatchedToken, extractedToken);
                 yield return tokenClone;
             }
--- a/src/lib/Microsoft.Health.Fhir.Ingest.Template/MeasurementExtractor.cs
+++ b/src/lib/Microsoft.Health.Fhir.Ingest.Template/MeasurementExtractor.cs
@@ -126,15 +126,27 @@ namespace Microsoft.Health.Fhir.Ingest.Template
             EnsureArg.IsNotNull(token, nameof(token));
             var evaluator = CreateRequiredExpressionEvaluator(Template.TypeMatchExpression, nameof(Template.TypeMatchExpression));
 
+            var counter = 0;
+
+            JObject tokenClone = null;
+
             foreach (var extractedToken in evaluator.SelectTokens(token))
             {
                 // Add the extracted data as an element of the original data.
                 // This allows subsequent expressions access to data from the original event data
+                counter++;
 
-                var tokenClone = new JObject(token);
+                if (tokenClone == null)
+                {
+                    tokenClone = new JObject(token);
+                }
+
                 tokenClone.Add(MatchedToken, extractedToken);
                 yield return tokenClone;
+                tokenClone.Remove(MatchedToken);
             }
+
+            // return Enumerable.Empty<JToken>();
         }
 
         protected IExpressionEvaluator CreateRequiredExpressionEvaluator(TemplateExpression expression, string expressionName)
--- a/src/lib/Microsoft.Health.Fhir.Ingest/Data/EventDataWithJsonBodyToJTokenConverter.cs
+++ b/src/lib/Microsoft.Health.Fhir.Ingest/Data/EventDataWithJsonBodyToJTokenConverter.cs
@@ -3,9 +3,11 @@
 // Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // -------------------------------------------------------------------------------------------------
 
+using System.IO;
 using System.Text;
 using EnsureThat;
 using Microsoft.Azure.EventHubs;
+using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 
 namespace Microsoft.Health.Fhir.Ingest.Data
@@ -15,10 +17,18 @@ namespace Microsoft.Health.Fhir.Ingest.Data
         public JToken Convert(EventData input)
         {
             EnsureArg.IsNotNull(input, nameof(input));
+            var body = null as JToken;
 
-            var body = input.Body.Count > 0
-                ? JToken.Parse(Encoding.UTF8.GetString(input.Body.Array, input.Body.Offset, input.Body.Count))
-                : null;
+            if (input.Body.Count > 0)
+            {
+                using (var stream = new MemoryStream(input.Body.Array))
+                using (StreamReader sr = new StreamReader(stream, Encoding.UTF8))
+                using (JsonReader reader = new JsonTextReader(sr))
+                {
+                    body = JToken.ReadFrom(reader);
+                }
+            }
+
             var data = new { Body = body, input.Properties, input.SystemProperties };
             var token = JToken.FromObject(data);
             return token;
--- a/src/lib/Microsoft.Health.Fhir.Ingest/Data/EventDataWithJsonBodyToJTokenConverter.cs
+++ b/src/lib/Microsoft.Health.Fhir.Ingest/Data/EventDataWithJsonBodyToJTokenConverter.cs
@@ -14,10 +14,16 @@ namespace Microsoft.Health.Fhir.Ingest.Data
 {
     public class EventDataWithJsonBodyToJTokenConverter : IConverter<EventData, JToken>
     {
+        private const string BodyAttr = "Body";
+        private const string PropertiesAttr = "Properties";
+        private const string SystemPropertiesAttr = "SystemProperties";
+        private readonly JsonSerializer jsonSerializer = JsonSerializer.CreateDefault();
+
         public JToken Convert(EventData input)
         {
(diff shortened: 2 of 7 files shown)


src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementEventNormalizationService.cs
```diff
@@ -201,5 +189,70 @@ private Task<bool> ProcessErrorAsync(Exception ex, EventData data)
...
+                    }
+                    else
+                    {
+                        break;
+                    }
+                }
+                catch (Exception ex)
+                {
+                    // Translate all Normalization Mapping exceptions into a common type for easy identification.
+                    storeNormalizedException(ex);
```
rogordon01, line 247: @dustinburson Do we skip processing the remainder of the DeviceEvent now that one Measurement produced an Exception? Or do we simply skip this Measurement, and allow the remaining Measurements to be generated?

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementEventNormalizationService.cs
```diff
@@ -201,5 +189,70 @@ private Task<bool> ProcessErrorAsync(Exception ex, EventData data)
             var handled = _exceptionTelemetryProcessor.HandleException(ex, _log);
             return Task.FromResult(!handled);
         }
+
+        private IEnumerable<(string, IMeasurement)> CreateMeasurementIterator(JToken token, string partitionId, ConcurrentBag<Exception> exceptions, Func<Exception, EventData, Task<bool>> errorConsumer, EventData evt)
```
rogordon01, line 193: Previously, we batched all Measurements per DeviceEvent before sending down the processing pipeline. If there was an error with one Measurement we wouldn't send any Measurements for that DeviceEvent.

This change streams each `Measurement` onto the processing pipeline as soon as it is generated

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementEventNormalizationService.cs
```diff
@@ -201,5 +189,70 @@ private Task<bool> ProcessErrorAsync(Exception ex, EventData data)
...
+                        break;
+                    }
+                }
+                catch (Exception ex)
+                {
+                    // Translate all Normalization Mapping exceptions into a common type for easy identification.
+                    storeNormalizedException(ex);
+                    shouldLoop = false;
+                }
+
```
rogordon01, line 250: @dustinburson Do we skip processing the remainder of the DeviceEvent now that one Measurement produced an Exception? Or do we simply skip this Measurement, and allow the remaining Measurements to be generated?

src/lib/Microsoft.Health.Fhir.Ingest.Template/MeasurementExtractor.cs
```diff
@@ -126,15 +126,23 @@ protected virtual IEnumerable<JToken> MatchTypeTokens(JObject token)
...
                 // This allows subsequent expressions access to data from the original event data
+                if (tokenClone == null)
+                {
+                    tokenClone = new JObject(token);
+                }
 
-                var tokenClone = token.DeepClone() as JObject;
                 tokenClone.Add(MatchedToken, extractedToken);
                 yield return tokenClone;
+                tokenClone.Remove(MatchedToken);
```
dustinburson, line 142: Does this work?  We are returning the tokenClone and them updating the same reference to remove the property.  I don't think we can do the remove.  Anything downstream of this will potentially be missing the MatchedToken.

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementEventNormalizationService.cs
```diff
@@ -45,7 +46,7 @@ public class MeasurementEventNormalizationService : IDataNormalizationService<Ev
             Data.IConverter<EventData, JToken> converter,
             IExceptionTelemetryProcessor exceptionTelemetryProcessor,
             int maxParallelism,
-            int asyncCollectorBatchSize = 200)
+            int asyncCollectorBatchSize = 25)
```
dustinburson, line 49: Do we have any data on the impact of this change?

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementEventNormalizationService.cs
```diff
@@ -201,5 +189,70 @@ private Task<bool> ProcessErrorAsync(Exception ex, EventData data)
...
+
+            (string, IMeasurement) currentValue = (partitionId, null);
+            var shouldLoop = true;
+            var stopWatch = new Stopwatch();
+
+            while (shouldLoop)
+            {
+                try
+                {
+                    stopWatch.Start();
```
dustinburson, line 223: We have some helper classes that simplify this.  We may need to port them from PaaS to OSS.

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementEventNormalizationService.cs
```diff
@@ -115,19 +115,7 @@ private async Task StartConsumer(ISourceBlock<EventData> producer, IEnumerableAs
 
                         var token = _converter.Convert(evt);
 
-                        try
-                        {
-                            foreach (var measurement in _contentTemplate.GetMeasurements(token))
-                            {
-                                measurement.IngestionTimeUtc = evt.SystemProperties.EnqueuedTimeUtc;
-                                createdMeasurements.Add((partitionId, measurement));
```
dustinburson, line 123: Is there a reason we can't just yield return here?

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementEventNormalizationService.cs
```diff
@@ -201,5 +189,70 @@ private Task<bool> ProcessErrorAsync(Exception ex, EventData data)
...
+                var normalizationException = new NormalizationDataMappingException(ex);
+                if (errorConsumer(normalizationException, evt).ConfigureAwait(false).GetAwaiter().GetResult())
+                {
+                    exceptions.Add(normalizationException);
+                }
+            };
+
+            try
+            {
+                enumerable = _contentTemplate.GetMeasurements(token).GetEnumerator();
```
dustinburson, line 207: Shouldn't be we able to remove the lower while loop and just yield return as we iterate through each element here?

src/lib/Microsoft.Health.Fhir.Ingest.Template/MeasurementExtractor.cs
rogordon01: This should work as intended as this is a synchronous `IEnumerable`. We first `yield` the cloned `JToken` and then produce a `Measurement` from it. Finally we return here to remove the `MatchedToken` property and start the loop again. The code producing the `Measurement` does not hold onto a reference for the token it was given. 

We have various unit tests which should be testing out this scenario, such as: https://github.com/microsoft/iomt-fhir/blob/main/test/Microsoft.Health.Fhir.Ingest.Template.UnitTests/CalculatedFunctionContentTemplateTests.cs#L79

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementEventNormalizationService.cs
rogordon01: No hard numbers. I had originally put in `200` as a starting point, though. With the intent on tweaking it as needed. I can revert this change.

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementEventNormalizationService.cs
rogordon01: Unfortunately a `yield` statement cannot existing inside of a `try` block if there is a `catch` block.

https://docs.microsoft.com/en-us/archive/msdn-magazine/2017/june/essential-net-custom-iterators-with-yield#yield-statement-requirements

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementEventNormalizationService.cs
rogordon01: See my comment about `yield` statements inside of try/catch blocks.

rogordon01: Switching this to an `AsyncEnumberable` could be a good idea. At the very least it would simplify things. I did make use of the pipelines ability to batch, though. But we could see if that could be refactored.

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementEventNormalizationService.cs
rogordon01: That is true if we were to spin. But I had thought that we decided not to spin, and simply discard bad data? Until we could send it over to the EMS?

dustinburson: We will continue to spin by default though I did add the ability to change the strategy if desired.

src/lib/Microsoft.Health.Fhir.Ingest/Data/EventDataWithJsonBodyToJTokenConverter.cs
```diff
@@ -15,12 +17,20 @@ public class EventDataWithJsonBodyToJTokenConverter : IConverter<EventData, JTok
...
+            {
+                using (var stream = new MemoryStream(input.Body.Array))
+                using (StreamReader sr = new StreamReader(stream, Encoding.UTF8))
+                using (JsonReader reader = new JsonTextReader(sr))
+                {
+                    body = JToken.ReadFrom(reader);
+                }
+            }
+
+            var data = new { Body = body, input.Properties, input.SystemProperties };
```
dustinburson, line 32: From my testing this actually adds quite a bit of overhead.  In my branch I tweaked it to the following:

```
        private const string BodyAttr = "Body";
        private const string PropertiesAttr = "Properties";
        private const string SystemPropertiesAttr = "SystemProperties";

        private readonly JsonSerializer jsonSerializer = JsonSerializer.CreateDefault();

        public JToken Convert(IEventMessage input)
        {
            EnsureArg.IsNotNull(input, nameof(input));

            JToken token = new JObject();
            JToken body = null;

            if (input.Body.Length > 0)
            {
                using StreamReader streamReader = new StreamReader(input.Body.AsStream(), Encoding.UTF8);
                using JsonReader jsonReader = new JsonTextReader(streamReader);
                body = JToken.ReadFrom(jsonReader);
            }

            token[BodyAttr] = body;
            token[PropertiesAttr] = JToken.FromObject(input.Properties, jsonSerializer);
            token[SystemPropertiesAttr] = JToken.FromObject(input.SystemProperties, jsonSerializer);

            return token;
        }
```

The stream processing will work a little differently in my example (I updated the code to have the original ReadOnlyMemory<byte> and I am taking advantage of that) but the rest should be achievable in this version.  I think the biggest cost is building the new JToken from the object created in line 32.  Starting with an empty JObject and adding the properties to it manually resulted in a ~ 35% increase in number of bytes processed per minute.  I am still not applying the templates but the increase is promising.

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementEventNormalizationService.cs
dustinburson: In the other perf PR we are no longer using this class. Thoughts on reverting this change?  It will still be used for the Azure Function implementation but we wouldn't be taking the risk of potentially introducing a new bug for that path.  The new metric is great, and we can work to port it over to the new implementation.

dustinburson (approved):

rogordon01:
Closing PR as most of these perf improvements have been made redundant by other PRs. The relevant improvements have been moved into PR #199

PR #197, opened 2022-06-13
Add new Streaming EventConsumerService for Normalization

dustinburson:
- Add abstract implementation of a StreamingEventConsumerService that processes events immediately while emitting progress metrics and checkpointing when thresholds are hit.
- Add new Normalization service based on StreamingEventConsumerService.
- Fix bug where checkpoint client will find delete blobs
- Update EventMessageJTokenCoverter to add properties directly to the JObject rather then new object declarations
- Add missing metrics to NormalizationStreamingEventConsumerService


src/lib/Microsoft.Health.Fhir.Ingest/Service/NormalizationStreamingEventConsumerService.cs
```diff
@@ -0,0 +1,119 @@
...
+                    await _collector.AddAsync(measurement);
+                    projectedMeasurements++;
+                }
+
+                RecordNormalizedEventMetrics(eventArg, projectedMeasurements);
+            }
+            catch (Exception ex)
+            {
+                // Translate all Normalization Mapping exceptions into a common type for easy identification.
+                throw new NormalizationDataMappingException(ex);
```
wi-y, line 82: This exception makes sense to me if _contentTemplate.GetMeasurements fails, but if _collector.AddAsync fails due to an Event Hub issue would we want to throw a NormalizationDataMappingException?

dustinburson: Good point.  I will move the try block inside the loop.

wi-y:
The PR looks good to me since it will allow us to control the memory consumed for large messages. JFYI testing with a batch of smaller messages yielded a slight degradation (took about 1.5x longer) when compared to the original code. The new code hits the Event Hub more frequently, but the memory is more controllable so it is probably worth the trade off

src/lib/Microsoft.Health.Events/EventConsumers/StreamingEventConsumerService.cs
```diff
@@ -0,0 +1,141 @@
...
+using Microsoft.Health.Events.Telemetry;
+using Microsoft.Health.Logging.Telemetry;
+
+namespace Microsoft.Health.Events.EventConsumers
+{
+    public abstract class StreamingEventConsumerService : IEventConsumerService
+    {
+        protected const string ThresholdWaitReached = nameof(ThresholdWaitReached);
+        protected const string ThresholdCountReached = nameof(ThresholdCountReached);
+        protected const int PartitionCheckpointThreshold = 100;
```
rogordon01, line 24: I'd add this into the `EventBatchingOptions`, with a default of 100. That allows us to experiment with this value by changing an environment variable.

src/lib/Microsoft.Health.Fhir.Ingest/Data/EventMessageJTokenConverter.cs
```diff
@@ -0,0 +1,47 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System.IO;
+using System.Text;
+using EnsureThat;
+using Microsoft.Health.Events.Model;
+using Microsoft.Toolkit.HighPerformance;
```
rogordon01, line 10: Out of curiosity, how is this package being used?

src/lib/Microsoft.Health.Fhir.Ingest/Service/NormalizationStreamingEventConsumerService.cs
```diff
@@ -0,0 +1,119 @@
...
+
+        private async Task NormalizeMessage(JToken token, IEventMessage eventArg)
+        {
+            try
+            {
+                int projectedMeasurements = 0;
+                foreach (var measurement in _contentTemplate.GetMeasurements(token))
+                {
+                    measurement.IngestionTimeUtc = eventArg.EnqueuedTime.UtcDateTime;
+                    await _collector.AddAsync(measurement);
```
rogordon01, line 73: This will cause a single Measurement to be sent over to the internal EventHub at a time. In my testing it is faster to send over a batch of events versus a single event. 
An argument could be made that the `IEnumerableAsyncCollector` should handle the batching, and perhaps accept a `flush` call. But that would require a refactor.

src/lib/Microsoft.Health.Fhir.Ingest/Service/NormalizationStreamingEventConsumerService.cs
```diff
@@ -0,0 +1,119 @@
...
+                }
+            }
+        }
+
+        private async Task NormalizeMessage(JToken token, IEventMessage eventArg)
+        {
+            try
+            {
+                int projectedMeasurements = 0;
+                foreach (var measurement in _contentTemplate.GetMeasurements(token))
```
rogordon01, line 70: Can we also add telemetry for the time it takes to create a single  Measurement?

src/lib/Microsoft.Health.Fhir.Ingest/Service/NormalizationStreamingEventConsumerService.cs
dustinburson: It should be done in the IAsync collector.  I will see what can be done there.

dustinburson: Actually for now, I may just collect the events from the project and push those as a collection.  Does that match what is being done today?  Collecting and flushing across partitions will be more complicate and should be saved for a later refactor.

rogordon01: Yes, collecting all projections from the DeviceEvent pushing them as a collection is what is done today. I think that a later refactor to deal with that batching in the collector makes sense.

src/lib/Microsoft.Health.Fhir.Ingest/Service/NormalizationStreamingEventConsumerService.cs
dustinburson: Done

src/lib/Microsoft.Health.Fhir.Ingest/Service/NormalizationStreamingEventConsumerService.cs
dustinburson: After moving the try/catch block inside the for loop that is harder to implement now but thanks for the suggestion.

I ended up doing a list that is reused for each message processed.  During processing, the generated measurements are stored and then sent as a batch to the async collector. Afterward the list is cleared of elements and reused for the next message.  The underlying array in the list will remain.  I set the initial capacity to 50 but it can grow beyond.

Long term, you are correct, I would like to make this the responsibility of the async collector.  I am creating a work item of additional tasks not covered by this PR and will include this.

src/lib/Microsoft.Health.Fhir.Ingest/Service/NormalizationStreamingEventConsumerService.cs
```diff
@@ -0,0 +1,142 @@
...
+
+        private async Task NormalizeMessage(JObject token, IEventMessage eventArg)
+        {
+            Stopwatch sw = Stopwatch.StartNew();
+            foreach (var measurement in _contentTemplate.GetMeasurements(token))
+            {
+                try
+                {
+                    measurement.IngestionTimeUtc = eventArg.EnqueuedTime.UtcDateTime;
+                    NormalizationBatch.Add(measurement);
```
rogordon01, line 79: Is this `NormalizationBatch` threadsafe? I ask because I think the previous `IEventConsumerService` implementations are singletons. So this would be used across partitions/threads.

src/lib/Microsoft.Health.Fhir.Ingest/Service/NormalizationStreamingEventConsumerService.cs
```diff
@@ -0,0 +1,142 @@
...
+                }
+                catch (Exception ex)
+                {
+                    // Translate all Normalization Mapping exceptions into a common type for easy identification.
+                    throw new NormalizationDataMappingException(ex);
+                }
+            }
+
+            sw.Stop();
```
rogordon01, line 88: This will record the time to collect all of the projections. It would also be nice to have the time it takes to produce a single projection. But we can add that telemetry in later.

dustinburson: Good point, I was thinking we had a NormalizationStreamingEventConsumerService per partition but that isn't the case.  I switched to a new instance local in the function for now.

rogordon01 (approved):

PR #198, opened 2022-06-15
Add error reporting triggers, include payload of message that triggered the error

wi-y:
When we encounter an exception during Normalization or MeasurementToFhir, we now have the ability to send the exception to a destination (IErrorMessageService) if it is supplied at startup. The exception will include the message payload. Note: for MeasurementToFhir the payload is the normalized payload and not the original device data payload.

There is not implementation of IErrorMessageService included in this PR. However, if one is injected then it will be used.

### DIFF
--- a/src/lib/Microsoft.Health.Common/Errors/IErrorMessageService.cs
+++ b/src/lib/Microsoft.Health.Common/Errors/IErrorMessageService.cs
@@ -1,0 +1,39 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Health.Common.Errors
+{
+    public interface IErrorMessageService
+    {
+        /// <summary>
+        /// Writes an ErrorMessage to a destination
+        /// </summary>
+        /// <remarks>
+        /// The ErrorMessage Type property will be the ExceptionType.
+        /// and Details property will be the Exception Message.
+        /// </remarks>
+        /// <param name="exception">An Exception.</param>
+        /// <param name="cancellationToken">A cancellation token.</param>
+        /// <returns><see cref="Task"/>Returns the successfully stored error message.</returns>
+        Task ReportError(Exception exception, CancellationToken cancellationToken);
+
+        /// <summary>
+        /// Writes an IEnumerable of ErrorMessage to a destination.
+        /// </summary>
+        /// <remarks>
+        /// The ErrorMessage Type property will be the ExceptionType.
+        /// and Details property will be the Exception Message.
+        /// </remarks>
+        /// <param name="exceptions">A collection of exceptions.</param>
+        /// <param name="cancellationToken">A cancellation token.</param>
+        /// <returns><see cref="Task"/>Returns the successfully stored error message.</returns>
+        Task<IEnumerable<ErrorMessage>> ReportError(IEnumerable<Exception> exceptions, CancellationToken cancellationToken);
+    }
+}
--- a/src/lib/Microsoft.Health.Fhir.Ingest/Config/MeasurementFhirImportOptions.cs
+++ b/src/lib/Microsoft.Health.Fhir.Ingest/Config/MeasurementFhirImportOptions.cs
@@ -4,6 +4,7 @@
 // -------------------------------------------------------------------------------------------------
 
 using Microsoft.Health.Common.Config;
+using Microsoft.Health.Common.Errors;
 using Microsoft.Health.Fhir.Ingest.Telemetry;
 using Microsoft.Health.Fhir.Ingest.Template;
 using Microsoft.Health.Logging.Telemetry;
@@ -12,10 +13,24 @@ namespace Microsoft.Health.Fhir.Ingest.Config
 {
     public class MeasurementFhirImportOptions
     {
-        public virtual ParallelTaskOptions ParallelTaskOptions { get; } = new ParallelTaskOptions { MaxConcurrency = 10 };
+        public MeasurementFhirImportOptions()
+        {
+            ParallelTaskOptions = new ParallelTaskOptions { MaxConcurrency = 10 };
+            TemplateFactory = CollectionFhirTemplateFactory.Default;
+            ExceptionService = new FhirExceptionTelemetryProcessor();
+        }
 
-        public virtual IExceptionTelemetryProcessor ExceptionService { get; } = new FhirExceptionTelemetryProcessor();
+        public MeasurementFhirImportOptions(IErrorMessageService errorMessageService)
+        {
+            ParallelTaskOptions = new ParallelTaskOptions { MaxConcurrency = 10 };
+            TemplateFactory = CollectionFhirTemplateFactory.Default;
+            ExceptionService = new FhirExceptionTelemetryProcessor(errorMessageService);
+        }
 
-        public virtual ITemplateFactory<string, ITemplateContext<ILookupTemplate<IFhirTemplate>>> TemplateFactory { get; } = CollectionFhirTemplateFactory.Default;
+        public virtual ParallelTaskOptions ParallelTaskOptions { get; }
+
+        public virtual IExceptionTelemetryProcessor ExceptionService { get; }
+
+        public virtual ITemplateFactory<string, ITemplateContext<ILookupTemplate<IFhirTemplate>>> TemplateFactory { get; }
     }
 }
--- a/src/lib/Microsoft.Health.Fhir.Ingest/Host/DeviceDataNormalizationExtensions.cs
+++ b/src/lib/Microsoft.Health.Fhir.Ingest/Host/DeviceDataNormalizationExtensions.cs
@@ -9,6 +9,7 @@ using Microsoft.Azure.WebJobs;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Options;
+using Microsoft.Health.Common.Errors;
 using Microsoft.Health.Extensions.Host;
 using Microsoft.Health.Fhir.Ingest.Config;
 using Microsoft.Health.Fhir.Ingest.Service;
@@ -46,6 +47,8 @@ namespace Microsoft.Health.Fhir.Ingest.Host
             EnsureArg.IsNotNull(serviceCollection, nameof(serviceCollection));
             EnsureArg.IsNotNull(configuration, nameof(configuration));
 
+            serviceCollection.AddSingleton<IErrorMessageService, ErrorMessageService>();
+
             serviceCollection.Configure<NormalizationServiceOptions>(configuration.GetSection(NormalizationServiceOptions.Settings));
             serviceCollection.AddSingleton(TelemetryProcessorFactory);
 
@@ -56,11 +59,24 @@ namespace Microsoft.Health.Fhir.Ingest.Host
         {
             var options = serviceProvider.GetRequiredService<IOptions<NormalizationServiceOptions>>();
 
-            return options.Value.ErrorHandlingPolicy switch
+            var errorMessageService = serviceProvider.GetService<IErrorMessageService>();
+
+            if (errorMessageService == null)
             {
-                NormalizationErrorHandlingPolicy.DiscardMatchAndExtractErrors => new NormalizationExceptionTelemetryProcessor(typeof(NormalizationDataMappingException)),
-                _ => new NormalizationExceptionTelemetryProcessor(),
-            };
+                return options.Value.ErrorHandlingPolicy switch
+                {
+                    NormalizationErrorHandlingPolicy.DiscardMatchAndExtractErrors => new NormalizationExceptionTelemetryProcessor(typeof(NormalizationDataMappingException)),
+                    _ => new NormalizationExceptionTelemetryProcessor(),
+                };
+            }
+            else
+            {
+                return options.Value.ErrorHandlingPolicy switch
+                {
+                    NormalizationErrorHandlingPolicy.DiscardMatchAndExtractErrors => new NormalizationExceptionTelemetryProcessor(errorMessageService, typeof(NormalizationDataMappingException)),
+                    _ => new NormalizationExceptionTelemetryProcessor(errorMessageService),
+                };
+            }
         }
     }
 }
--- a/src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementEventNormalizationService.cs
+++ b/src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementEventNormalizationService.cs
@@ -198,7 +198,22 @@ namespace Microsoft.Health.Fhir.Ingest.Service
         /// <returns>Returns true if the exception is handled and false otherwise.</returns>
         private Task<bool> ProcessErrorAsync(Exception ex, EventData data)
         {
-            var handled = _exceptionTelemetryProcessor.HandleException(ex, _log);
+            JObject request = new JObject();
+            request["Properties"] = new JObject();
+            request["SystemProperties"] = new JObject();
+
+            foreach (var prop in data.Properties)
+            {
+                request["Properties"][prop.Key] = prop.Value.ToString();
+            }
+
+            foreach (var prop in data.SystemProperties)
+            {
+                request["SystemProperties"][prop.Key] = prop.Value.ToString();
+            }
+
+            request["Body"] = JObject.Parse(System.Text.Encoding.Default.GetString(data.Body.ToArray()));
+            var handled = _exceptionTelemetryProcessor.HandleException(ex, request, _log);
             return Task.FromResult(!handled);
         }
     }
--- a/src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementFhirImportService.cs
+++ b/src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementFhirImportService.cs
@@ -73,7 +73,14 @@ namespace Microsoft.Health.Fhir.Ingest.Service
                             }
                             catch (Exception ex)
                             {
-                                if (!Options.ExceptionService.HandleException(ex, log))
+                                // capture the message payload for error message service
+                                JArray payload = new JArray();
+                                foreach (var measurement in m.Data)
+                                {
+                                    payload.Add(JObject.Parse(System.Text.Encoding.Default.GetString(measurement.Payload.ToArray())));
+                                }
+
+                                if (!Options.ExceptionService.HandleException(ex, payload, log))
                                 {
                                     throw;
                                 }
@@ -111,23 +118,28 @@ namespace Microsoft.Health.Fhir.Ingest.Service
             var partitionId = data.FirstOrDefault()?.PartitionId;
 
             // Deserialize events into measurements and then group according to the device, type, and other factors
-            return data.Select(e => JsonConvert.DeserializeObject<Measurement>(System.Text.Encoding.Default.GetString(e.Body.ToArray())))
+            return data.Select(e =>
+                {
+                    var measurement = JsonConvert.DeserializeObject<Measurement>(System.Text.Encoding.Default.GetString(e.Body.ToArray()));
+                    measurement.Payload = e.Body;
+                    return measurement;
+                })
                 .GroupBy(m => $"{m.DeviceId}-{m.Type}-{m.PatientId}-{m.EncounterId}-{m.CorrelationId}")
                 .Select(g =>
-                {
-                    var measurements = g.ToList();
-                    _ = CalculateMetricsAsync(measurements, log, partitionId).ConfigureAwait(false);
-                    return new MeasurementGroup
                     {
-                        Data = measurements,
-                        MeasureType = measurements[0].Type,
-                        CorrelationId = measurements[0].CorrelationId,
-                        DeviceId = measurements[0].DeviceId,
-                        EncounterId = measurements[0].EncounterId,
-                        PatientId = measurements[0].PatientId,
-                    };
-                })
-                .ToArray();
+                        var measurements = g.ToList();
+                        _ = CalculateMetricsAsync(measurements, log, partitionId).ConfigureAwait(false);
+                        return new MeasurementGroup
+                        {
+                            Data = measurements,
+                            MeasureType = measurements[0].Type,
+                            CorrelationId = measurements[0].CorrelationId,
+                            DeviceId = measurements[0].DeviceId,
+                            EncounterId = measurements[0].EncounterId,
+                            PatientId = measurements[0].PatientId,
+                        };
+                    })
+                    .ToArray();
         }
 
         private static async Task CalculateMetricsAsync(IList<Measurement> measurements, ITelemetryLogger log, string partitionId = null)
--- a/src/lib/Microsoft.Health.Fhir.Ingest/Telemetry/FhirExceptionTelemetryProcessor.cs
+++ b/src/lib/Microsoft.Health.Fhir.Ingest/Telemetry/FhirExceptionTelemetryProcessor.cs
@@ -5,12 +5,14 @@
 
 using System;
 using EnsureThat;
+using Microsoft.Health.Common.Errors;
 using Microsoft.Health.Common.Telemetry;
 using Microsoft.Health.Extensions.Fhir;
 using Microsoft.Health.Fhir.Ingest.Data;
 using Microsoft.Health.Fhir.Ingest.Service;
 using Microsoft.Health.Fhir.Ingest.Template;
 using Microsoft.Health.Logging.Telemetry;
+using Newtonsoft.Json.Linq;
 
 namespace Microsoft.Health.Fhir.Ingest.Telemetry
 {
@@ -18,7 +20,9 @@ namespace Microsoft.Health.Fhir.Ingest.Telemetry
     {
         private readonly string _connectorStage = ConnectorOperation.FHIRConversion;
 
-        public FhirExceptionTelemetryProcessor()
+        private IErrorMessageService _errorMessageService;
+
+        public FhirExceptionTelemetryProcessor(IErrorMessageService errorMessageService = null)
             : base (
                 typeof(PatientDeviceMismatchException),
                 typeof(ResourceIdentityNotDefinedException),
@@ -29,14 +33,26 @@ namespace Microsoft.Health.Fhir.Ingest.Telemetry
                 typeof(CorrelationIdNotDefinedException),
                 typeof(InvalidQuantityFhirValueException))
         {
+            _errorMessageService = errorMessageService;
         }
 
-        public override bool HandleException(Exception ex, ITelemetryLogger logger)
+        public override bool HandleException(Exception ex, JToken message, ITelemetryLogger logger)
         {
             EnsureArg.IsNotNull(ex, nameof(ex));
             EnsureArg.IsNotNull(logger, nameof(logger));
 
             var exceptionTypeName = ex.GetType().Name;
+
+            // send to error message service
+            if (_errorMessageService != null)
+            {
+                var errorMessage = new ErrorMessage();
+                errorMessage.InputMessage = message;
+                errorMessage.Details = ex.Message;
+                errorMessage.Type = exceptionTypeName;
+                _errorMessageService.ReportError(errorMessage, default);
+             }
+
             var handledExceptionMetric = ex is NotSupportedException ? IomtMetrics.NotSupported() : IomtMetrics.HandledException(exceptionTypeName, _connectorStage);
             return HandleException(ex, logger, handledExceptionMetric, IomtMetrics.UnhandledException(exceptionTypeName, _connectorStage));
         }
--- a/src/lib/Microsoft.Health.Fhir.Ingest/Telemetry/NormalizationExceptionTelemetryProcessor.cs
+++ b/src/lib/Microsoft.Health.Fhir.Ingest/Telemetry/NormalizationExceptionTelemetryProcessor.cs
@@ -6,9 +6,11 @@
 using System;
 using System.Linq;
 using EnsureThat;
+using Microsoft.Health.Common.Errors;
 using Microsoft.Health.Common.Telemetry;
 using Microsoft.Health.Fhir.Ingest.Template;
 using Microsoft.Health.Logging.Telemetry;
+using Newtonsoft.Json.Linq;
 
 namespace Microsoft.Health.Fhir.Ingest.Telemetry
 {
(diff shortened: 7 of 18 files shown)


src/lib/Microsoft.Health.Fhir.Ingest/Host/DeviceDataNormalizationExtensions.cs
```diff
@@ -46,6 +47,8 @@ public static IServiceCollection AddNormalizationExceptionTelemetryProcessor(thi
             EnsureArg.IsNotNull(serviceCollection, nameof(serviceCollection));
             EnsureArg.IsNotNull(configuration, nameof(configuration));
 
+            serviceCollection.AddSingleton<IErrorMessageService, ErrorMessageService>();
```
namalu, line 50: ErrorMessageService is causing a compile time error for me. It looks like it is not included in this branch.

src/lib/Microsoft.Health.Common/Errors/IErrorMessageService.cs
```diff
@@ -0,0 +1,39 @@
...
+        /// Writes an ErrorMessage to a destination
+        /// </summary>
+        /// <remarks>
+        /// The ErrorMessage Type property will be the ExceptionType.
+        /// and Details property will be the Exception Message.
+        /// </remarks>
+        /// <param name="exception">An Exception.</param>
+        /// <param name="cancellationToken">A cancellation token.</param>
+        /// <returns><see cref="Task"/>Returns the successfully stored error message.</returns>
+        Task ReportError(Exception exception, CancellationToken cancellationToken);
```
namalu, line 25: Any reason why this does not return anything, but the method below returns a collection of ErrorMessage?

src/lib/Microsoft.Health.Fhir.Ingest/Host/DeviceDataNormalizationExtensions.cs
```diff
@@ -56,11 +59,24 @@ private static NormalizationExceptionTelemetryProcessor TelemetryProcessorFactor
         {
             var options = serviceProvider.GetRequiredService<IOptions<NormalizationServiceOptions>>();
 
-            return options.Value.ErrorHandlingPolicy switch
+            var errorMessageService = serviceProvider.GetService<IErrorMessageService>();
+
+            if (errorMessageService == null)
```
namalu, line 62: I'm not sure you need this switch, or even the alternate (zero parameter) constructor for NormalizationExceptionTelemetryProcessor. NormalizationExceptionTelemetryProcessor internally handles the case where errorMessageService is null. I think it makes sense to keep the logic that handles null errorMessageService in the consuming class, and simplify this code to not check if errorMessageService is null.

src/lib/Microsoft.Health.Fhir.Ingest/Host/DeviceDataNormalizationExtensions.cs
wi-y: yep, my bad. I needed to remove this line of code for OSS. Should be fixed now.

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementEventNormalizationService.cs
```diff
@@ -198,7 +198,22 @@ await asyncCollectorConsumer.Completion
...
+                request["Properties"][prop.Key] = prop.Value.ToString();
+            }
+
+            foreach (var prop in data.SystemProperties)
+            {
+                request["SystemProperties"][prop.Key] = prop.Value.ToString();
+            }
+
+            request["Body"] = JObject.Parse(System.Text.Encoding.Default.GetString(data.Body.ToArray()));
+            var handled = _exceptionTelemetryProcessor.HandleException(ex, request, _log);
```
namalu, line 216: It might make sense to move this code to an EventData extension that you could call - e.g. data.ToJObject();

src/lib/Microsoft.Health.Fhir.Ingest/Telemetry/NormalizationExceptionTelemetryProcessor.cs
```diff
@@ -5,30 +5,50 @@
...
-            return HandleException(ex, logger, IomtMetrics.HandledException(exceptionTypeName, _connectorStage), IomtMetrics.UnhandledException(exceptionTypeName, _connectorStage));
+            var shouldContinue = HandleException(ex, logger, IomtMetrics.HandledException(exceptionTypeName, _connectorStage), IomtMetrics.UnhandledException(exceptionTypeName, _connectorStage));
+
+            if (_errorMessageService != null)
+            {
+                var errorMessage = new ErrorMessage();
+                errorMessage.InputMessage = message;
+                errorMessage.Details = ex.Message;
+                errorMessage.Type = exceptionTypeName;
+                _errorMessageService.ReportError(errorMessage, default);
```
namalu, line 48: Are we going to report errors here regardless of whether or not the exception is handled? I'm just wondering if that might cause confusion, especially if there is no action that is required by the customer.

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementFhirImportService.cs
```diff
@@ -111,23 +118,28 @@ private static IEnumerable<IMeasurementGroup> ParseEventData(IEnumerable<IEventM
             var partitionId = data.FirstOrDefault()?.PartitionId;
 
             // Deserialize events into measurements and then group according to the device, type, and other factors
-            return data.Select(e => JsonConvert.DeserializeObject<Measurement>(System.Text.Encoding.Default.GetString(e.Body.ToArray())))
+            return data.Select(e =>
+                {
+                    var measurement = JsonConvert.DeserializeObject<Measurement>(System.Text.Encoding.Default.GetString(e.Body.ToArray()));
+                    measurement.Payload = e.Body;
+                    return measurement;
```
namalu, line 125: It's hard to tell easily, but are we not going to provide the customer with the original message (device payload) when errors occur in the Measurement to FHIR transformation?

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementFhirImportService.cs
```diff
@@ -73,7 +73,14 @@ private async Task ProcessMeasurementGroups(IEnumerable<IMeasurementGroup> measu
                             }
                             catch (Exception ex)
                             {
-                                if (!Options.ExceptionService.HandleException(ex, log))
+                                // capture the message payload for error message service
+                                JArray payload = new JArray();
+                                foreach (var measurement in m.Data)
+                                {
+                                    payload.Add(JObject.Parse(System.Text.Encoding.Default.GetString(measurement.Payload.ToArray())));
```
namalu, line 80: I notice that we are doing quite a bit of converting data in several places in the code. It makes me wonder if you could just provide multiple overloads for HandleException (one that takes type EventData and one that takes type IEnumerable<IMeasurementGroup>). You could just pass the event or measurements to the HandleException method and either do the conversions in the method or add multiple constructors to ErrorMessage and handle all conversions in one class. It would definitely make it easier to track down where data is being converted.

namalu (changes requested):

src/lib/Microsoft.Health.Fhir.Ingest/Telemetry/FhirExceptionTelemetryProcessor.cs
```diff
@@ -29,14 +33,26 @@ public FhirExceptionTelemetryProcessor()
...
         {
             EnsureArg.IsNotNull(ex, nameof(ex));
             EnsureArg.IsNotNull(logger, nameof(logger));
 
             var exceptionTypeName = ex.GetType().Name;
+
+            // send to error message service
+            if (_errorMessageService != null)
+            {
+                var errorMessage = new ErrorMessage();
```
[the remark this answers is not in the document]
dustinburson: The processor is defined once, the message is supposed to be per exception so don't think the JToken in the ctor is appropriate. 

Overall I think we might want to refactor this to introduce a context that would wrap the Exception, a reference to what is being processed, and any other data that is relevant.  This context can then be used to generate the error message.

This doesn't need to be available in the Azure Function OSS version. That may allow you more freedom in the implementation.

src/lib/Microsoft.Health.Fhir.Ingest/Telemetry/NormalizationExceptionTelemetryProcessor.cs
dustinburson: Well examples of handled exceptions would be invalid patient id, duplicate patient id, bad data in the device message, etc.  Handled just means we don't block the pipeline and transform it to telemetry.  Most non-handled exceptions are actually ones that we wouldn't necessary emit error messages for (cannot connect to the Event Hub, incorrect permissions, invalid FHIR server).  These are the ones we retry.

dustinburson (changes requested):

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementFhirImportService.cs
wi-y: Right now for FHIR Transformation we will not provide original device message, it is the measurement group payload (though it may be updated to be the individual measurement payload). I would like to provide the original device message since it would be really useful, but the challenge is that I do not think we can guarantee the original device message + normalized message will fit in the normalized event hub given the 1MB limit. Even if we use compression the compression ratio varies depending on the payload content and there could be cases where we would exceed 1MB.

One option could be to just throw an error in the case above and write the original device message to a dead letter queue. One implication though is that we'd be telling users that the message size limit is now (size of input message + size of normalized message) < ~1MB which is less that current state and it may impact users that are close to 1MB already.

Another option could be to not include the original device message if including it would exceed the limit but that could be a inconsistent experience.

The measurement/measurement group should be enough to troubleshoot, but on the other hand there would be nothing to resubmit... so maybe one of the other options is preferable?

src/lib/Microsoft.Health.Fhir.Ingest/Telemetry/NormalizationExceptionTelemetryProcessor.cs
wi-y: Yeah I was on the fence about this. We should probably only report errors that are handled like patient id not found, duplicate patient id, etc, so I can update the code to only emit error messages for "handled" exceptions. 

I do think there is some value for alerting the customer on non-handled errors like (cannot connect to Event Hub, invalid FHIR server, etc if they are not transient) and I do think there is some value for having one location that contains all the errors, but it can be solved another way like logs or portal warnings.

For this project it is probably best to have actionable dead letter errors only.

test/Microsoft.Health.Fhir.Ingest.UnitTests/Telemetry/FhirExceptionTelemetryProcessorTests.cs
```diff
@@ -30,7 +31,7 @@ public void GivenHandledExceptionTypes_WhenHandleExpection_ThenMetricLoggedAndTr
             var ex = Activator.CreateInstance(exType) as Exception;
 
             var exProcessor = new FhirExceptionTelemetryProcessor();
-            var handled = exProcessor.HandleException(ex, log);
+            var handled = exProcessor.HandleException(ex, new JArray(), log);
```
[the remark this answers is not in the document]
wi-y: For the Fhir Transformation was trying to include the measurement group as the payload logged with the error which I believe is an array. If we do a bit more work to get the individual measurement then I believe JObject would be more appropriate yes. Might end up just using a byte array though like you suggested to avoid the conversion

src/lib/Microsoft.Health.Fhir.Ingest/Telemetry/FhirExceptionTelemetryProcessor.cs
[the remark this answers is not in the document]
wi-y: Thanks Nate. Having the mapping in the ErrorMessage class or internally in another class makes sense to me.

@dustinburson we could have an ErrorContext class that wraps an exception and a reference to what is processed with a "ToErrorMessage" method. Is the advantage of that so that we have a strongly typed class that includes exception + payload that we pass around instead of independently passing an exception and a reference to the payload as separate arguments?

src/lib/Microsoft.Health.Fhir.Ingest/Telemetry/NormalizationExceptionTelemetryProcessor.cs
namalu: It makes me wonder if error reporting is enabled, the term 'handled' is obsolete. If an exception is encountered, do we always just report the error and move on?

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementFhirImportService.cs
namalu: I think having the original message is very important. The measurement can always be recreated using the codepath that created the measurement in the first place - The customer can do this with our mapping tools, or we can generate the a new measurement after the fact. This is not true for the device message, if we don't include it, it's gone. There are quite a few error scenarios (unexpected values, missing values, etc.) where having the device message would be the only way to figure out what is happening.

src/lib/Microsoft.Health.Common/Errors/IErrorMessageService.cs
wi-y: the error does not need to be returned, my mistake. updated this interface to not return the error

src/lib/Microsoft.Health.Fhir.Ingest/Host/DeviceDataNormalizationExtensions.cs
wi-y: the previous code as-is required the error message service to be passed in because we were new'ing the class, but I've refactored things a bit so we can use dependency injection and not "new" the class, so the null check is removed now

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementEventNormalizationService.cs
wi-y: removed the conversion here. we now just pass the event object to the error message service without converting it to json

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementFhirImportService.cs
wi-y: removed the conversion here. now we just pass the event object to the error message service. we can convert it there if needed depending on what destination we choose

src/lib/Microsoft.Health.Common/Errors/IErrorMessageService.cs
```diff
@@ -0,0 +1,39 @@
...
+        /// Writes an IEnumerable of ErrorMessage to a destination.
+        /// </summary>
+        /// <remarks>
+        /// The ErrorMessage Type property will be the ExceptionType.
+        /// and Details property will be the Exception Message.
+        /// </remarks>
+        /// <param name="exceptions">A collection of exceptions.</param>
+        /// <param name="cancellationToken">A cancellation token.</param>
+        /// <returns><see cref="Task"/>Returns the successfully stored error message.</returns>
+        Task<IEnumerable<ErrorMessage>> ReportError(IEnumerable<Exception> exceptions, CancellationToken cancellationToken);
```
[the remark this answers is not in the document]
wi-y: I don't think we need to return the ErrorMessage, so the code/interface is now updated to not return them.

src/lib/Microsoft.Health.Fhir.Ingest/Config/MeasurementFhirImportOptions.cs
```diff
@@ -12,10 +13,24 @@ namespace Microsoft.Health.Fhir.Ingest.Config
...
+            TemplateFactory = CollectionFhirTemplateFactory.Default;
+            ExceptionService = new FhirExceptionTelemetryProcessor();
+        }
 
-        public virtual IExceptionTelemetryProcessor ExceptionService { get; } = new FhirExceptionTelemetryProcessor();
+        public MeasurementFhirImportOptions(IErrorMessageService errorMessageService)
+        {
+            ParallelTaskOptions = new ParallelTaskOptions { MaxConcurrency = 10 };
+            TemplateFactory = CollectionFhirTemplateFactory.Default;
+            ExceptionService = new FhirExceptionTelemetryProcessor(errorMessageService);
```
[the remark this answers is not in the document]
wi-y: yeah, we don't need to include the processor in the options - however that it was the previous code had. Went ahead and tried to refactor it so now we do not include the processor in the options and instead use DI.

src/lib/Microsoft.Health.Common/Errors/IErrorMessageService.cs
```diff
@@ -0,0 +1,39 @@
...
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Health.Common.Errors
+{
+    public interface IErrorMessageService
```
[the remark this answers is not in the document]
namalu: The return type needs to be changed to void for these not to be async. It looks like they are still Tasks.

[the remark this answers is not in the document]
wi-y: thanks - updated

src/lib/Microsoft.Health.Fhir.Ingest/Telemetry/NormalizationExceptionTelemetryProcessor.cs
namalu: I'm still confused about what we are doing here. I thought one of the main reasons we are adding the Error Message Service is to eliminate the possibility of the pipeline being blocked by a poison message. The current implementation will still allow that to happen, but flood the Error Message Service with the same error every time it happens.

In my mind there is no such thing as a "handled" exception anymore when Error Message Service is enabled; either a message is successfully processed or it isn't. If a message isn't successfully processed, it should result in an error message. The problem with labeling an error as "transient" is we don't know that an issue was transient until we look back. If an error could be transient, I would propose that we retry a fixed number of times but not indefinitely. **The indefinite retry is where we run into the possibility of the pipeline being blocked and in an unrecoverable state, which will result in data loss.**

wi-y: I think the first step is exposing the error to users if it is actionable to them, and in parallel we should be updating the code to throw appropriate exceptions so that we know when to retry and when not to retry, and when to expose the error to the error message service. We currently assume that if the exception is "unknown" or "unhandled" then we should retry forever. One partial solution could be to update our code - for example we recently added a "NormalizationDataMappingException" which we can now use to determine if we can immediately log an error and move on. Previously we retried forever and this was a "poisoned message" scenario but now it is not.

Another solution for "unknown" or "unhandled" exceptions is, liked you mentioned, to retry X number of times and then log an error. That can work in a number of cases especially if the error is with the message itself. One case where it may be ambiguous is with connectivity errors like HTTP 500 or a managed identity related issue. Blips happen and I think we should retry forever in this scenario and attempt to fix the underlying connectivity issue, rather than retrying X number of times and logging an error, since the user can't really fix the error. On the other hand, if the connectivity issue is not resolved in a timely manner and the data fades out then it could result in data loss.

imo we should update the various places in our code to be able to detect a poisoned message and throw an exception like in the case of the NormalizationDataMappingException. And in the case where the exception is still not known then we can retry X times and log an error, but need to take into consideration connectivity related errors and handle those appropriately

namalu: I think it's fine retrying connectivity issues (connectivity to our internal resources only) for a specific time period (maybe 15 minutes), and then sending an error to the Error Message Service so the potential of data loss is reduced, but are there other scenarios that would warrant retrying? I can't think of anything that isn't under the customer's control.

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementFhirImportService.cs
wi-y: based on a previous conversation we are going to focus on normalization for now and include the device message for normalization errors. we are going to leave it as a future item on how to best include the device message for measurementtofhir.

namalu: Are there any user stories/features tracking this future work?

namalu:
It looks like there is a case in [Normalize Processor.cs (line 65)](https://github.com/microsoft/iomt-fhir/blob/b34ff703915384b158e8152456835c249f167449/src/console/Normalize/Processor.cs#L65), where an exception for an invalid template can be thrown that is not handled by the Error Message Service.

wi-y:
> It looks like there is a case in [Normalize Processor.cs (line 65)](https://github.com/microsoft/iomt-fhir/blob/b34ff703915384b158e8152456835c249f167449/src/console/Normalize/Processor.cs#L65), where an exception for an invalid template can be thrown that is not handled by the Error Message Service.

Yes it throws a ValidationException. I believe I have accounted for this by including ValidationException in NormalizationExceptionTelemetryProcessor.cs when the Error Message Service is enabled.

The Normalize Processor is no longer used since we recently replaced it with NormalizationEventConsumerService, so I did not update the Normalize Processor. I can add the same logic to the Normalize Processor if we want to, but we should probably just delete it.

namalu:
> 

Yes. We should just delete this if it is not being used anymore.

wi-y:
> > 
> 
> Yes. We should just delete this if it is not being used anymore.

done

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementFhirImportService.cs
```diff
@@ -41,10 +44,59 @@ public async Task ProcessStreamAsync(Stream data, string templateDefinition, ITe
...
+                }
+                else
+                {
+                    Console.WriteLine(events.Count());
+
+                    // remove the event(s) from the batch
+                    var eventsWithExceptions = ex.GetRelatedLegacyEvents();
+                    events = events.Where(e => !eventsWithExceptions.Contains(e)).ToList();
+
+                    Console.WriteLine(events.Count());
```
namalu, line 93: It doesn't look like this else statement is needed. Should we remove it?

namalu (approved):
I added a few more comments. I think the most important thing is my comment about how Exceptions are serialized in our logs, and if the Event Data can leak PHI/PII into them.

src/lib/Microsoft.Health.Fhir.Ingest/Service/NormalizationEventConsumerService.cs
```diff
@@ -104,19 +119,59 @@ private async Task<IContentTemplate> GetNormalizationTemplate()
...
-
+            // Step 1: Get the normalization mapping template
+            IContentTemplate template = null;
+            try
+            {
+                template = await GetNormalizationTemplate();
+            }
+            catch (Exception ex)
+            {
+                ex.AddEventContext(events);
```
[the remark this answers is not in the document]
wi-y: currently the exception.Data does not appear in the logs, but for PaaS will still probably add some guardrails to remove the exception.Data in case the current behavior changes.

src/lib/Microsoft.Health.Fhir.Ingest/Telemetry/NormalizationExceptionTelemetryProcessor.cs
```diff
@@ -4,31 +4,57 @@
...
         public override bool HandleException(Exception ex, ITelemetryLogger logger)
         {
             EnsureArg.IsNotNull(ex, nameof(ex));
             EnsureArg.IsNotNull(logger, nameof(logger));
 
             var exceptionTypeName = ex.GetType().Name;
-            return HandleException(ex, logger, IomtMetrics.HandledException(exceptionTypeName, _connectorStage), IomtMetrics.UnhandledException(exceptionTypeName, _connectorStage));
+            var shouldContinue = HandleException(ex, logger, IomtMetrics.HandledException(exceptionTypeName, _connectorStage), IomtMetrics.UnhandledException(exceptionTypeName, _connectorStage));
+
+            if (_errorMessageService != null)
```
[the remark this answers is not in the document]
wi-y: yep, good catch

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementFhirImportService.cs
wi-y: yep, it should be removed. was being used to debug and forgot to remove it

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementFhirImportService.cs
```diff
@@ -41,10 +44,59 @@ public async Task ProcessStreamAsync(Stream data, string templateDefinition, ITe
...
+
+            // Step 2: Transform the events into Measurements and group into Measurement Groups
+            Tuple<IEnumerable<IMeasurementGroup>, Dictionary<IMeasurement, IEventMessage>> parseEventData;
+            IEnumerable<IMeasurementGroup> measurementGroups = null;
+            Dictionary<IMeasurement, IEventMessage> measurementToEventMapping = null;
+
+            try
+            {
+                parseEventData = ParseEventData(events, log);
+                measurementGroups = parseEventData.Item1;
```
dustinburson, line 76: Please use named tuples for more readable code, i.e. you can do something like ```Tuple<IEnumerable<IMeasurementGroup> groups, Dictionary<IMeasurement, IEventMessage> lookup> parseEventData```

and then do
```
measurementGroups = parseEventData.groups;
measurementToEventMapping = parseEventData.lookup;
```

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementFhirImportService.cs
```diff
@@ -41,10 +44,59 @@ public async Task ProcessStreamAsync(Stream data, string templateDefinition, ITe
...
+            }
+            catch (Exception ex)
+            {
+                if (!_exceptionTelemetryProcessor.HandleException(ex, log))
+                {
+                    throw;
+                }
+                else
+                {
+                    Console.WriteLine(events.Count());
```
dustinburson, line 87: Can we replace console writeline with Traces or remove?  We should avoid Count() or other repeated calls that require a full enumeration of the collect if it can be avoided.

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementFhirImportService.cs
```diff
@@ -106,28 +171,43 @@ private static async Task<IEnumerable<IMeasurementGroup>> ParseAsync(Stream data
...
-                        Data = measurements,
-                        MeasureType = measurements[0].Type,
-                        CorrelationId = measurements[0].CorrelationId,
-                        DeviceId = measurements[0].DeviceId,
-                        EncounterId = measurements[0].EncounterId,
-                        PatientId = measurements[0].PatientId,
-                    };
-                })
-                .ToArray();
+                    var measurement = JsonConvert.DeserializeObject<Measurement>(System.Text.Encoding.Default.GetString(e.Body.ToArray()));
```
dustinburson, line 184: Suggestion:  I realize this old code, but rather than convert to an array, it would be better to load from a memory stream to avoid excessive memory allocation. See EventMessageJObjectConverter for an example.

src/lib/Microsoft.Health.Fhir.Ingest/Service/NormalizationEventConsumerService.cs
```diff
@@ -125,21 +180,11 @@ private async Task ConsumeAsyncImpl(IEnumerable<IEventMessage> events)
...
-                NormalizeMessage(token: token, evt: evt, template: template, collector: collector);
-            }
-            catch (Exception ex)
-            {
-                if (!_exceptionTelemetryProcessor.HandleException(ex, _logger))
-                {
-                    throw; // Immediately throw original exception if it is not handled
-                }
-            }
+            NormalizeMessage(token: token, evt: evt, template: template, collector: collector);
```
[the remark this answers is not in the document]
dustinburson: It is to have a named tuple and avoid needed to reference Item1, Item2, etc.

src/lib/Microsoft.Health.Fhir.Ingest/Service/NormalizationEventConsumerService.cs
```diff
@@ -158,7 +203,9 @@ private void NormalizeMessage(JObject token, IEventMessage evt, IContentTemplate
                 catch (Exception ex)
                 {
                     // Translate all Normalization Mapping exceptions into a common type for easy identification.
-                    throw new NormalizationDataMappingException(ex);
+                    var mappingException = new NormalizationDataMappingException(ex);
+                    mappingException.AddEventContext(evt);
```
dustinburson, line 207: Suggestion: You can make AddEventContext return the exception allowing you to chain statements fluent style, i.e. 
```
throw new NormalizationDataMappingException(ex)
    .AddEventContext(evt);
```

dustinburson (approved):

src/lib/Microsoft.Health.Events/Errors/ErrorMessage.cs
```diff
@@ -0,0 +1,81 @@
...
+        public DateTimeOffset ErrorTimestamp { get; set; }
+
+        /// <summary>
+        /// The message payload(s) that led to an error to be thrown.
+        /// </summary>
+        /// <remarks>
+        /// This property provides the input message payload so the user can determine
+        /// what message was correlated with the error.
+        /// </remarks>
+        public IEnumerable<IEventMessage> RelatedLegacyEvents { get; set; }
```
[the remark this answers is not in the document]
wi-y: yes but decided it was too much to include for this pull request

src/lib/Microsoft.Health.Events/Errors/ErrorMessage.cs
```diff
@@ -0,0 +1,81 @@
...
+
+using System;
+using System.Collections.Generic;
+using EnsureThat;
+using Microsoft.Azure.EventHubs;
+using Microsoft.Health.Events.Model;
+
+namespace Microsoft.Health.Events.Errors
+{
+    public class ErrorMessage
```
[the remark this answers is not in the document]
wi-y: yeah it was to do both of those things. Created separate interfaces for the error message context and the data contract for the error service

src/console/Normalize/Processor.cs
```diff
@@ -1,131 +0,0 @@
-﻿// -------------------------------------------------------------------------------------------------
-// Copyright (c) Microsoft Corporation. All rights reserved.
-// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
-// -------------------------------------------------------------------------------------------------
-
-using System;
```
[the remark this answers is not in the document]
wi-y: yep

src/console/Startup.cs
```diff
@@ -81,14 +83,16 @@ public virtual List<IEventConsumer> ResolveEventConsumers(IServiceProvider servi
                 var collector = ResolveEventCollector(serviceProvider);
                 var collectionContentFactory = serviceProvider.GetRequiredService<CollectionTemplateFactory<IContentTemplate, IContentTemplate>>();
                 var exceptionTelemetryProcessor = serviceProvider.GetRequiredService<NormalizationExceptionTelemetryProcessor>();
-                var deviceDataNormalization = new NormalizationEventConsumerService(new EventMessageJObjectConverter(), template, templateManager, collector, logger, collectionContentFactory, exceptionTelemetryProcessor);
+                var errorMessageService = serviceProvider.GetService<IErrorMessageService>();
```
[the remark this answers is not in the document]
wi-y: yes to both

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementFhirImportService.cs
wi-y: updated

PR #215, opened 2022-07-13
Handle exceptions related to invalid event message format.

pallar-ms:
The normalization processor expects the input event message to be in a format that can be parsed into a JSON object. However, the event data sent from the associated source event hub needn't be in this format. This could result in a JsonReaderException being thrown when attempting to convert the payload. 
This change includes:
* Adding a custom exception with a specific error message to provide more clarity to customers regarding the issue, and is also exposed via an error metric (and logs in the near future).
* This is also being added as a "handled" exception, so we don't stall on event data of unsupported formats.
* Minor reorganization of exception related classes into a separate folder.


src/lib/Microsoft.Health.Fhir.Ingest/Telemetry/NormalizationExceptionTelemetryProcessor.cs
```diff
@@ -7,15 +7,20 @@
...
 namespace Microsoft.Health.Fhir.Ingest.Telemetry
 {
     public class NormalizationExceptionTelemetryProcessor : ExceptionTelemetryProcessor
     {
         private readonly string _connectorStage = ConnectorOperation.Normalization;
-        private static readonly Type[] DefaultExceptions = new[] { typeof(IncompatibleDataException) };
+        private static readonly Type[] DefaultExceptions = new[]
+        {
+            typeof(IncompatibleDataException),
+            typeof(InvalidDataFormatException),
```
pallar-ms, line 22: Adding it as a handled exception so we don't stall on events that are on unsupported formats, but let me know if this is not desired.

namalu (approved):

wi-y (approved):

dustinburson (approved):

PR #217, opened 2022-07-18
Add ability to easily track call duration and emit to a metric

dustinburson:
Added ITimed with extension support to track call durations and emit to metrics.

### DIFF
--- a/src/lib/Microsoft.Health.Fhir.Ingest/Service/NormalizationEventConsumerService.cs
+++ b/src/lib/Microsoft.Health.Fhir.Ingest/Service/NormalizationEventConsumerService.cs
@@ -5,7 +5,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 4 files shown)


wi-y (approved):

src/lib/Microsoft.Health.Fhir.Ingest/Service/NormalizationEventConsumerService.cs
```diff
@@ -189,30 +188,34 @@ private void ProcessEvent(IEventMessage evt, IContentTemplate template, IList<(s
...
-                    measurement.IngestionTimeUtc = evt.EnqueuedTime.UtcDateTime;
-                    collector.Add((evt.PartitionId, measurement));
-                    projections++;
-                }
-                catch (Exception ex)
-                {
-                    // Translate all Normalization Mapping exceptions into a common type for easy identification.
-                    throw new NormalizationDataMappingException(ex, nameof(NormalizationDataMappingException))
-                        .AddEventContext(evt);
+                    try
```
rogordon01, line 197: This will record the time taken to normalize the device event. But  are we still capturing the time it takes to produce an individual measurement?
 
Also, is `NormalizationEventConsumerService` replacing `MeasurementEventNormalizationService`?

rogordon01 (approved):

PR #218, opened 2022-07-18
Ability to flag errors as customer facing

artunduman:
This PR:
- Adds an exception extension to flag any arbitrary error as customer facing
- Adds a base class for structured exceptions to extend if all of their instances should be logged to customers
- Adds a dimension to exception logs based on the above

I followed the extension approach to be flexible enough to log arbitrary exceptions, but added the base class to not have to switch on the type of the error at the point of logging. Exception.LogToCustomer() should be the source of truth for whether the exception will be logged to customer or not.

### DIFF
--- a/src/lib/Microsoft.Health.Common/Extension/ExceptionExtensions.cs
+++ b/src/lib/Microsoft.Health.Common/Extension/ExceptionExtensions.cs
@@ -9,16 +9,16 @@ namespace Microsoft.Health.Common.Extension
 {
     public static class ExceptionExtensions
     {
-        private const string LogToCustomerAttribute = "LogToCustomer";
+        private const string LogForwardingAttribute = "ShouldForwardLog";
 
         /// <summary>
         /// Flags the exception as being logged to the customer or not.
         /// </summary>
         /// <param name="error">The root exception</param>
-        /// <param name="logToCustomer">Boolean to set whether the exception should be logged to third-party customers</param>
-        public static Exception SetLogForwarding(this Exception error, bool logToCustomer)
+        /// <param name="shouldForwardLog">Boolean to set whether the exception should be logged to third-party</param>
+        public static Exception SetLogForwarding(this Exception error, bool shouldForwardLog)
         {
-            error.Data[LogToCustomerAttribute] = logToCustomer;
+            error.Data[LogForwardingAttribute] = shouldForwardLog;
             return error;
         }
 
@@ -26,10 +26,10 @@ namespace Microsoft.Health.Common.Extension
         /// Returns true iff the exception is flagged as being logged to the customer.
         /// </summary>
         /// <param name="error">The root exception</param>
-        /// <returns>True iff the exception should be logged to third-party customers</returns>
+        /// <returns>True iff the exception should be logged to third-party</returns>
         public static bool IsLogForwardingEnabled(this Exception error)
         {
-            return error.Data[LogToCustomerAttribute] as bool? ?? false;
+            return error.Data[LogForwardingAttribute] as bool? ?? false;
         }
     }
 }
--- a/src/lib/Microsoft.Health.Common/Extension/ExceptionExtensions.cs
+++ b/src/lib/Microsoft.Health.Common/Extension/ExceptionExtensions.cs
@@ -11,11 +11,21 @@ namespace Microsoft.Health.Common.Extension
     {
         private const string LogToCustomerAttribute = "LogToCustomer";
 
+        /// <summary>
+        /// Flags the exception as being logged to the customer or not.
+        /// </summary>
+        /// <param name="error">The root exception</param>
+        /// <param name="logToCustomer">Boolean to set whether the exception should be logger to third-party customers</param>
         public static void SetLogToCustomer(this Exception error, bool logToCustomer)
         {
             error.Data[LogToCustomerAttribute] = logToCustomer;
         }
 
+        /// <summary>
+        /// Returns true iff the exception is flagged as being logged to the customer.
+        /// </summary>
+        /// <param name="error">The root exception</param>
+        /// <returns>True iff the exception should be logged to third-party customers</returns>
         public static bool LogToCustomer(this Exception error)
         {
             return error.Data[LogToCustomerAttribute] as bool? ?? false;
--- a/src/lib/Microsoft.Health.Common/Extension/ExceptionExtensions.cs
+++ b/src/lib/Microsoft.Health.Common/Extension/ExceptionExtensions.cs
@@ -16,9 +16,10 @@ namespace Microsoft.Health.Common.Extension
         /// </summary>
         /// <param name="error">The root exception</param>
         /// <param name="logToCustomer">Boolean to set whether the exception should be logged to third-party customers</param>
-        public static void SetShouldLogToCustomer(this Exception error, bool logToCustomer)
+        public static Exception SetLogForwarding(this Exception error, bool logToCustomer)
         {
             error.Data[LogToCustomerAttribute] = logToCustomer;
+            return error;
         }
 
         /// <summary>
@@ -26,7 +27,7 @@ namespace Microsoft.Health.Common.Extension
         /// </summary>
         /// <param name="error">The root exception</param>
         /// <returns>True iff the exception should be logged to third-party customers</returns>
-        public static bool ShouldLogToCustomer(this Exception error)
+        public static bool IsLogForwardingEnabled(this Exception error)
         {
             return error.Data[LogToCustomerAttribute] as bool? ?? false;
         }
--- a/src/lib/Microsoft.Health.Common/Extension/ExceptionExtensions.cs
+++ b/src/lib/Microsoft.Health.Common/Extension/ExceptionExtensions.cs
@@ -1,0 +1,24 @@
+// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+
+namespace Microsoft.Health.Common.Extension
+{
+    public static class ExceptionExtensions
+    {
+        private const string LogToCustomerAttribute = "LogToCustomer";
+
+        public static void SetLogToCustomer(this Exception error, bool logToCustomer)
+        {
+            error.Data[LogToCustomerAttribute] = logToCustomer;
+        }
+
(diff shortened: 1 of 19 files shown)


src/lib/Microsoft.Health.Fhir.Ingest.Template/InvalidTemplateException.cs
```diff
@@ -39,5 +41,9 @@ public override string Message
         public ILineInfo GetLineInfo => _lineInfo;
 
         public bool HasLineInfo => _lineInfo.HasLineInfo();
+
+        public override string ErrSource => nameof(ErrorSource.User);
+
+        public override string ErrSeverity => ErrorSeverity.Critical;
```
wi-y, line 47: is the error severity exposed to the customer?

wondering if it would be useful to have Critical and Warning for errors as well as "Information" traces in the future.

src/lib/Microsoft.Health.Events/Telemetry/Exceptions/UnauthorizedAccessEventHubException.cs
```diff
@@ -9,7 +9,7 @@
 
 namespace Microsoft.Health.Events.Telemetry.Exceptions
 {
-    public sealed class UnauthorizedAccessEventHubException : IomtTelemetryFormattableException
+    public sealed class UnauthorizedAccessEventHubException : CustomerLoggedFormattableException
```
kyclai, line 12: Some more classes to derive from `CustomerLoggedFormattableException`:
- [InvalidEventHubExceptions](https://github.com/microsoft/iomt-fhir/blob/main/src/lib/Microsoft.Health.Events/Telemetry/Exceptions/InvalidEventHubException.cs)
[InvalidFhirServiceException](https://github.com/microsoft/iomt-fhir/blob/main/src/lib/Microsoft.Health.Extensions.Fhir.R4/Telemetry/Exceptions/InvalidFhirServiceException.cs)
- [UnauthorizedAccessFhirServiceException](https://github.com/microsoft/iomt-fhir/blob/main/src/lib/Microsoft.Health.Extensions.Fhir.R4/Telemetry/Exceptions/UnauthorizedAccessFhirServiceException.cs)

src/lib/Microsoft.Health.Fhir.Ingest.Template/InvalidTemplateException.cs
artunduman: Currently, it's not. I think I can expose this and make a default severity level for unstructured exceptions. Do you think I should add this to the current PR or should it be a part of the next iteration?

src/lib/Microsoft.Health.Logger/Telemetry/TelemetryExtensions.cs
```diff
@@ -8,12 +8,16 @@
...
+using Microsoft.Health.Common.Extension;
 
 namespace Microsoft.Health.Logging.Telemetry
 {
     public static class TelemetryExtensions
     {
         private static readonly string _namespace = MetricIdentifier.DefaultMetricNamespace;
+        private static readonly string _helplinkAttribute = "helpLink";
+        private static readonly string _messageAttribute = "message";
+        private static readonly string _logToCustomerAttribute = "logToCustomerIot";
```
pallar-ms, line 20: Just curious why we need to add the IoT suffix? The column could be used by other services too?

artunduman: Yes, we are sharing the input source. 
An alternative is to use both this attribute and the resource type to filter, but that would complicate the mds agent's query and I don't know where exactly that query is processed or if it can overwhelm the agent.

rogordon01 (approved):
Overall it looks good to me. There are some open questions from others so if you can address those before you commit that would be great.

src/lib/Microsoft.Health.Common/Extension/ExceptionExtensions.cs
```diff
@@ -0,0 +1,34 @@
...
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+
+namespace Microsoft.Health.Common.Extension
+{
+    public static class ExceptionExtensions
+    {
+        private const string LogToCustomerAttribute = "LogToCustomer";
```
dustinburson, line 12: Per our conversation yesterday, the concern is the concept of customer is very PaaS centric and in the OSS it doesn't have  a direct application.  

What do you think of naming this something like FowardLog or LogFowarding to make it a little more generic and OSS appropriate?

src/lib/Microsoft.Health.Logger/Telemetry/TelemetryExtensions.cs
dustinburson, line 20: Why a static string instead of a constant?

dustinburson: Are the properties visible in the logs on the customer's subscription or only used internally for filtering?

src/lib/Microsoft.Health.Common/Extension/ExceptionExtensions.cs
dustinburson: Another option is we look at moving this into PaaS but update OSS with an exception factory concept and have a default implementation.  In PaaS we can inject another one, and as part of the factories logic it would set this extension when creating them.

src/lib/Microsoft.Health.Common/Extension/ExceptionExtensions.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        /// <summary>
+        /// Flags the exception as being logged to the customer or not.
+        /// </summary>
+        /// <param name="error">The root exception</param>
+        /// <param name="logToCustomer">Boolean to set whether the exception should be logged to third-party customers</param>
+        public static void SetShouldLogToCustomer(this Exception error, bool logToCustomer)
```
dustinburson, line 19: Suggestion: You can update this to return Exception and allow you to use the extension in a fluent syntax.  I.E.
```
new Exception("Error")
.SetShouldLogToCustomer()
.OtherMethodHere();
```

src/lib/Microsoft.Health.Fhir.Ingest.Template/InvalidTemplateException.cs
```diff
@@ -39,5 +41,9 @@ public override string Message
         public ILineInfo GetLineInfo => _lineInfo;
 
         public bool HasLineInfo => _lineInfo.HasLineInfo();
+
+        public override string ErrSource => nameof(ErrorSource.User);
```
dustinburson, line 45: How is this being used?  It seems like a one off.  I would think if it is used and important I should see this being set in more exceptions beyond this one.

src/lib/Microsoft.Health.Fhir.Ingest.Template/InvalidTemplateException.cs
dustinburson: I believe it is exposed on the error metrics we emit today.

dustinburson (changes requested):

src/lib/Microsoft.Health.Logger/Telemetry/TelemetryExtensions.cs
artunduman: They are used internally, the visible fields are specified in the query in the paas PR

src/lib/Microsoft.Health.Common/Extension/ExceptionExtensions.cs
artunduman: I am a bit confused on how we would follow the factory approach. What do you think the API for tagging arbitrary logs would be?

dustinburson: Sure, I will try will look at adding an example

src/lib/Microsoft.Health.Common/Telemetry/Exceptions/CustomerLoggedFormattableException.cs
```diff
@@ -0,0 +1,40 @@
...
+        }
+
+        public CustomerLoggedFormattableException(string message, Exception innerException)
+            : base(message, innerException)
+        {
+            this.SetLogToCustomer(true);
+        }
+
+        public CustomerLoggedFormattableException(
+            string message,
```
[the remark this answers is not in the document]
artunduman: Sorry for getting back late. Apparently, what the customer sees cannot be dynamic (you can't unpack a property bag dynamically). It can contain dynamic columns, but log analytics recommends against it. 
I'm not sure if I'm supposed to post internal documentation here, but I can send you the reference

src/lib/Microsoft.Health.Logger/Telemetry/TelemetryExtensions.cs
```diff
@@ -8,12 +8,16 @@
...
+using Microsoft.Health.Common.Extension;
 
 namespace Microsoft.Health.Logging.Telemetry
 {
     public static class TelemetryExtensions
     {
         private static readonly string _namespace = MetricIdentifier.DefaultMetricNamespace;
+        private const string _helplinkAttribute = "helpLink";
+        private const string _messageAttribute = "message";
+        private const string _LogForwardingAttribute = "logToCustomerIot";
```
dustinburson, line 20: nit: casing inconsistency.  We should have style cop rules that will enforce this but I believe private constants the style should be camel case.

test/Microsoft.Health.Common.UnitTests/ExceptionExtensionsTest.cs
```diff
@@ -0,0 +1,31 @@
...
+    {
+        [Fact]
+        public void ExceptionNotLoggedByDefault()
+        {
+            var exception = new Exception();
+            Assert.False(exception.IsLogForwardingEnabled());
+        }
+
+        [Fact]
+        public void ExceptionSetLogForwarding()
```
dustinburson, line 22: Suggestion:  It may be good to add explicit tests for the types of exceptions that you expect to be forward automatically as well.

dustinburson (approved):
Changes look good.  Please correct the one code style issue before committing.  Thanks!

src/lib/Microsoft.Health.Logger/Telemetry/TelemetryExtensions.cs
artunduman: Thanks for catching this!

PR #219, opened 2022-07-21
Update solution nugets to latest where possible

dustinburson:
- Update to latest Nuget package versions where applicable.
- Address after package upgrade: 'TelemetryConfiguration.InstrumentationKey.set' is obsolete: 'InstrumentationKey based global ingestion is being deprecated. Use TelemetryConfiguration.ConnectionString. See https://github.com/microsoft/ApplicationInsights-dotnet/issues/2560
- Update unit test for new packages

### DIFF
--- a/Microsoft.Health.Fhir.Ingest.sln
+++ b/Microsoft.Health.Fhir.Ingest.sln
@@ -1,9 +1,9 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.29009.5
+# Visual Studio Version 17
+VisualStudioVersion = 17.2.32630.192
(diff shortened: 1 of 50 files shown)


rogordon01 (approved):

rogordon01:
FYI: I built in image for Normalization after updating to use the latest EventHubs SDK. I received the below error when starting the image. Could just be me, but we may have to update our Docker images to get around this.


```
It was not possible to find any compatible framework version
The framework 'Microsoft.AspNetCore.App', version '6.0.0' (x64) was not found.
  - No frameworks were found.

You can resolve the problem by installing the specified framework and/or SDK.

The specified framework can be found at:
  - https://aka.ms/dotnet-core-applaunch?framework=Microsoft.AspNetCore.App&framework_version=6.0.0&arch=x64&rid=mariner.2.0-x64
```

PR #228, opened 2022-08-05
Add compression support

wi-y:
Summary of changes

Normalization
- Add ability to compress the measurementGroup and send the compressed measurementGroup.
- Add way to flag if we are sending a measurement or measurementGroup (set IsMeasurementGroup property)
- Add way to flag if we are sending a compressed event payload (set ContentType to application/gzip)
- The compression on the normalization side is off by default, since we would like to first support it on the MeasurementToFhir side first

MeasurementToFhir
- Detect if event batch contains measurements, measurementGroups, or both
- Add ability to process any of the above, or combinations of the above
- Add ability to decompress event payload if it is compressed

### DIFF
--- a/src/lib/Microsoft.Health.Common/IO/Compression.cs
+++ b/src/lib/Microsoft.Health.Common/IO/Compression.cs
@@ -10,6 +10,8 @@ namespace Microsoft.Health.Common.IO
 {
     public static class Compression
     {
+        public static string GzipContentType { get; } = "application/gzip";
+
         public static byte[] CompressWithGzip(byte[] bytes)
         {
             using (var memoryStream = new MemoryStream())
@@ -23,19 +25,31 @@ namespace Microsoft.Health.Common.IO
             }
         }
 
+        public static Stream DecompressWithGzip(Stream compressedStream)
+        {
+            var decompressedStream = new MemoryStream();
+
+            using (var gzip = new GZipStream(compressedStream, CompressionMode.Decompress))
+            {
+                gzip.CopyTo(decompressedStream);
+                decompressedStream.Position = 0;
+                return decompressedStream;
+            }
+        }
+
         public static byte[] DecompressWithGzip(byte[] bytes)
         {
             using (var compressedStream = new MemoryStream(bytes))
             {
-                var decompressedStream = new MemoryStream();
+                using (var decompressedStream = new MemoryStream())
                 {
                     using (var gzip = new GZipStream(compressedStream, CompressionMode.Decompress))
                     {
                         gzip.CopyTo(decompressedStream);
                     }
-                }
 
-                return decompressedStream.ToArray();
+                    return decompressedStream.ToArray();
+                }
             }
         }
     }
--- a/src/lib/Microsoft.Health.Common/IO/Compression.cs
+++ b/src/lib/Microsoft.Health.Common/IO/Compression.cs
@@ -1,0 +1,42 @@
+// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System.IO;
+using System.IO.Compression;
+
+namespace Microsoft.Health.Common.IO
+{
+    public static class Compression
+    {
+        public static byte[] CompressWithGzip(byte[] bytes)
+        {
+            using (var memoryStream = new MemoryStream())
+            {
+                using (var gzipStream = new GZipStream(memoryStream, CompressionLevel.Fastest))
+                {
+                    gzipStream.Write(bytes, 0, bytes.Length);
+                }
+
+                return memoryStream.ToArray();
+            }
+        }
+
+        public static byte[] DecompressWithGzip(byte[] bytes)
+        {
+            using (var compressedStream = new MemoryStream(bytes))
+            {
+                var decompressedStream = new MemoryStream();
+                {
+                    using (var gzip = new GZipStream(compressedStream, CompressionMode.Decompress))
+                    {
+                        gzip.CopyTo(decompressedStream);
+                    }
+                }
+
+                return decompressedStream.ToArray();
+            }
+        }
+    }
+}
--- a/src/lib/Microsoft.Health.Fhir.Ingest/Data/MeasurementToEventMessageAsyncCollector.cs
+++ b/src/lib/Microsoft.Health.Fhir.Ingest/Data/MeasurementToEventMessageAsyncCollector.cs
@@ -31,12 +31,12 @@ namespace Microsoft.Health.Fhir.Ingest.Data
             IEventHubMessageService eventHubService,
             IHashCodeFactory hashCodeFactory,
             ITelemetryLogger telemetryLogger,
-            IOptions<MeasurementToEventMessageAsyncCollectorOptions> options = null)
+            IOptions<MeasurementToEventMessageAsyncCollectorOptions> options)
         {
             _eventHubService = EnsureArg.IsNotNull(eventHubService, nameof(eventHubService));
             _hashCodeFactory = EnsureArg.IsNotNull(hashCodeFactory, nameof(hashCodeFactory));
             _telemetryLogger = EnsureArg.IsNotNull(telemetryLogger, nameof(telemetryLogger));
-            _options = options;
+            _options = options ?? Options.Create(new MeasurementToEventMessageAsyncCollectorOptions());
         }
 
         public async Task AddAsync(IMeasurement item, CancellationToken cancellationToken = default(CancellationToken))
@@ -78,10 +78,9 @@ namespace Microsoft.Health.Fhir.Ingest.Data
                     if (_options != null && _options.Value.UseCompressionOnSend == true)
                     {
                         var measurementGroupContent = JsonConvert.SerializeObject(grp, Formatting.None);
-                        var contentBytes = Encoding.UTF8.GetBytes(measurementGroupContent);
-                        var compressedBytes = Common.IO.Compression.CompressWithGzip(contentBytes);
+                        var compressedBytes = Common.IO.Compression.CompressWithGzip(System.Text.Json.JsonSerializer.SerializeToUtf8Bytes(grp));
                         var eventData = new EventData(compressedBytes);
-                        eventData.ContentType = "application/gzip";
+                        eventData.ContentType = Common.IO.Compression.GzipContentType;
                         eventData.Properties["IsMeasurementGroup"] = true;
 
                         if (!currentEventDataBatch.TryAdd(eventData))
--- a/src/lib/Microsoft.Health.Fhir.Ingest/Data/MeasurementToEventMessageAsyncCollector.cs
+++ b/src/lib/Microsoft.Health.Fhir.Ingest/Data/MeasurementToEventMessageAsyncCollector.cs
@@ -23,6 +23,7 @@ namespace Microsoft.Health.Fhir.Ingest.Data
         private readonly IEventHubMessageService _eventHubService;
         private readonly IHashCodeFactory _hashCodeFactory;
         private readonly ITelemetryLogger _telemetryLogger;
+        private readonly string _compressionEnabled;
 
         public MeasurementToEventMessageAsyncCollector(
             IEventHubMessageService eventHubService,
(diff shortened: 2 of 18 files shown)


src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementFhirImportService.cs
```diff
@@ -90,9 +120,85 @@ public async Task ProcessEventsAsync(IEnumerable<IEventMessage> events, string t
...
+                {
+                    IEnumerable<MeasurementGroup> groupedMeasurements;
+
+                    try
+                    {
+                        var partitionId = events.FirstOrDefault()?.PartitionId;
+
+                        // decompress the measurement group if it is compressed
+                        byte[] bytes;
+                        if (evt.BodyContentType == "application/gzip")
```
rogordon01, line 147: If we go with my previous suggestions we could avoid extra object allocations here with something like:

```.net
                        IEnumerable<Measurement> measurementGroup = null;

                        if (evt.BodyContentType == "application/gzip")
                        {
                            using var stream = Compression.DecompressWithGzip(evt.Body.AsStream()); 
                            measurementGroup = System.Text.Json.JsonSerializer.Deserialize<IEnumerable<Measurement>>(stream);
                        }
                        else
                        {
                            using var stream = evt.Body.AsStream();
                            measurementGroup = System.Text.Json.JsonSerializer.Deserialize<IEnumerable<Measurement>>(stream)
                        }
```

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementFhirImportService.cs
```diff
@@ -90,9 +120,85 @@ public async Task ProcessEventsAsync(IEnumerable<IEventMessage> events, string t
...
+                        {
+                            bytes = Compression.DecompressWithGzip(evt.Body.ToArray());
+                            evt.Body = new ReadOnlyMemory<byte>(bytes);
```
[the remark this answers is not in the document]
wi-y: It is technically being used in the case where an exception is being thrown. If an exception is thrown I wanted the error message service to have access to the uncompressed bytes rather than the compressed bytes. But the downside is that we are setting this evt.Body on every message whether an exception is thrown or not. We could change this to have the error message service deal with the uncompressing if that is preferred

src/lib/Microsoft.Health.Fhir.Ingest/Data/MeasurementToEventMessageAsyncCollector.cs
```diff
@@ -23,15 +25,18 @@ public class MeasurementToEventMessageAsyncCollector :
...
         private readonly IHashCodeFactory _hashCodeFactory;
         private readonly ITelemetryLogger _telemetryLogger;
+        private readonly IOptions<MeasurementToEventMessageAsyncCollectorOptions> _options;
 
         public MeasurementToEventMessageAsyncCollector(
             IEventHubMessageService eventHubService,
             IHashCodeFactory hashCodeFactory,
-            ITelemetryLogger telemetryLogger)
+            ITelemetryLogger telemetryLogger,
+            IOptions<MeasurementToEventMessageAsyncCollectorOptions> options = null)
```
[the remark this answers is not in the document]
wi-y: that makes sense

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementFhirImportService.cs
dustinburson, line 147: We should have constants for this.

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementFhirImportService.cs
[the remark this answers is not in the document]
dustinburson: I kind of like that approach but if we do that then we would want separate the responsibility for parsing the message into it's own class so we can reuse the logic.

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementFhirImportService.cs
```diff
@@ -90,9 +120,85 @@ public async Task ProcessEventsAsync(IEnumerable<IEventMessage> events, string t
...
+                        }
+                        else
+                        {
+                            bytes = evt.Body.ToArray();
+                        }
+
+                        // deserialize
+                        IEnumerable<Measurement> measurementGroup = JsonConvert.DeserializeObject<IEnumerable<Measurement>>(System.Text.Encoding.UTF8.GetString(bytes));
```
dustinburson, line 158: Is there an option with compression that we can get a stream instead of a byte array?  We can get the readonly memory stream from event body.  So if we can use a stream and instead of byte we should be able to avoid the array allocation for the byte array.

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementFhirImportService.cs
```diff
@@ -90,9 +120,85 @@ public async Task ProcessEventsAsync(IEnumerable<IEventMessage> events, string t
...
+                        var partitionId = events.FirstOrDefault()?.PartitionId;
+
+                        // decompress the measurement group if it is compressed
+                        byte[] bytes;
+                        if (evt.BodyContentType == "application/gzip")
+                        {
+                            bytes = Compression.DecompressWithGzip(evt.Body.ToArray());
+                            evt.Body = new ReadOnlyMemory<byte>(bytes);
+                        }
+                        else
```
dustinburson, line 152: Personally, I prefer using the switch expressions for scenarios like this.  See https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/switch-expression.  It doesn't really matter for one if/else statement but if we add other types the switch expression I feel results in more readable code compared to a large if/if else/else statement.

It would also be good to have a scenario if content is set to a type we don't support/know and raise an error

dustinburson (changes requested):
Overall changes look good. 

Two main things.  Where we are using byte[] if we can use a stream instead please look into making the switch.  This should reduce the memory allocation pressure we are seeing in perf tests.  Second, update the test cases so they also cover the compression scenarios.

Thanks!

src/lib/Microsoft.Health.Common/IO/Compression.cs
```diff
@@ -0,0 +1,42 @@
...
+            {
+                var decompressedStream = new MemoryStream();
+                {
+                    using (var gzip = new GZipStream(compressedStream, CompressionMode.Decompress))
+                    {
+                        gzip.CopyTo(decompressedStream);
+                    }
+                }
+
+                return decompressedStream.ToArray();
```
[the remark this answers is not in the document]
wi-y: thanks for the suggestion. updated

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementFhirImportService.cs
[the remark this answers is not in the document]
wi-y: I removed it for now and will add a work item for a refactor for the error message service

src/lib/Microsoft.Health.Common/IO/Compression.cs
```diff
@@ -0,0 +1,42 @@
...
+                using (var gzipStream = new GZipStream(memoryStream, CompressionLevel.Fastest))
+                {
+                    gzipStream.Write(bytes, 0, bytes.Length);
+                }
+
+                return memoryStream.ToArray();
+            }
+        }
+
+        public static byte[] DecompressWithGzip(byte[] bytes)
```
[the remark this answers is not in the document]
wi-y: updated to use a stream

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementFhirImportService.cs
wi-y: yeah we can use a stream. updated

test/Microsoft.Health.Events.UnitTest/EventBatchingTests.cs
```diff
@@ -154,7 +154,7 @@ public async void GivenBatchServiceRunning_WhenNewEventConsumed_ThenCheckpointUp
             var eventReader = new EventBatchingService(_eventConsumerService, _options, _checkpointClient, _logger);
 
             var firstEventTime = DateTime.UtcNow.AddSeconds(-301);
-            var firstEvent = new EventMessage("0", new ReadOnlyMemory<byte>(), 1, 1, firstEventTime, new Dictionary<string, object>(), new ReadOnlyDictionary<string, object>(new Dictionary<string, object>()));
+            var firstEvent = new EventMessage("0", new ReadOnlyMemory<byte>(), null, 1, 1, firstEventTime, new Dictionary<string, object>(), new ReadOnlyDictionary<string, object>(new Dictionary<string, object>()));
```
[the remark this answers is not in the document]
wi-y: added some additional unit tests in MeasurementFhirImportServiceTests.cs to cover the new scenarios

test/Microsoft.Health.Common.UnitTests/Microsoft.Health.Common.UnitTests.csproj
```diff
@@ -30,7 +30,6 @@
 			<PrivateAssets>all</PrivateAssets>
 			<IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
 		</PackageReference>
-		<PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
```
dustinburson, line 33: Reason for removal?  This was recently added because the Microsoft Test package has a Newtonsoft dependency and this ensures 13.0.1 or higher is used.

test/Microsoft.Health.Fhir.Ingest.UnitTests/Microsoft.Health.Fhir.Ingest.UnitTests.csproj
```diff
@@ -18,7 +18,6 @@
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.2.0" />
-    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
```
dustinburson, line 21: Same comment as before.  We should keep these references in the Test projects.

test/Microsoft.Health.Fhir.R4.Ingest.UnitTests/Microsoft.Health.Fhir.R4.Ingest.UnitTests.csproj
```diff
@@ -18,7 +18,6 @@
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.2.0" />
-    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
```
dustinburson, line 21: Same comment as before.  We should keep these references in the Test projects.

test/Microsoft.Health.Tests.Common/Microsoft.Health.Tests.Common.csproj
```diff
@@ -27,7 +27,6 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
-	<PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
```
dustinburson, line 30: Same comment as before.  We should keep these references in the Test projects.

dustinburson (changes requested):
Changes to use streams instead of byte[] arrays look good.

Only issue is the removal of the ```<PackageReference Include="Newtonsoft.Json" Version="13.0.1" />```  package references in the Test projects.  Please add them back in.  Please let me know if these were removed for a reason so we can see what options we have.  Thanks!

wi-y:
> Changes to use streams instead of byte[] arrays look good.
> 
> Only issue is the removal of the `<PackageReference Include="Newtonsoft.Json" Version="13.0.1" />` package references in the Test projects. Please add them back in. Please let me know if these were removed for a reason so we can see what options we have. Thanks!

I see. I've added the dependency back so that each test project has a Newtonsoft.Json 13.0.1 version.

There were two cases where there were duplicates, which caused a build error, so I have removed the duplicates in those two cases. All of the test csproj files should have exactly one Newtonsoft.Json now which I believe is the correct approach.

Cases where we had duplicates:
Microsoft.Health.Tests.Common.csproj
https://github.com/microsoft/iomt-fhir/blob/main/test/Microsoft.Health.Tests.Common/Microsoft.Health.Tests.Common.csproj#L30

Microsoft.Health.Fhir.Ingest.UnitTests.csproj
https://github.com/microsoft/iomt-fhir/blob/main/test/Microsoft.Health.Fhir.Ingest.UnitTests/Microsoft.Health.Fhir.Ingest.UnitTests.csproj#L34

dustinburson (approved):

PR #231, opened 2022-08-11
Remove dependency on deprecated Microsoft.Azure.EventHubs package

dustinburson:
- Remove dependency on deprecated Microsoft.Azure.EventHubs package from Microsoft.Health.Events and Microsoft.Health.Fhir.Ingest libraries.
- Create new library Microsoft.Health.Fhir.Ingest.Legacy and move classes with dependencies to Microsoft.Azure.EventHubs here.  Only the Azure Function project references the legacy project.  This is because the package Azure.Messaging.EventHubs has breaking changes and the function app can't be updated with out introducing backward compat issues.
- Remove old Stream Analytics query project.


src/lib/Microsoft.Health.Fhir.Ingest/Service/EventProcessingMeter.cs
```diff
@@ -19,21 +20,23 @@ public Task<EventStats> CalculateEventStats(IEnumerable<EventData> events)
...
             {
-                TotalEventsProcessedBytes = ingressSizeBytes,
+                TotalEventsProcessedBytes = Convert.ToDouble(ingressSizeBytes),
             };
 
             return Task.FromResult(eventStats);
         }
 
-        private double CalculateDictionarySizeBytes(IDictionary<string, object> dictionary)
+        public static int CalculateDictionarySizeBytes(IEnumerable<KeyValuePair<string, object>> dictionary)
```
rogordon01, line 37: Will we lose precision by moving from a `double` to an `int`? Should we change this to a `long`?

rogordon01, line 37: Will we lose precision by moving from a `double` to an `int`? Should we change this to a `long`?

src/lib/Microsoft.Health.Fhir.Ingest.Legacy/Errors/ExceptionContextExtensions.cs
```diff
@@ -0,0 +1,21 @@
...
+
+namespace Microsoft.Health.Events.Errors
+{
+    public static class ExceptionContextExtensions
+    {
+        private const string EventsAttr = "RelatedEvents";
+
+        public static void AddEventContext(this Exception ex, IEnumerable<EventData> events)
+        {
+            ex.Data[EventsAttr] = events;
```
rogordon01, line 18: Why do we need to add in the list of events to the `Exception`? Where does that get used? Also, are we sure that this attribute won't get logged with the Exception??

src/lib/Microsoft.Health.Fhir.Ingest/Service/EventProcessingMeter.cs
dustinburson: The old code had an implicit conversion from int to double.  The code to the right of the assignment was still treating everything as an int.

The method Encoding.UTF8.GetByteCount returns int as does the Linq method Aggregate when it binds to the underlying delegates.  If we were worried about overflow once all the bytes are added I could see using a long for the variable we are adding everything into but in that case we wouldn't be able to use the Aggregate method.  Since this represents some of the bytes in an event hub message, and the max message size is 1 MB we should be well under the max size of an int.

Double is the signature for the metrics API which is why we do the conversion.

src/lib/Microsoft.Health.Fhir.Ingest.Legacy/Errors/ExceptionContextExtensions.cs
dustinburson: The original code was added as part of the Event Messaging service work @wi-y is doing so we can capture the context of the events that caused an error.  This is deployed and happening today and isn't automatically logged in the exception.

It is currently used in the class MeasurementEventNormalizationService and the method ProcessErrorAsync.  Previously this code was pointing to [src/lib/Microsoft.Health.Events/Errors/IErrorMessageWithEvents.cs](https://github.com/microsoft/iomt-fhir/pull/231/files/288c8cbe7544d15d79a916de4cae85f7650f6562#diff-7f7122b7defeb4d195cd3d421c1cc8e318c16a22cb4b583f8f62d221e2e2bd8f).  This is no longer valid since this class was updated to remove the old event hub SDK in favor of the new one.  The updated extension in the Microsoft.Health.Events library still consumes EventData but the type is different since it comes from the new library, not the old one.

The other option is I can remove this extension and the corresponding code in the class MeasurementEventNormalizationService
@wi-y do you have any preferences?  The class MeasurementEventNormalizationService is no longer used outside the legacy functions.

wi-y: We don't really need it since we don't really plan on supporting error message service for the function based app. We can remove this and the corresponding code in MeasurementEventNormalizationService

dustinburson: Thanks I will update to remove this piece.

rogordon01 (approved):

pallar-ms (approved):

PR #233, opened 2022-08-23
Add attempt to resolve a single observation when FHIR server returns multiple records in a bundle

dustinburson:
- Update MultipleResourceFoundException to include optional collection of ids that will be included in the error message.
- Add logic to R4FhirImportService to not immediately fail if multiple Observations are return.  Attempt to find only one observation in the returned collection that matches the desired identifier. If multiple observations still exist throw the expection as normal.
- Add method, GetMeasurementKey, to ensure same logic is used for generating keys for measurement grouping.

### DIFF
--- a/src/lib/Microsoft.Health.Extensions.Fhir.R4/BundleExtensions.cs
+++ b/src/lib/Microsoft.Health.Extensions.Fhir.R4/BundleExtensions.cs
@@ -62,7 +62,7 @@ namespace Microsoft.Health.Extensions.Fhir
 
             if (throwOnMultipleFound && resourceCount > 1)
             {
-                throw new MultipleResourceFoundException<TResource>(resourceCount);
+                throw new MultipleResourceFoundException<TResource>(resourceCount, resources.Select(r => r.Id));
             }
 
             return resources.FirstOrDefault();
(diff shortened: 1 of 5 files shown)


rogordon01 (approved):

src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4FhirImportService.cs
```diff
@@ -202,7 +197,35 @@ protected static Model.Identifier GenerateObservationIdentifier(IObservationGrou
...
+            if (foundObservations.Length == 0)
+            {
+                return null;
+            }
+
+            // Discovered an issue where FHIR Service is only matching on first 128 characters of the identifier.
+            // Match observations that match the desired identifier.
+            // This will also prevent merging observations that share the first 128 characters.
+
+            var matchedObservations = foundObservations
```
pallar-ms, line 213: nit: Should we log (warning) the fact that we even got multiple `foundObservations` for a given identifier (even post the fhir service bug, since it is not an expected scenario)?

pallar-ms (approved):

src/lib/Microsoft.Health.Extensions.Fhir.R4/BundleExtensions.cs
```diff
@@ -78,7 +78,7 @@ public static int EntryCount(this Bundle bundle)
             return bundle?.Entry?.Count ?? 0;
         }
 
-        private static async Task<IEnumerable<TResource>> ReadFromBundleWithContinuationAsync<TResource>(
+        public static async Task<IEnumerable<TResource>> ReadFromBundleWithContinuationAsync<TResource>(
```
kyclai, line 81: nit: Can unit tests be added for this method? - Backlog item?

kyclai (approved):

PR #234, opened 2022-08-24
Handle case where an event reader with epoch ownership is connected

wi-y:
The recommended way to read from Event Hub from production workloads is using the [EventProcessorClient](https://docs.microsoft.com/en-us/azure/event-hubs/event-processor-balance-partition-load). This utilizes blob storage to manage ownership and partition balancing. The iomt-fhir project current using the EventProcessorClient to connect to Event Hub.

However there are also other ways to read from Event Hub. When reading from an Event Hub using the [EventHubConsumerClient](https://docs.microsoft.com/en-us/dotnet/api/azure.messaging.eventhubs.consumer.eventhubconsumerclient?view=azure-dotnet) (not recommended for production) or the [EventHubClient (legacy)](https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.eventhubs.eventhubclient.createepochreceiver?view=azure-dotnet) it is possible to supply a unique identifier that indicates partition ownership. For EventHubConsumerClient ownership is done via [ReadEventOptions.OwnerLevel](https://docs.microsoft.com/en-us/dotnet/api/azure.messaging.eventhubs.consumer.readeventoptions.ownerlevel?view=azure-dotnet#azure-messaging-eventhubs-consumer-readeventoptions-ownerlevel) and for EventHubClient ownership is done via [CreateEpochReceiver](https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.eventhubs.eventhubclient.createepochreceiver?view=azure-dotnet).

It is possible that a user could do the following two things simultaneously:
1. User sets up an event reader to read from their device Event Hub and supplies ownership values
2. User tries to run the Iot Connector (which also reads from the device Event Hub).

In this scenario, when the Iot Connector uses the EventProcessorClient to attempt to read messages, the outcomes is that the Iot Connector will not be able to read because it is not able to obtain ownership. An error will be thrown:

`Receiver 'nil' with a higher epoch <epoch value> already exists. Receiver 'validate' with epoch 0 cannot be created. Make sure you are creating receiver with increasing epoch value to ensure connectivity, or ensure all old epoch receivers are closed or disconnected`

This PR will handle the exception that is currently being thrown and will log an appropriate error message to the user. We will then also try and reconnect in 1 minute.

### DIFF
--- a/src/lib/Microsoft.Health.Events/EventHubProcessor/BaseEventProcessor.cs
+++ b/src/lib/Microsoft.Health.Events/EventHubProcessor/BaseEventProcessor.cs
@@ -104,5 +104,22 @@ namespace Microsoft.Health.Events.EventHubProcessor
                 EventHubExceptionProcessor.ProcessException(ex, Logger, errorMetricName: EventHubErrorCode.EventHubPartitionInitFailed.ToString());
             }
         }
+
+        protected virtual void HandleOwnershipFailure(AggregateException ex)
+        {
+            foreach (var e in ex.InnerExceptions)
+            {
+                var eventHubException = e as EventHubsException;
+
+                // another consumer with a higher epoch has ownership
+                if (eventHubException != null && eventHubException.Reason == EventHubsException.FailureReason.ConsumerDisconnected)
+                {
+                    EventHubExceptionProcessor.ProcessException(eventHubException, Logger, errorMetricName: EventHubErrorCode.ConfigurationError.ToString());
+                    return;
+                }
+            }
+
+            throw ex;
+        }
     }
 }
--- a/src/lib/Microsoft.Health.Events/EventHubProcessor/BaseEventProcessor.cs
+++ b/src/lib/Microsoft.Health.Events/EventHubProcessor/BaseEventProcessor.cs
@@ -4,6 +4,7 @@
 // -------------------------------------------------------------------------------------------------
 
 using System;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Azure.Messaging.EventHubs;
@@ -107,16 +108,14 @@ namespace Microsoft.Health.Events.EventHubProcessor
 
         protected virtual void HandleOwnershipFailure(AggregateException ex)
         {
-            foreach (var e in ex.InnerExceptions)
-            {
-                var eventHubException = e as EventHubsException;
+            var consumerDisconnectedEx = ex.InnerExceptions
+                .OfType<EventHubsException>()
+                .FirstOrDefault(e => e.Reason == EventHubsException.FailureReason.ConsumerDisconnected);
 
-                // another consumer with a higher epoch has ownership
-                if (eventHubException != null && eventHubException.Reason == EventHubsException.FailureReason.ConsumerDisconnected)
-                {
-                    EventHubExceptionProcessor.ProcessException(eventHubException, Logger, errorMetricName: EventHubErrorCode.ConfigurationError.ToString());
-                    return;
-                }
+            if (consumerDisconnectedEx != null)
+            {
+                EventHubExceptionProcessor.ProcessException(consumerDisconnectedEx, Logger, errorMetricName: EventHubErrorCode.ConfigurationError.ToString());
+                return;
             }
 
             throw ex;
--- a/src/lib/Microsoft.Health.Events/EventHubProcessor/EventProcessor.cs
+++ b/src/lib/Microsoft.Health.Events/EventHubProcessor/EventProcessor.cs
@@ -46,7 +46,7 @@ namespace Microsoft.Health.Events.EventHubProcessor
                 }
 
                 Logger.LogTrace("Unable to read from event hub. Retrying in 1 minute");
-                await Task.Delay(TimeSpan.FromMinutes(1));
+                await Task.Delay(TimeSpan.FromMinutes(1), ct);
             }
 
             // Wait indefinitely until cancellation is requested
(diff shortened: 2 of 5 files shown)


src/lib/Microsoft.Health.Events/EventHubProcessor/EventProcessor.cs
```diff
@@ -25,25 +25,38 @@ public override async Task RunAsync(CancellationToken ct)
...
+            bool connected = false;
 
-            try
+            while (!connected)
             {
-                Console.WriteLine($"Starting event hub processor at {DateTime.UtcNow}");
-                await EventProcessorClient.StartProcessingAsync(ct);
+                try
+                {
+                    EventProcessorClient.ProcessEventAsync += ProcessEventHandler;
```
dustinburson, line 34: Suggestion: Do we need to remove/add the event handlers when there is an exception? Or can we do the loop just around the start/wait and and have a try/catch just for ownership failures?

src/lib/Microsoft.Health.Events/EventHubProcessor/BaseEventProcessor.cs
```diff
@@ -104,5 +104,22 @@ protected virtual async Task ProcessInitializingHandler(PartitionInitializingEve
                 EventHubExceptionProcessor.ProcessException(ex, Logger, errorMetricName: EventHubErrorCode.EventHubPartitionInitFailed.ToString());
             }
         }
+
+        protected virtual void HandleOwnershipFailure(AggregateException ex)
```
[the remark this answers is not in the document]
wi-y: The event hub code seems to always throw an aggregate exception for the ownership failure: https://github.com/Azure/azure-sdk-for-net/blob/ae550cb5d4be459e61d0f42f46c00033b0fc9a83/sdk/eventhub/Azure.Messaging.EventHubs/src/Primitives/EventProcessor.cs#L1266

As far as I know this is the only exception we've seen at startup. We could have a catch-all for any other Exception and we could periodically retry just like we are doing here for the ownership failure, but it is sort of unclear if that is appropriate. In some cases it may be better to throw the exception and crash and see if the restart fixes the issue.

src/lib/Microsoft.Health.Events/EventHubProcessor/EventProcessor.cs
```diff
@@ -25,25 +25,38 @@ public override async Task RunAsync(CancellationToken ct)
...
+                    EventProcessorClient.ProcessErrorAsync += ProcessErrorHandler;
+                    EventProcessorClient.PartitionInitializingAsync += ProcessInitializingHandler;
 
-                // Wait indefinitely until cancellation is requested
-                ct.WaitHandle.WaitOne();
+                    Console.WriteLine($"Starting event hub processor at {DateTime.UtcNow}");
```
rogordon01, line 38: Why the use of `Console` versus the TelemetryLogger?

src/lib/Microsoft.Health.Events/EventHubProcessor/BaseEventProcessor.cs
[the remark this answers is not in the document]
wi-y: updated to use the LINQ suggestion

src/lib/Microsoft.Health.Events/EventHubProcessor/EventProcessor.cs
wi-y: we can add/remove the handlers outside of the while loop. updated the code to add/remove outside of the while loop

src/lib/Microsoft.Health.Events/EventHubProcessor/EventProcessor.cs
```diff
@@ -29,12 +29,29 @@ public override async Task RunAsync(CancellationToken ct)
...
+                    connected = true;
+                    break;
+                }
+                catch (AggregateException ex)
+                {
+                    HandleOwnershipFailure(ex);
+                }
+
+                Logger.LogTrace("Unable to read from event hub. Retrying in 1 minute");
+                await Task.Delay(TimeSpan.FromMinutes(1));
```
rogordon01, line 49: We should pass in the `cancellationToken` into `Task.Delay`. In case a cancellation request is called on the token.

src/lib/Microsoft.Health.Events/EventHubProcessor/EventProcessor.cs
```diff
@@ -29,12 +29,29 @@ public override async Task RunAsync(CancellationToken ct)
             EventProcessorClient.ProcessErrorAsync += ProcessErrorHandler;
             EventProcessorClient.PartitionInitializingAsync += ProcessInitializingHandler;
 
-            try
+            // Try to connect and read events
+            bool connected = false;
+            while (!connected)
```
rogordon01, line 34: This looks correct to me. The other option is to implement a Polly retry policy here. Not sure if that is overkill

src/lib/Microsoft.Health.Events/EventHubProcessor/EventProcessor.cs
```diff
@@ -29,12 +29,29 @@ public override async Task RunAsync(CancellationToken ct)
             EventProcessorClient.ProcessErrorAsync += ProcessErrorHandler;
             EventProcessorClient.PartitionInitializingAsync += ProcessInitializingHandler;
 
-            try
+            // Try to connect and read events
```
rogordon01, line 32: One thing that I forgot to mention: I don't think that we use this class in PaaS. We use this [class](https://dev.azure.com/microsofthealth/Health/_git/workspace-platform?path=/iot-connector/core/iomt-fhir/Microsoft.Health.Fhir.Ingest.Console.PaaS/ResumableEventProcessor.cs), which extends `BaseEventProcessor`. So we might also need these changes in there.

rogordon01 (approved):

src/lib/Microsoft.Health.Events/EventHubProcessor/EventProcessor.cs
wi-y: yeah I will update it there when this is merged

PR #235, opened 2022-08-30
Use separate classes for compression support, toggle compression classes via env variable

wi-y:
This PR will revert the state of the Normalization and MeasurementToFhir applications to a state before compression was supported.

Reverted MeasurementToEventMessageAsyncCollector.cs to original state (compression off)
Created MeasurementGroupToCompressedEventMessageAsyncCollector.cs (compression on)
Can toggle between the two classes using CollectorOptions. Default is compression off.

Reverted MeasurementFhirImportService.cs to original state (no compression support)
Created MeasurementGroupFhirImportService.cs (has compression support)
Toggle between them using CollectorOptions (am open to creating another options class with the same properties, or renaming). The default is that compression support is off.


src/lib/Microsoft.Health.Fhir.Ingest/Data/MeasurementGroupToCompressedEventMessageAsyncCollector.cs
```diff
@@ -0,0 +1,99 @@
...
+            IEventHubMessageService eventHubService,
+            IHashCodeFactory hashCodeFactory,
+            ITelemetryLogger telemetryLogger)
+        {
+            _eventHubService = EnsureArg.IsNotNull(eventHubService, nameof(eventHubService));
+            _hashCodeFactory = EnsureArg.IsNotNull(hashCodeFactory, nameof(hashCodeFactory));
+            _telemetryLogger = EnsureArg.IsNotNull(telemetryLogger, nameof(telemetryLogger));
+        }
+
+        public async Task AddAsync(IMeasurement item, CancellationToken cancellationToken = default(CancellationToken))
```
dustinburson, line 37: I believe this is only used in the in the legacy code (Azure Function).  Long term it would be best to split implementations.  Have the simple one for backwards compatibility in the 

I will create a user story to collect this and other clean up refactors for a later PR.

src/console/MeasurementCollectionToFhir/ProcessorStartup.cs
```diff
@@ -49,19 +49,32 @@ public void ConfigureServices(IServiceCollection services)
 
             services.TryAddSingleton<ResourceManagementService>();
             services.TryAddSingleton<IExceptionTelemetryProcessor, FhirExceptionTelemetryProcessor>();
+            
             services.TryAddSingleton<MeasurementFhirImportOptions>();
-            services.TryAddSingleton<MeasurementFhirImportService>();
+
+            var collectorOptions = new CollectorOptions();
```
dustinburson, line 55: Suggestion: It would be better if this is a factory or a factory extension method that is part of the Events library.  This should allow easier reuse in PaaS.

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementGroupFhirImportService.cs
```diff
@@ -0,0 +1,362 @@
...
+        private readonly IExceptionTelemetryProcessor _exceptionTelemetryProcessor;
+
+        public MeasurementGroupFhirImportService(FhirImportService fhirImportService, MeasurementFhirImportOptions options, IExceptionTelemetryProcessor exceptionTelemetryProcessor)
+            : base(options, options?.ParallelTaskOptions?.MaxConcurrency ?? 1)
+        {
+            _fhirImportService = EnsureArg.IsNotNull(fhirImportService, nameof(fhirImportService));
+            _exceptionTelemetryProcessor = exceptionTelemetryProcessor;
+        }
+
+        public async Task ProcessStreamAsync(Stream data, string templateDefinition, ITelemetryLogger log)
```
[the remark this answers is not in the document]
dustinburson: Decided to create a user story for future refactor and clean up.  Will share with you to review.

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementFhirImportService.cs
```diff
@@ -18,13 +17,12 @@
...
 using Microsoft.Health.Fhir.Ingest.Template;
 using Microsoft.Health.Logging.Telemetry;
-using Microsoft.Toolkit.HighPerformance;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 
 namespace Microsoft.Health.Fhir.Ingest.Service
 {
-    public class MeasurementFhirImportService : ParallelTaskWorker<MeasurementFhirImportOptions>
+    public class MeasurementFhirImportService : ParallelTaskWorker<MeasurementFhirImportOptions>, IImportService
```
dustinburson, line 25: One minor change to guard against potential data loss if the Normalization is switched to compression but the FHIR Transformation isn't.

In the Linq query of ParseEventData where we are parsing events, if the content type is set then throw an exception since this class doesn't know how to handle it.  I would say throw a not supported exception but I suspect we handle and ignore those so we might need to use a new type.

src/console/MeasurementCollectionToFhir/ProcessorStartup.cs
wi-y: updated to use an extension here and in the other location where we reference the options and inject the appropriate class

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementFhirImportService.cs
wi-y: maybe we should throw a `DefinitelyNotSupportedException`?

kidding. added CompressionNotSupportedException to MeasurementFhirImportService.cs and added a new unit test for this scenario.

src/lib/Microsoft.Health.Fhir.Ingest/Exceptions/CompressionNotSupportedException.cs
```diff
@@ -0,0 +1,21 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using Microsoft.Health.Common.Telemetry;
+using Microsoft.Health.Common.Telemetry.Exceptions;
+
+namespace Microsoft.Health.Fhir.Ingest.Exceptions
+{
+    public class CompressionNotSupportedException : IomtTelemetryFormattableException
```
dustinburson, line 11: Does this mean it is customer facing?  If so I think that is incorrect since this wouldn't be actionable to the customer

dustinburson (approved):

src/lib/Microsoft.Health.Fhir.Ingest/Exceptions/CompressionNotSupportedException.cs
wi-y: No it won't be customer facing since IomtTelemetryFormattableException is not customer facing by default. If we were to instead inherit from [ThirdPartyLoggedFormattableException ](https://github.com/microsoft/iomt-fhir/blob/0ecfa7336df1516844bce009725f769f37b35510/src/lib/Microsoft.Health.Common/Telemetry/Exceptions/ThirdPartyLoggedFormattableException.cs) which inherits from IomtTelemetryFormattableException 
 then it would be customer facing.

wi-y: confirmed it does not get logged as a metric either. not sure we really need to inherit from IomtTelemetryFormattableException, though doing so does give some context to the developer about the severity, error type, stage even though it is not surfaced to an end user

dustinburson: Thanks for confirming!  Sorry I was misremembering the content from Artun's prior PR.

PR #236, opened 2022-09-02
Reset batch queues when partition initialized

wi-y:
There is a bug where we can read from a partition, then lose partition ownership to a another reader, then regain ownership. In the case where we regain ownership it is possible for previous events to exists in the event batching queues for that partition (from the previous point in time when reader had ownership). This is a problem because the next event read will flush the queue with the old events and update the checkpoint with the timestamp of the last old event. 

In the case where we regain partition ownership we should delete any previous knowledge/events, and start reading from the checkpoint provided. This PR ensures that the queue in cleared (deleted) when we begin reading from a partition so that it is not possible to retain any old events.

### DIFF
--- a/src/lib/Microsoft.Health.Events/EventHubProcessor/BaseEventProcessor.cs
+++ b/src/lib/Microsoft.Health.Events/EventHubProcessor/BaseEventProcessor.cs
@@ -91,6 +91,7 @@ namespace Microsoft.Health.Events.EventHubProcessor
 
             try
             {
+                EventConsumerService.NewPartitionInitialized(partitionId);
                 var checkpoint = await CheckpointClient.GetCheckpointForPartitionAsync(partitionId, initArgs.CancellationToken);
                 initArgs.DefaultStartingPosition = EventPosition.FromEnqueuedTime(checkpoint.LastProcessed);
                 Logger.LogTrace($"Starting to read partition {partitionId} from checkpoint {checkpoint.LastProcessed}");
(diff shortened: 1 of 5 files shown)


src/lib/Microsoft.Health.Events/EventHubProcessor/BaseEventProcessor.cs
```diff
@@ -91,6 +91,7 @@ protected virtual async Task ProcessInitializingHandler(PartitionInitializingEve
 
             try
             {
+                EventConsumerService.NewPartitionInitialized(partitionId);
```
dustinburson, line 94: This will work but it would be nice to free up the partition data when the ownership is lost.  There isn't a way to detect when ownership is lost?

dustinburson (approved):

src/lib/Microsoft.Health.Events/EventHubProcessor/BaseEventProcessor.cs
wi-y: yeah good point. there appears to a method `PartitionClosingAsync`  that gets called when the partition is being claimed by another processor: https://docs.microsoft.com/en-us/dotnet/api/azure.messaging.eventhubs.eventprocessorclient.partitionclosingasync?view=azure-dotnet. I can also clear the data there. 

I would like to also leave to logic at initialization to clear just in case

dustinburson: Sounds good, thanks!

PR #237, opened 2022-09-06
Clear partition data when partition ownership is lost

wi-y:
As part of the feedback for [PR#226](https://github.com/microsoft/iomt-fhir/pull/236) we should also consider resetting the partition data in our batching service when ownership is lost rather than just clearing the partition data on initialization. Clearing the partition data on ownership loss would free up memory when we clear the events from the partition queue of the partition we are no longer reader from.

This is a draft solution to reset the partition data when partition ownership is lost. The real solution unfortunately is not that simple. There could be cases where partition ownership is lost (we delete the queue), but this occurs at the same time that we are finishing [flushing events from the queue](https://github.com/microsoft/iomt-fhir/blob/main/src/lib/Microsoft.Health.Events/EventConsumers/Service/EventBatchingService.cs#L120). We should probably handle this potential race condition so that there isn't a case where the state of the queue cannot be altered while in the middle of flushing/publishing/checkpointing. This is a bit complex to implement.

### DIFF
--- a/src/lib/Microsoft.Health.Events/EventConsumers/Service/EventBatchingService.cs
+++ b/src/lib/Microsoft.Health.Events/EventConsumers/Service/EventBatchingService.cs
@@ -57,7 +57,7 @@ namespace Microsoft.Health.Events.EventConsumers.Service
             return _eventPartitions[partitionId];
         }
 
-        private bool EventPartitionExists(string partitionId)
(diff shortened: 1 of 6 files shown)


wi-y:
This PR is not worth pursuing due to the complexity. We clear the partition data on initialization which is sufficient

PR #241, opened 2022-10-21
Adapt batch size when horizontal scaling enabled

wi-y:
For the FHIR transformation there are some performance gains from scaling out and increasing the batch size. Currently the batch size is static but as we scale out (think 1 partition per application instance) then we technically have more memory and can increase the batch size. This will result in fewer FHIR application calls (assuming most of the data is for once device, which often is the case for needing to scale out) and will increase throughput.

The PR allows a variable to be set called `EventBatching__MaxEventsAllPartitions` which allows the user to specify the maximum events we should process across all the partitions. If this variable is set, and we are processing only 1 partition for example, then the number of events specified in EventBatching__MaxEventsAllPartitions will be the maximum batch size for that partition. If the application instance starts processing events from another partition (2 partitions total) then the number of total events allocated to each partition will be split (EventBatching__MaxEventsAllPartition divided by 2), and so on. The maximum batch size is calculated based on how many partitions the application is reading from, which correlates to how horizontally scaled the application is.

### DIFF
--- a/src/lib/Microsoft.Health.Events/EventConsumers/Service/EventBatchingService.cs
+++ b/src/lib/Microsoft.Health.Events/EventConsumers/Service/EventBatchingService.cs
@@ -23,6 +23,8 @@ namespace Microsoft.Health.Events.EventConsumers.Service
     {
         private ConcurrentDictionary<string, EventPartition> _eventPartitions;
         private int _maxEvents;
+        private int? _scaledOutMaxEventsAllPartitions;
+        private int? _scaledOutMaxEventsPerPartition;
         private TimeSpan _flushTimespan;
         private IEventConsumerService _eventConsumerService;
         private IEventProcessingMetricMeters _eventProcessingMetricMeters;
@@ -40,6 +42,8 @@ namespace Microsoft.Health.Events.EventConsumers.Service
             _eventConsumerService = eventConsumerService;
             _eventProcessingMetricMeters = eventProcessingMetricMeters;
             _maxEvents = options.MaxEvents;
+            _scaledOutMaxEventsAllPartitions = options.MaxEventsAllPartitions;
+            _scaledOutMaxEventsPerPartition = null;
             _flushTimespan = TimeSpan.FromSeconds(options.FlushTimespan);
             _checkpointClient = checkpointClient;
             _logger = logger;
@@ -64,7 +68,7 @@ namespace Microsoft.Health.Events.EventConsumers.Service
 
         private EventPartition CreatePartitionIfMissing(string partitionId, DateTime initTime, TimeSpan flushTimespan)
         {
-            return _eventPartitions.GetOrAdd(partitionId, new EventPartition(partitionId, initTime, flushTimespan, _logger));
+            return _eventPartitions.GetOrAdd(partitionId, new EventPartition(partitionId, initTime, flushTimespan, _maxEvents, _logger));
         }
 
         public void NewPartitionInitialized(string partitionId)
@@ -73,6 +77,34 @@ namespace Microsoft.Health.Events.EventConsumers.Service
             {
                 _eventPartitions.TryRemove(partitionId, out _);
             }
+
+            // if _scaledOutMaxEventsAllPartitions is set then assume horizontal scaling is enabled
+            if (_scaledOutMaxEventsAllPartitions != null)
+            {
+                CalculateMaximumEventsPerPartition();
+            }
+        }
+
+        public void PartitionProcessingStopped(string partitionId)
+        {
+            if (EventPartitionExists(partitionId))
+            {
+                _eventPartitions.TryRemove(partitionId, out _);
+            }
+
+            if (_scaledOutMaxEventsAllPartitions != null)
+            {
+                CalculateMaximumEventsPerPartition();
+            }
+        }
+
+        public void CalculateMaximumEventsPerPartition()
+        {
+            // if 1 partition is held then allocate all events to that partition
+            // if 2 partitions is held then allocate half of the events to one partition and half to the other partition
+            // etc
+            // if the result would be lower than MaxEvents, then use MaxEvents
+            _scaledOutMaxEventsPerPartition = Math.Max(_scaledOutMaxEventsAllPartitions.Value / Math.Max(_eventPartitions.Count, 1), _maxEvents);
         }
 
         public async Task ConsumeEvent(IEventMessage eventArg)
@@ -107,7 +139,7 @@ namespace Microsoft.Health.Events.EventConsumers.Service
                     await ThresholdTimeReached(partitionId, eventArg, windowThresholdTime);
                 }
 
-                if (partition.GetPartitionBatchCount() >= _maxEvents)
+                if (partition.GetPartitionBatchCount() >= partition.GetMaxEventsForPartition())
                 {
                     await ThresholdCountReached(partitionId);
                 }
@@ -116,9 +148,16 @@ namespace Microsoft.Health.Events.EventConsumers.Service
 
         private async Task ThresholdCountReached(string partitionId)
         {
-            _logger.LogTrace($"Partition {partitionId} threshold count {_maxEvents} was reached.");
-            var events = await GetPartition(partitionId).Flush(_maxEvents);
+            var partition = GetPartition(partitionId);
+            _logger.LogTrace($"Partition {partitionId} threshold count {partition.GetMaxEventsForPartition()} was reached.");
+            var events = await partition.FlushMaxEvents();
             await CompleteProcessing(partitionId, events, triggerReason: nameof(ThresholdCountReached));
+
+            // option to change the batch size in horizontal scaling scenarios
+            if (_scaledOutMaxEventsPerPartition != null)
+            {
+                partition.ChangeMaxEventsForPartition(_scaledOutMaxEventsPerPartition.Value);
+            }
         }
 
         private async Task ThresholdTimeReached(string partitionId, IEventMessage eventArg, DateTime windowEnd)
--- a/src/lib/Microsoft.Health.Events/EventConsumers/Service/EventBatchingService.cs
+++ b/src/lib/Microsoft.Health.Events/EventConsumers/Service/EventBatchingService.cs
@@ -24,7 +24,9 @@ namespace Microsoft.Health.Events.EventConsumers.Service
         private ConcurrentDictionary<string, EventPartition> _eventPartitions;
         private int _maxEvents;
         private int? _scaledOutMaxEventsAllPartitions;
-        private int? _scaledOutMaxEventsPerPartition;
+        private int _scaledOutMaxEventsPerPartition;
+        private static object _lock = new object();
+        private int _partitionCount;
         private TimeSpan _flushTimespan;
         private IEventConsumerService _eventConsumerService;
         private IEventProcessingMetricMeters _eventProcessingMetricMeters;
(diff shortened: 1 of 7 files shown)


src/lib/Microsoft.Health.Events/EventConsumers/Service/EventBatchingService.cs
```diff
@@ -64,7 +68,7 @@ public bool EventPartitionExists(string partitionId)
 
         private EventPartition CreatePartitionIfMissing(string partitionId, DateTime initTime, TimeSpan flushTimespan)
         {
-            return _eventPartitions.GetOrAdd(partitionId, new EventPartition(partitionId, initTime, flushTimespan, _logger));
+            return _eventPartitions.GetOrAdd(partitionId, new EventPartition(partitionId, initTime, flushTimespan, _maxEvents, _logger));
```
dustinburson, line 71: I am not sure of the logic flow.  After reviewing the code it appears that partitions are initially configured with the current default batch size.  Then the batch size will be adjusted on the first flush of that partition?  While this can work I think is fairly complicated which may lead to edge cases/bugs we aren't considering.

One potential issue I see is we will have multiple threads potentially reading/adjusting the max partition value.  This means we should interlocks at min on the partition count value the partition classes are using.

To simplify things I would propose the following.  Create a new class like PartitionSync (or a better :) ) so we can pass by ref to the partitions.  This class would contain the current max events to read.  We could add methods to pull the current value in a thread safe manner and used in the ConsumeEvent and FlushMax events.  The main advantage is you don't need to update the partition value, you can just reference the value the EventBatchingService maintains in a thread safe manner.

src/lib/Microsoft.Health.Events/EventConsumers/Service/EventBatchingService.cs
```diff
@@ -116,9 +148,16 @@ public async Task ConsumeEvent(IEventMessage eventArg)
 
         private async Task ThresholdCountReached(string partitionId)
         {
-            _logger.LogTrace($"Partition {partitionId} threshold count {_maxEvents} was reached.");
-            var events = await GetPartition(partitionId).Flush(_maxEvents);
+            var partition = GetPartition(partitionId);
+            _logger.LogTrace($"Partition {partitionId} threshold count {partition.GetMaxEventsForPartition()} was reached.");
+            var events = await partition.FlushMaxEvents();
```
dustinburson, line 153: It appears the partitions only need to know the MaxEvents when you call FlushMaxEvents.  Can we simplify even further over my avoid suggestion by just keep the partition count in the batching service and passing into this method the number of events to flush?  I think that would simplify the code greatly.  This class ends up being the only one that needs to maintain and track the target count.

src/lib/Microsoft.Health.Events/EventConsumers/Service/EventBatchingService.cs
```diff
@@ -73,6 +77,34 @@ public void NewPartitionInitialized(string partitionId)
             {
                 _eventPartitions.TryRemove(partitionId, out _);
             }
+
+            // if _scaledOutMaxEventsAllPartitions is set then assume horizontal scaling is enabled
```
dustinburson, line 81: Suggestion: We might want to have a more direct signal or at least modify the comment.  This absolutely makes since for the FHIR Conversion but I believe we are still using the EventBatchingService for normalization as well but have a very small batch size and I don't see that changing as we scale out since we don't get much benefit with batching for events.

src/lib/Microsoft.Health.Events/EventConsumers/Service/EventBatchingService.cs
wi-y: Updating to just have the max events in the batching class, rather than the partition class, and implemented some locking.

src/lib/Microsoft.Health.Events/EventConsumers/Service/EventBatchingService.cs
wi-y: updated to just keep the partition count in the batching service

src/lib/Microsoft.Health.Events/EventConsumers/Service/EventBatchingService.cs
wi-y: updated the comment. yeah we will probably not set this environment variable for normalization

src/lib/Microsoft.Health.Events/EventHubProcessor/BaseEventProcessor.cs
```diff
@@ -121,5 +121,13 @@ protected virtual void HandleOwnershipFailure(AggregateException ex)
 
             throw ex;
         }
+
+        protected virtual void PartitionClosingAsync(PartitionClosingEventArgs partitionClosingEventArgs)
+        {
+            var partitionId = partitionClosingEventArgs.PartitionId;
+            var reason = partitionClosingEventArgs.Reason;
+            Logger.LogTrace($"Stopping processing for partition {partitionId}. Reason {reason}");
```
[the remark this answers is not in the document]
wi-y: unfortunately I don't think the interface we use supports the structured logging

rogordon01:
Just to confirm: `EventBatching__MaxEventsAllPartitions` only applies within a specific instance, and not across all instances, correct? In other words it limits a specific instance to 1000 events, versus limiting all running instances to collectively working on 1000 events.

rogordon01:
How does adapting the batch size

wi-y:
> Just to confirm: `EventBatching__MaxEventsAllPartitions` only applies within a specific instance, and not across all instances, correct? In other words it limits a specific instance to 1000 events, versus limiting all running instances to collectively working on 1000 events.

@rogordon01 yeah EventBatching__MaxEventsAllPartitions applies to a specific application instance

src/lib/Microsoft.Health.Events/EventConsumers/Service/EventBatchingService.cs
```diff
@@ -73,6 +80,75 @@ public void NewPartitionInitialized(string partitionId)
...
+        {
+            lock (_lock)
+            {
+                _partitionCount--;
+            }
+        }
+
+        public int GetMaximumEventsForPartition()
+        {
+            if (_scaledOutMaxEventsPerPartition <= 0)
```
dustinburson, line 144: This should also have some thread safety for access as well.

src/lib/Microsoft.Health.Events/EventConsumers/Service/EventConsumerService.cs
```diff
@@ -51,5 +51,10 @@ public void NewPartitionInitialized(string partitionId)
         {
             // do nothing
         }
+
```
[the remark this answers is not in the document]
dustinburson: Suggestion: Add logging messages to identify when the partition is stopped.

src/lib/Microsoft.Health.Events/EventConsumers/Service/EventBatchingService.cs
```diff
@@ -23,6 +23,10 @@ public class EventBatchingService : IEventConsumerService
     {
         private ConcurrentDictionary<string, EventPartition> _eventPartitions;
         private int _maxEvents;
+        private int? _scaledOutMaxEventsAllPartitions;
+        private int _scaledOutMaxEventsPerPartition;
+        private static object _lock = new object();
```
dustinburson, line 28: Suggestion:  While you can lock I believe using the lighter weight interlock is more appropriate. See https://learn.microsoft.com/en-us/dotnet/api/system.threading.interlocked?view=net-7.0.  It looks like you just need to read, increment, or exchange values.  If so, the interlock corresponding options will work.  If you have multiple statements in the lock that need to be preserved as an atomic section then interlock wouldn't be a good solution.  It didn't appear you had any scenario like that but I could be mistaken.

dustinburson (changes requested):
Please add some thread safety to the method GetMaximumEventsForPartition.  Either adding a lock or using Interlocked.Read if you decide to switch to the interlock methods for thread safety.

src/lib/Microsoft.Health.Events/EventConsumers/Service/EventConsumerService.cs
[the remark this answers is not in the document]
wi-y: the method is only needed because of the interface. Originally it was added to support multiple event consumers, but in practice we only have one consumer. I think the flow is: Processor (read events from event hub) -> Consumers (do something with them) and we do log that the partition is stopped in the Processor ([BaseEventProcessor.cs](https://github.com/microsoft/iomt-fhir/pull/241/files/1eb834f77a7ec97ca83fd469c3af76f1e996de7a#diff-21400b2a07723f26ba56809facbbcf88a8f0121ad9c14218d2891db69e8677a1)) so didn't feel like we needed to log it again here. We should probably delete the EventConsumerService class in the long run. It appears to still be used but am not sure it really serves a purpose since we only have one consumer and we should probably refactor it out.

src/lib/Microsoft.Health.Events/EventConsumers/Service/EventBatchingService.cs
wi-y: We could add some thread safety but I chose not too due to the potential performance impact of locking/interlocking to read the variable and perform the check on each event to determine if we should flush or not. It should be fine to read the variable and perform the flush (or not perform the flush) based on whatever the value this method returns, even if the method returned a value that was slightly out of date.

The one condition we did need to account for is if the _scaledOutMaxEventsPerPartition were to increase such that _scaledOutMaxEventsPerPartition is greater than the current amount of events in the queue for the partition but in this PR have updated the code to account for that.

src/lib/Microsoft.Health.Events/EventConsumers/Service/EventConsumerService.cs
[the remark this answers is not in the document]
dustinburson: Thanks for the explanation Will.  Could you please encapsulate the details of what you are thinking in a User Story and send it to me?

src/lib/Microsoft.Health.Events/EventConsumers/Service/Infrastructure/EventPartition.cs
```diff
@@ -54,19 +54,25 @@ public int GetPartitionBatchCount()
...
 
-        // flush a fixed number of events
-        public Task<List<IEventMessage>> Flush(int numEvents)
+        public Task<List<IEventMessage>> FlushMaxEvents(int maxEvents)
         {
             var count = 0;
             var events = new List<IEventMessage>();
 
-            while (count < numEvents)
+            while (count < maxEvents)
```
dustinburson, line 62: Suggestion: Code can be simplified if 
```
while (count < maxEvents && _partition.TryDequeue(out var dequeuedEvent))
```
This should eliminate the need for the if/else with the break

src/lib/Microsoft.Health.Events/EventConsumers/Service/EventBatchingService.cs
dustinburson: In reviewing the code it seems you only have one section where both are used, ChangeMaximumEventsPerPartition.  You could do Interlock.Reads on both then interlock exchange on the max events but that would run the risk of potentially an invalid buffer if we have had two partition changes back to back and the last one finished processing first.  So in this case the lock would be appropriate.  Thanks for the explanation.

src/lib/Microsoft.Health.Events/EventConsumers/Service/EventBatchingService.cs
dustinburson: That is fair. It would be good to add comment on the property though explaining why we aren't applying thread safety so it doesn't appear like a bug when someone is looking at this later.

dustinburson (approved):

src/lib/Microsoft.Health.Events/EventConsumers/Service/EventConsumerService.cs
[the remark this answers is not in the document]
wi-y: @dustinburson created work item 98220

PR #244, opened 2022-11-07
Include version and update datetime on MultipleResourceFound exceptions

dustinburson:
* Update MultipleResourceFound exception to also include the version and last updated data in addition to the internal id.
* Fix issue with solution & Legacy library.

### DIFF
--- a/src/lib/Microsoft.Health.Extensions.Fhir/MultipleResourceFoundException.cs
+++ b/src/lib/Microsoft.Health.Extensions.Fhir/MultipleResourceFoundException.cs
@@ -5,6 +5,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Text;
 using Microsoft.Health.Common.Telemetry;
 using Microsoft.Health.Common.Telemetry.Exceptions;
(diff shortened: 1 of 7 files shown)


src/lib/Microsoft.Health.Extensions.Fhir/MultipleResourceFoundException.cs
```diff
@@ -43,14 +44,14 @@ public MultipleResourceFoundException(string message, Exception innerException)
...
+        private static string GenerateErrorMessage<TResource>(int resourceCount, IEnumerable<IResourceMetadata> metadata)
         {
             var sb = new StringBuilder($"Multiple resources {resourceCount} of type {typeof(T)} found, expected one.");
 
-            if (ids != null)
+            if (metadata != null)
             {
-                sb.Append(" Resource internal ids: ");
-                sb.Append(string.Join(", ", ids));
+                sb.Append(" Resource metadata: ");
```
rogordon01, line 53: Nit; the Append function returns a Stringbuilder. So you should be able to chain these calls together.

rogordon01 (approved):

PR #259, opened 2023-02-08
Prevent duplicate observations from being created

wi-y:
In order to prevent duplicate observations from being created, we will be changing two things.

1. Pass cancellation tokens through to downstream methods. If an event processor loses ownership of a partition we should receive a cancellation token and pass this cancellation token to our fhir import service and stop any additional requests.
2. While cancellations are a good start, they do not necessarily handle in flight requests. So in addition, when creating an observation, we should not let the server generate the observation id. Previously, when two processors were writing to the fhir service it was possible that two creates could happen at the same time which would result in 2 observations. In this pull request the client (medtech service) will hash the id so that the id is deterministic, and only 1 observation could be created.

There are still some limitations. With option 2 you could theoretically still have a race condition. If two clients theoretically are creating an observation of the FHIR server at the same time, then two HTTP puts will occur and the 2nd HTTP put will win. This scenario should be rare because of the cancellation tokens now being checked, but I could see a case where the propagation of the cancellation token is not fast enough. If this scenario ever occurs we throw an exception.

### DIFF
--- a/src/console/Microsoft.Health.Fhir.Ingest.Console.Common/FhirTransformation/Processor.cs
+++ b/src/console/Microsoft.Health.Fhir.Ingest.Console.Common/FhirTransformation/Processor.cs
@@ -45,17 +45,19 @@ namespace Microsoft.Health.Fhir.Ingest.Console.FhirTransformation
             EventMetrics.SetConnectorOperation(ConnectorOperation.FHIRConversion);
         }
 
-        public async Task ConsumeAsync(IEnumerable<IEventMessage> events)
+        public async Task ConsumeAsync(IEnumerable<IEventMessage> events, CancellationToken ct)
         {
             EnsureArg.IsNotNull(events);
 
-            var policyResult = await _retryPolicy.ExecuteAndCaptureAsync(async () => await ConsumeAsyncImpl(events, _templateManager.GetTemplateAsString(_templateName)));
+            var policyResult = await _retryPolicy.ExecuteAndCaptureAsync(async () => await ConsumeAsyncImpl(events, _templateManager.GetTemplateAsString(_templateName), ct));
 
             // This is a fallback option to skip any bad messages.
             // In known cases, the exception would be caught earlier and logged to the error message service.
             // If processing reaches this point in the code, the exception is unknown and retry attempts have failed.
             // If the error message service is enabled, the expectation is to log an error, and all events, and move on.
-            if (_errorMessageService != null && policyResult.FinalException != null)
+            if (_errorMessageService != null
+                && policyResult.FinalException != null
+                && policyResult.FinalException is not TaskCanceledException)
             {
                 policyResult.FinalException.AddEventContext(events);
                 var errorMessage = new IomtErrorMessage(policyResult.FinalException);
@@ -64,15 +66,21 @@ namespace Microsoft.Health.Fhir.Ingest.Console.FhirTransformation
 
         }
 
-        private async Task ConsumeAsyncImpl(IEnumerable<IEventMessage> events, string templateContent)
+        private async Task ConsumeAsyncImpl(IEnumerable<IEventMessage> events, string templateContent, CancellationToken ct)
         {
-            await _measurementImportService.ProcessEventsAsync(events, templateContent, _logger).ConfigureAwait(false);
+            await _measurementImportService.ProcessEventsAsync(events, templateContent, _logger, ct).ConfigureAwait(false);
         }
 
         private static AsyncPolicy CreateRetryPolicy(ITelemetryLogger logger, IErrorMessageService? errorMessageService)
         {
             bool ExceptionRetryableFilter(Exception ee)
             {
+                if (ee is TaskCanceledException)
+                {
+                    logger.LogTrace($"The task was cancelled.");
+                    return false;
+                }
+
                 logger.LogTrace($"Encountered retryable exception {ee.GetType()}");
                 logger.LogError(ee);
                 TrackExceptionMetric(ee, logger);
--- a/src/console/Microsoft.Health.Fhir.Ingest.Console.Common/FhirTransformation/Processor.cs
+++ b/src/console/Microsoft.Health.Fhir.Ingest.Console.Common/FhirTransformation/Processor.cs
@@ -77,7 +77,7 @@ namespace Microsoft.Health.Fhir.Ingest.Console.FhirTransformation
             {
                 if (ee is TaskCanceledException)
                 {
-                    logger.LogTrace($"The task was cancelled.");
+                    logger.LogTrace($"The FHIR import was aborted due to task cancellation.");
                     return false;
                 }
 
--- a/src/lib/Microsoft.Health.Events/EventCheckpointing/StorageCheckpointClient.cs
+++ b/src/lib/Microsoft.Health.Events/EventCheckpointing/StorageCheckpointClient.cs
@@ -56,10 +56,11 @@ namespace Microsoft.Health.Events.EventCheckpointing
             return _storageClient;
         }
 
-        public async Task UpdateCheckpointAsync(Checkpoint checkpoint)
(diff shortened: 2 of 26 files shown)


dustinburson (approved):
Overall looks good. Made a few suggestions.

src/console/Microsoft.Health.Fhir.Ingest.Console.Common/FhirTransformation/Processor.cs
```diff
@@ -64,15 +66,21 @@ public async Task ConsumeAsync(IEnumerable<IEventMessage> events)
...
+            await _measurementImportService.ProcessEventsAsync(events, templateContent, _logger, ct).ConfigureAwait(false);
         }
 
         private static AsyncPolicy CreateRetryPolicy(ITelemetryLogger logger, IErrorMessageService? errorMessageService)
         {
             bool ExceptionRetryableFilter(Exception ee)
             {
+                if (ee is TaskCanceledException)
+                {
+                    logger.LogTrace($"The task was cancelled.");
```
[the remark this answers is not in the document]
rogordon01: Are there any `TaskCanceledExceptions` that we'd want to retry? For example, I assume that we have a timeout set when communicating with the FhirService. If those calls timed out and threw a `TaskCanceledException` I think that we'd want to retry that specific call. 

Or would we simply restart processing by re-reading the message from the Eventhub and trying again?

src/lib/Microsoft.Health.Events/EventCheckpointing/StorageCheckpointClient.cs
```diff
@@ -152,7 +152,7 @@ public async Task<Checkpoint> GetCheckpointForPartitionAsync(string partitionIde
             }
         }
 
-        public async Task SetCheckpointAsync(IEventMessage eventArgs, IEnumerable<KeyValuePair<Metric, double>> metrics = null)
+        public async Task SetCheckpointAsync(IEventMessage eventArgs, CancellationToken ct, IEnumerable<KeyValuePair<Metric, double>> metrics = null)
```
rogordon01, line 155: Shouldn't this cancellation token get passed into the call to `UpdateCheckpointAsync`? Where is it being used?

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementFhirImportService.cs
```diff
@@ -41,10 +42,10 @@ public async Task ProcessStreamAsync(Stream data, string templateDefinition, ITe
             var template = BuildTemplate(templateDefinition, log);
             var measurementGroups = await ParseAsync(data, log).ConfigureAwait(false);
 
-            await ProcessMeasurementGroups(measurementGroups, template, log).ConfigureAwait(false);
+            await ProcessMeasurementGroups(measurementGroups, template, log, CancellationToken.None).ConfigureAwait(false);
```
rogordon01, line 45: Is this meant to be `CancelationToken.None`? Should it be a cancellation token that is passed into `ProcessStreamAsync`?

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementGroupFhirImportService.cs
```diff
@@ -41,10 +42,10 @@ public async Task ProcessStreamAsync(Stream data, string templateDefinition, ITe
             var template = BuildTemplate(templateDefinition, log);
             var measurementGroups = await ParseAsync(data, log).ConfigureAwait(false);
 
-            await SendMeasurementGroups(measurementGroups, template, log).ConfigureAwait(false);
+            await SendMeasurementGroups(measurementGroups, template, log, CancellationToken.None).ConfigureAwait(false);
```
rogordon01, line 45: Is this meant to be `CancellationToken.None`?

src/console/Microsoft.Health.Fhir.Ingest.Console.Common/FhirTransformation/Processor.cs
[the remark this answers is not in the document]
dustinburson: In general I can't think of any right now.  A task cancelled is usually the signal for aborting the current process.

src/lib/Microsoft.Health.Events/EventCheckpointing/StorageCheckpointClient.cs
```diff
@@ -74,6 +75,8 @@ public async Task UpdateCheckpointAsync(Checkpoint checkpoint)
 
             try
             {
+                ct.ThrowIfCancellationRequested();
+
                 try
                 {
                     await blobClient.SetMetadataAsync(metadata);
```
rogordon01, line 82: Continue passing the token into the below async methods. This will give them the chance to stop operations on a cancellation request.

src/lib/Microsoft.Health.Events/EventCheckpointing/StorageCheckpointClient.cs
```diff
@@ -74,6 +75,8 @@ public async Task UpdateCheckpointAsync(Checkpoint checkpoint)
 
             try
             {
+                ct.ThrowIfCancellationRequested();
```
rogordon01, line 78: If we keep passing the token into the below async methods do we need to explicitly call this?

src/lib/Microsoft.Health.Events/EventCheckpointing/StorageCheckpointClient.cs
wi-y: yep thanks

src/lib/Microsoft.Health.Events/EventCheckpointing/StorageCheckpointClient.cs
wi-y: good point. I don't think so. removed it

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementFhirImportService.cs
wi-y: This particular method is the legacy azure functions code path which we'll be deprecating soon so I didn't spend much time on it and just wanted to keep the current state behavior

src/lib/Microsoft.Health.Events/EventCheckpointing/StorageCheckpointClient.cs
wi-y: yeah, thanks Rob. I updated it

rogordon01 (approved):

src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4FhirImportService.cs
```diff
@@ -117,7 +122,15 @@ public virtual async Task<string> SaveObservationAsync(ILookupTemplate<IFhirTemp
                     if (existingObservation == null)
```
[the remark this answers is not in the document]
wi-y: added

src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4FhirImportService.cs
```diff
@@ -117,7 +122,15 @@ public virtual async Task<string> SaveObservationAsync(ILookupTemplate<IFhirTemp
                     if (existingObservation == null)
                     {
                         var newObservation = GenerateObservation(config, observationGroup, identifier, ids);
-                        return (await _fhirService.CreateResourceAsync(newObservation).ConfigureAwait(false), ResourceOperation.Created);
+
+                        var result = await _fhirService.UpdateResourceAsync(newObservation).ConfigureAwait(false);
+
+                        if (result.VersionId != "1")
+                        {
+                            _logger.LogError(new Exception($"Two processes modified the same Obersvation: {result.Id}"));
```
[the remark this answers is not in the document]
wi-y: ITelemetryLogger doesn't support warnings so I left it as an exception. Added a metric though

PR #260, opened 2023-02-09
Add freshness delay metric

ShaunDonn:
Added freshness delay metric and logic for it to be invoked in the case of a healthy service (i.e. if a large freshness delay were to be reported after these changes, we would expect it to be due to a large volume of events, rather than an issue with the service itself). The next iteration of this work will involve adding logic to invoke this metric in the case where processing fails, leading to a backlog of events to process.

**Testing**
Provisioned a MedTech service in my personal cluster and observed the metric flowing through as expected.

### DIFF
--- a/src/lib/Microsoft.Health.Events/EventConsumers/Service/EventBatchingService.cs
+++ b/src/lib/Microsoft.Health.Events/EventConsumers/Service/EventBatchingService.cs
@@ -263,6 +263,8 @@ namespace Microsoft.Health.Events.EventConsumers.Service
             // If no events were flushed for the partition (eg: trigger reason is ThresholdWaitReached - due to receival of MaxTimeEvent), then use the current timestamp.
             var eventTimestampLastProcessed = events?.Any() ?? false ? events.Last().EnqueuedTime.UtcDateTime : DateTime.UtcNow;
             _logger.LogMetric(EventMetrics.EventTimestampLastProcessedPerPartition(partitionId, triggerReason), double.Parse(eventTimestampLastProcessed.ToString("yyyyMMddHHmmss")));
+
+            _logger.LogMetric(EventMetrics.EventFreshnessDelayPerPartition(partitionId, triggerReason), DateTime.UtcNow.Subtract(eventTimestampLastProcessed).TotalMinutes);
         }
     }
 }
--- a/src/lib/Microsoft.Health.Events/EventConsumers/Service/EventBatchingService.cs
+++ b/src/lib/Microsoft.Health.Events/EventConsumers/Service/EventBatchingService.cs
@@ -264,7 +264,7 @@ namespace Microsoft.Health.Events.EventConsumers.Service
             var eventTimestampLastProcessed = events?.Any() ?? false ? events.Last().EnqueuedTime.UtcDateTime : DateTime.UtcNow;
             _logger.LogMetric(EventMetrics.EventTimestampLastProcessedPerPartition(partitionId, triggerReason), double.Parse(eventTimestampLastProcessed.ToString("yyyyMMddHHmmss")));
 
-            _logger.LogMetric(EventMetrics.EventFreshnessDelayPerPartition(partitionId), DateTime.UtcNow.Subtract(eventTimestampLastProcessed).TotalMinutes);
+            _logger.LogMetric(EventMetrics.EventFreshnessDelayPerPartition(partitionId), DateTime.UtcNow.Subtract(eventTimestampLastProcessed).TotalMilliseconds);
         }
     }
 }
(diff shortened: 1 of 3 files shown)


src/lib/Microsoft.Health.Events/Telemetry/Metrics/EventMetrics.cs
```diff
@@ -122,6 +122,19 @@ public static Metric EventTimestampLastProcessedPerPartition(string partitionId,
                 .AddDimension(_reasonDimension, triggerReason);
         }
 
+        /// <summary>
+        /// Signals the amount of time that has passed since the timestamp corresponding to the last processed event per partition. Calculated by subtracting the timestamp of the last processed event from the current time. If there are no events to be processed, the delay is 0.
+        /// </summary>
+        /// <param name="partitionId">The partition id of the event hub</param>
+        /// <param name="triggerReason">The trigger that caused the events to be flushed and processed</param
```
pallar-ms, line 129: I'm not sure we need the "triggerReason" for this metric as it doesn't really fit the description of this metric, but that is just my opinion. No harm having it, if we feel like it could help with investigations to know whether we are !delayed because we received timer events or not.

src/lib/Microsoft.Health.Events/EventConsumers/Service/EventBatchingService.cs
```diff
@@ -263,6 +263,8 @@ private void LogDataFreshness(string partitionId, string triggerReason, IEnumera
             // If no events were flushed for the partition (eg: trigger reason is ThresholdWaitReached - due to receival of MaxTimeEvent), then use the current timestamp.
             var eventTimestampLastProcessed = events?.Any() ?? false ? events.Last().EnqueuedTime.UtcDateTime : DateTime.UtcNow;
             _logger.LogMetric(EventMetrics.EventTimestampLastProcessedPerPartition(partitionId, triggerReason), double.Parse(eventTimestampLastProcessed.ToString("yyyyMMddHHmmss")));
+
+            _logger.LogMetric(EventMetrics.EventFreshnessDelayPerPartition(partitionId, triggerReason), DateTime.UtcNow.Subtract(eventTimestampLastProcessed).TotalMinutes);
```
pallar-ms, line 267: As discussed offline, we need to change where this gets emitted or add another point of emission for when the events fail to process.

ShaunDonn: Further discussed offline; will add point of emission for failed processing case in second iteration.

src/lib/Microsoft.Health.Events/Telemetry/Metrics/EventMetrics.cs
ShaunDonn: I thought about this some more and I agree. Based on the definition of this metric and depending on how we use it going forward, its invocation doesn't necessarily have to be tied to a flushing event, so the triggerReason may not be relevant in every case. I added changes to remove this dimension.

src/lib/Microsoft.Health.Events/EventConsumers/Service/EventBatchingService.cs
```diff
@@ -263,6 +263,8 @@ private void LogDataFreshness(string partitionId, string triggerReason, IEnumera
             // If no events were flushed for the partition (eg: trigger reason is ThresholdWaitReached - due to receival of MaxTimeEvent), then use the current timestamp.
             var eventTimestampLastProcessed = events?.Any() ?? false ? events.Last().EnqueuedTime.UtcDateTime : DateTime.UtcNow;
             _logger.LogMetric(EventMetrics.EventTimestampLastProcessedPerPartition(partitionId, triggerReason), double.Parse(eventTimestampLastProcessed.ToString("yyyyMMddHHmmss")));
+
+            _logger.LogMetric(EventMetrics.EventFreshnessDelayPerPartition(partitionId), DateTime.UtcNow.Subtract(eventTimestampLastProcessed).TotalMinutes);
```
dustinburson, line 267: I think by convention metrics we expose to customers that are time based should be in milliseconds.

dustinburson: @wi-y to confirm but it looks like based on https://1dsdocs.azurewebsites.net/articles/shoebox/shoebox-metrics.html#guidelines-for-units valid units for time should either be milliseconds or seconds.  Suggestion change to total seconds (that should be precise enough, I don't think we need milliseconds).

dustinburson (changes requested):

src/lib/Microsoft.Health.Events/EventConsumers/Service/EventBatchingService.cs
wi-y: For Latency metrics I believe only Milliseconds are supported based on a later section of the doc, but it may be worth asking the internal team if Seconds are supported since that seems easier to read for a user.

ShaunDonn: Sent an email to the Shoebox Core group mentioned in the article asking for an exception to the rule that latency metrics must have milliseconds as their unit.

dustinburson (approved):

PR #262, opened 2023-02-15
Fix event/observation updates being dropped

pallar-ms:
Changes include two fixes:
* When multiple messages that contain data for the same Observation have the same timestamp, the measurement data in the first message processed is the only data persisted. 
Fix includes taking the latest event data for the same timestamp within the batch and only adding the corresponding measurement in the measurement group.

* When merging sampled data (new data with an existing observation), if the timestamp between the new and the existing value collides, the existing value was being taken which resulted in the merged observation not being considered for an 'update'. 
Fix includes taking the new data value if there is timestamp collision so that merged observation represents latest data.

### DIFF
--- a/src/lib/Microsoft.Health.Fhir.Ingest/Data/SampledDataProcessor.cs
+++ b/src/lib/Microsoft.Health.Fhir.Ingest/Data/SampledDataProcessor.cs
@@ -67,38 +67,38 @@ namespace Microsoft.Health.Fhir.Ingest.Data
             return timeValues.ToArray();
         }
 
-        public virtual (DateTime Time, string Value)[] MergeData((DateTime Time, string Value)[] data1, (DateTime Time, string Value)[] data2)
+        public virtual (DateTime Time, string Value)[] MergeData((DateTime Time, string Value)[] existingData, (DateTime Time, string Value)[] newData)
         {
-            EnsureArg.IsNotNull(data1, nameof(data1));
-            EnsureArg.IsNotNull(data2, nameof(data2));
+            EnsureArg.IsNotNull(existingData, nameof(existingData));
+            EnsureArg.IsNotNull(newData, nameof(newData));
 
-            var output = new List<(DateTime Time, string Value)>(data1.Length + data2.Length);
-            var data1Pos = 0;
-            var data2Pos = 0;
+            var output = new List<(DateTime Time, string Value)>(existingData.Length + newData.Length);
+            var existingDataPos = 0;
+            var newDataPos = 0;
 
-            while (data1Pos < data1.Length || data2Pos < data2.Length)
+            while (existingDataPos < existingData.Length || newDataPos < newData.Length)
             {
-                if (data1Pos >= data1.Length)
+                if (existingDataPos >= existingData.Length)
                 {
-                    output.Add(data2[data2Pos++]);
+                    output.Add(newData[newDataPos++]);
                 }
-                else if (data2Pos >= data2.Length)
+                else if (newDataPos >= newData.Length)
                 {
-                    output.Add(data1[data1Pos++]);
+                    output.Add(existingData[existingDataPos++]);
                 }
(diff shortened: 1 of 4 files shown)


src/lib/Microsoft.Health.Fhir.Ingest/Data/SampledDataProcessor.cs
```diff
@@ -67,38 +67,38 @@ public virtual (DateTime Time, string Value)[] SampledDataToTimeValues(string st
...
                 }
                 else
                 {
-                    // Collision, take one and increment both
-                    output.Add(data1[data1Pos++]);
-                    data2Pos++;
+                    // In case of collision, take the new data to represent latest data, and increment both positions.
+                    output.Add(newData[newDataPos++]);
+                    existingDataPos++;
                 }
```
pallar-ms, line 102: Note: the real fix is in this snippet '99-101' to switch to the new data (data2) value being taken instead of existing (data1).
The remaining changes in this function is just var renames.

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementFhirImportService.cs
```diff
@@ -198,7 +198,10 @@ private static (IEnumerable<IMeasurementGroup> groups, Dictionary<IMeasurement,
             .GroupBy(m => $"{m.DeviceId}-{m.Type}-{m.PatientId}-{m.EncounterId}-{m.CorrelationId}")
             .Select(g =>
             {
-                IList<Measurement> measurements = g.ToList();
+                // In case of multiple measurements with the same OccurrenceTimeUtc within the batch, take the latest (based on IngestionTimeUtc) measurement
+                // to represent the recent/updated event data at that occurence timestamp.
```
dustinburson, line 202: Nit: Spelling, occurrence is in missing an r.

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementFhirImportService.cs
```diff
@@ -198,7 +198,10 @@ private static (IEnumerable<IMeasurementGroup> groups, Dictionary<IMeasurement,
             .GroupBy(m => $"{m.DeviceId}-{m.Type}-{m.PatientId}-{m.EncounterId}-{m.CorrelationId}")
             .Select(g =>
             {
-                IList<Measurement> measurements = g.ToList();
+                // In case of multiple measurements with the same OccurrenceTimeUtc within the batch, take the latest (based on IngestionTimeUtc) measurement
+                // to represent the recent/updated event data at that occurence timestamp.
+                IList<Measurement> measurements = g.GroupBy(x => x.OccurrenceTimeUtc).Select(y => y.OrderByDescending(d => d.IngestionTimeUtc).First()).ToList();
```
dustinburson, line 203: Suggestion:  I think this can be replaced with ``.Select(y => y.Max(d => d.IngestionTimeUtc)``` and will be slightly more efficient.  In 99.99%+ of cases it won't make a difference because there will be 1 or 2 values max in the group, but if there is ever a larger group you avoid needing to do the sort when all you want is the max value in the collection.

dustinburson (approved):

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementFhirImportService.cs
pallar-ms: Yeah I had actually tried that first, but that returns the max value and not the entry corresponding to the max value. MaxBy() would actually do that, but that is only supported in >.net6

dustinburson: Thanks for the clarification.  We will be looking to upgrade to .NET 7 soon.  At that time we can look to optimize.  Thanks!

PR #263, opened 2023-02-27
Enable MeasurementToFhir class that supports compression

wi-y:
In order to support compressed measurement groups sent by the normalization service (future), we will be enabling the MeasurementGroupFhirImportService by default which can decompress the measurement groups and save them to FHIR. The MeasurementGroupFhirImportService class is backwards compatible so it can support regular measurements (current state) in addition to measurement groups and compressed measurement groups (future state).

We should also adjust the batch sizes before enabling compression. Currently with measurements the max batch size is 300. But now with have measurement groups with many measurements, so if we keep the batch size of 300 then memory usage will be increased since 1 measurement group = 1 event instead of 1 measurement = 1 event.


src/console/Microsoft.Health.Fhir.Ingest.Console.Common/Extensions/FhirTransformationExtensions.cs
```diff
@@ -78,7 +79,7 @@ public static IServiceCollection AddFhirImportServices(this IServiceCollection s
 
             services.AddSingleton<IExceptionTelemetryProcessor, FhirExceptionTelemetryProcessor>();
             services.AddSingleton<MeasurementFhirImportOptions>();
-            MeasurementImportServiceExtensions.AddImportService(services, config);
+            services.AddSingleton<IImportService, MeasurementGroupFhirImportService>();
```
dustinburson, line 82: So we are just swapping the MeasurementFhirImportService for the MeasurementGroupFhirImportService since the latter supports both, making the configuration redundant?

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementImportServiceExtensions.cs
```diff
@@ -1,30 +0,0 @@
...
-    public class MeasurementImportServiceExtensions
-    {
-        public static void AddImportService(IServiceCollection services, IConfiguration configuration)
-        {
-            var collectorOptions = new CollectorOptions();
-            configuration.GetSection(CollectorOptions.Settings).Bind(collectorOptions);
-
-            if (!collectorOptions.UseCompressionOnSend)
-            {
-                services.TryAddSingleton<IImportService, MeasurementFhirImportService>();
```
dustinburson, line 22: Is there any reason to keep the MeasurementFhirImportService?  It looks like it is used in the legacy Azure Functions but since the new service is backwards compatible it seems to me like we should be able swap them there too.  Thoughts?  If so it wouldn't be for this PR but want to see if we avoid maintaining both services long term.

dustinburson (approved):

src/console/Microsoft.Health.Fhir.Ingest.Console.Common/Extensions/FhirTransformationExtensions.cs
wi-y: yeah I figured since the new class supports both we don't really need the configuration. I think the configuration was added so that we could toggle off the new class while we were investigating the one issue that ended up being due to a break in the event hub sdk partition assignment code.

src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementImportServiceExtensions.cs
wi-y: Yeah we should deprecate MeasurementFhirImportService when we deprecate the legacy Azure Functions code, or just swap it out for the new class. Created user story 101315 under the Azure Functions / Stream Analytics deprecation feature 99669.

PR #265, opened 2023-03-13
Fix duplicate dimension being added

wi-y:
The FHIRResourceContention metric was adding a dimension called 'Name' via .AddDimension but the metric extension method .CreateBaseMetric also adds a dimension called 'Name', so the end result was the 'Name' dimension was added twice which resulted in an exception the second time 'Name' was added to the metric dimensions dictionary.

Updated the FHIRResourceContention metric to use a different dimension name for the FHIR resource so that the 'Name' dimension isn't added twice.

Also updated the metric extension class so that when .AddDimension is called and there is already a dimension with that name we no longer throw an exception and instead log an error trace. The dimension is not added/overwritten.

### DIFF
--- a/src/lib/Microsoft.Health.Common/Telemetry/MetricExtension.cs
+++ b/src/lib/Microsoft.Health.Common/Telemetry/MetricExtension.cs
@@ -3,7 +3,9 @@
 // Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // -------------------------------------------------------------------------------------------------
 
+using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using EnsureThat;
 
 namespace Microsoft.Health.Common.Telemetry
@@ -33,12 +35,19 @@ namespace Microsoft.Health.Common.Telemetry
 
         public static Metric AddDimension(this Metric metric, string dimensionName, string dimensionValue)
         {
-            if (string.IsNullOrWhiteSpace(dimensionValue))
+            if (string.IsNullOrWhiteSpace(dimensionName) || string.IsNullOrWhiteSpace(dimensionValue))
             {
                 return metric;
             }
 
-            metric.Dimensions[dimensionName] = dimensionValue;
+            try
+            {
+                metric.Dimensions.Add(dimensionName, dimensionValue);
+            }
+            catch (Exception ex)
+            {
+                Trace.TraceError($"An error occurred while adding dimension {dimensionName} to metric {metric.Name}. {ex}");
+            }
 
             return metric;
         }
--- a/src/lib/Microsoft.Health.Common/Telemetry/MetricExtension.cs
+++ b/src/lib/Microsoft.Health.Common/Telemetry/MetricExtension.cs
@@ -38,7 +38,8 @@ namespace Microsoft.Health.Common.Telemetry
                 return metric;
             }
 
-            metric.Dimensions.Add(dimensionName, dimensionValue);
+            metric.Dimensions[dimensionName] = dimensionValue;
+
             return metric;
         }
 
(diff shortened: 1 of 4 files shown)


src/lib/Microsoft.Health.Common/Telemetry/MetricExtension.cs
```diff
@@ -38,7 +38,8 @@ public static Metric AddDimension(this Metric metric, string dimensionName, stri
                 return metric;
             }
 
-            metric.Dimensions.Add(dimensionName, dimensionValue);
+            metric.Dimensions[dimensionName] = dimensionValue;
```
rogordon01, line 41: From the description of this bug, throwing the exception on duplicate dimensions seems like the correct thing to do. Otherwise the value would have been overridden which could have been hard to track down. 
If we really need the ability to override dimension names, I'd add in an optional parameter to this extension. Something like `overrideIfPresent`. And then do the override if the situation calls for it.

rogordon01 (changes requested):

src/lib/Microsoft.Health.Common/Telemetry/MetricExtension.cs
dustinburson: If that is the direction might be good to switch to TryAdd and if the add fails that will be the trigger to throw.

wi-y: Throwing an exception in this case is also pretty difficult to track down since the error occurs at runtime instead of compile time, and in this case the error went through the FhirServiceExceptionProcessor which made the error pretty unclear. In the case of adding a duplicate metric key, an ArgumentException was thrown and then in [FhirServiceExceptionProcessor](https://github.com/microsoft/iomt-fhir/blob/c8f841610dab63c59519c14cdf7553b6085d6d48/src/lib/Microsoft.Health.Extensions.Fhir.R4/Telemetry/Exceptions/FhirServiceExceptionProcessor.cs#L62) there is a specific case for ArgumentException that expects a .ParamName which was null and so it threw a NullReferenceException and hid the original exception/ problematic code. 

Since the error also potentially affects message processing/retries it seemed that risk of overwriting the dimension value by default was safer than throwing an exception. For the PR I also just changed the dimension name from 'Name' to 'ResourceType' for the problematic metric so there wouldn't be a duplicate anymore, but wanted to put in some guardrails for the future because the error was really hard to track down today and there wasn't anything from stopping it from happening. We can remove it if we think it is a worse outcome to overwrite than to throw by default

dustinburson (approved):

src/lib/Microsoft.Health.Common/Telemetry/MetricExtension.cs
dustinburson: Maybe we can at least do an assert or some kind of trace so we have something we can track down if this ends up causing problems in the future and we need to track it down?

rogordon01: Trace works for me.

rogordon01 (approved):

src/lib/Microsoft.Health.Common/Telemetry/MetricExtension.cs
wi-y: hmm I suppose a compromise would be to instead add a try/catch in .AddDimension with a Trace log whenever a dimension key that already exists is being added. However we don't have access to ITelemetryLogger in Microsoft.Health.Common where this MetricExtension lives. We could log to System.Console. Even though Console isn't ideal the error would at least be logged somewhere, the metric dimension would not be overwritten, and the message processing wouldn't be affected, which was one of my concerns. Any objections with that?

dustinburson: I think if you use one of the methods in the System.Diagnostics it is better than a straight Console.WriteLine.  The System.Diagnostics can be easily configured/rerouted (even to one of our ILoggers)

Example https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.trace.traceinformation?view=net-8.0

wi-y: thanks for the suggestion. updated .AddDimension to log an error trace when there is an error adding the dimension. the dimension is not added/overwritten

PR #267, opened 2023-03-16
Spell observation correctly

wi-y:


src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4FhirImportService.cs
```diff
@@ -130,7 +130,7 @@ public virtual async Task<string> SaveObservationAsync(ILookupTemplate<IFhirTemp
 
                             if (result.VersionId != "1")
                             {
-                                _logger.LogError(new Exception($"Two processes modified the same Obersvation: {result.Id}"));
+                                _logger.LogError(new Exception($"Two processes modified the same Observation: {result.Id}"));
```
rogordon01, line 133: I vote that we leave it as-is :)

PR #268, opened 2023-03-29
Partition locking processor prototype

wi-y:
This is an initial proof of concept event hub processor that locks the partitions in advance and only reads from the partitions it has locked. The processor is smart enough to detect when other processors have been added or removed and will stop and recalculate which partitions it should be assigned and attempt to lock them.

This new processor is not enabled by default. It is a strategy we could take if we need this locking mechanism for our fhirtransformation pod to prevent multiple processors from reading from the same partition at the same time.

The original idea for the locking processor was sort of derived from an event hub sdk sample [here](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/eventhub/Azure.Messaging.EventHubs/samples/Sample08_CustomEventProcessor.md). In the sample it was suggested that Kubernetes or Service Fabric could pass in a static list of partitions to each processor and each processor could work on the list of partitions that it is assigned.


src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingProcessor.cs
```diff
@@ -0,0 +1,418 @@
...
+
+        private int _totalProcessorsRunning = 0;
+
+        // consider injecting these or querying event hub for max partitions
+        private int _maximumProcessors = 8;
+
+        private int _maximumPartitions = 8;
+
+        // settings that should be passed in as environment variables
+        private Uri _paritionManagerUri = new Uri("<blob uri goes here>");
```
dustinburson, line 52: Are you looking for sign off on the overall concept and then planning on doing a pass to wire these up before committing?

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingProcessor.cs
```diff
@@ -0,0 +1,418 @@
...
+        private string _processorId;
+
+        private List<int> _ownedPartitions;
+
+        private BlobContainerClient _containerClient;
+
+        private int _totalProcessorsRunning = 0;
+
+        // consider injecting these or querying event hub for max partitions
+        private int _maximumProcessors = 8;
```
dustinburson, line 47: Shouldn't be able to get this from the event hub?

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingProcessor.cs
```diff
@@ -0,0 +1,418 @@
...
+                await CheckProcessorCount(cts);
+                await StartStaticPartitionProcessor(cts.Token, _ownedPartitions);
+            }
+            catch (TaskCanceledException)
+            {
+                Console.WriteLine("A task was cancelled");
+            }
+        }
+
+        public async Task RunProcessor(CancellationTokenSource cts)
```
dustinburson, line 285: The cancellation token should be wired through the stack and used with Task.Delay, etc.

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingProcessor.cs
```diff
@@ -0,0 +1,418 @@
...
+        {
+            CreateContainerClient();
+            await RegisterProcessor();
+            await ContinuallyRegisterProcessor();
+
+            while (true)
+            {
+                var innerCts = new CancellationTokenSource();
```
dustinburson, line 293: Why aren't we using the original cts?

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingProcessor.cs
```diff
@@ -0,0 +1,418 @@
...
+                        availablePartitions.Remove(partitionId);
+                    }
+
+                    blobOwnerhip.Add(partitionId, ownership);
+                }
+
+                // if not available partitions then return and wait
+                if (availablePartitions.Count == 0)
+                {
+                    Console.WriteLine("No partitions are available... Waiting");
```
dustinburson, line 164: We would need to properly implement our logging.

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingProcessor.cs
```diff
@@ -0,0 +1,418 @@
...
+                                { "owner", _processorId },
+                                { "owneduntil", DateTime.UtcNow.AddMinutes(1).ToString() },
+                            },
+                            Conditions = new BlobRequestConditions()
+                            {
+                                IfMatch = eTag,
+                            },
+                        };
+
+                        var ownershipResult = await ownershipClient.UploadAsync(BinaryData.FromString(string.Empty), ownershipClientOptions);
```
dustinburson, line 362: Do we have to upload the blob again to refresh ownership?  It looks like the lease blob might be better for this, https://learn.microsoft.com/en-us/rest/api/storageservices/lease-blob

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingProcessor.cs
rogordon01: If you need a different CancellationTokenSource you should link it to the incoming source unless there is a reason not too. That way outer source can still cancel child operations.

`using var tokenSource = CancellationTokenSource.CreateLinkedTokenSource(stoppingToken);`

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingProcessor.cs
wi-y: Yeah mainly wanted to stop here and validate the concept before proceeding with additional items like dependency injection, logging, production readiness etc so that if we prefer to go another direction then it wouldn't be throw away work.

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingProcessor.cs
wi-y: agree, leases would be better. can switch to those

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingProcessor.cs
wi-y: in this implementation we have an inner loop that needs to be cancelled when we detect that the number of processors changes. thanks Rob for the suggestion - that is exactly what I should use

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingProcessor.cs
wi-y: yes added a comment above for querying this from event hub, but for the proof of concept it was hardcoded

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingProcessor.cs
```diff
@@ -0,0 +1,418 @@
...
+            if (activePodCount == 0)
+            {
+                throw new Exception("active pods is 0");
+            }
+
+            // divide the partitions among the pods
+            // todo: handle cases where there is a remainder
+            // for example 8 partitions and 6 pods, which would require 4 pods to have 1 partition, and 2 pods to have 2 partitions
+            // for now, we will round down
+            var suggestedPartitionsPerPod = Math.Floor((double)_maximumProcessors / activePodCount);
```
rogordon01, line 120: Conceptually this PR looks good. One thing to consider is that there will be cases when there are more active pods then `_maximumProcessors`. For example, during an upgrade K8s first creates new, upgraded pods and then terminates the old pods. So in that brief period of time there will not be enough partitions to go around. Not sure if this logic handles that case

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingProcessor.cs
```diff
@@ -0,0 +1,418 @@
...
+
+// todos:
+// - detect if a partition has no ownership updates for some time
+// - handle cases where the ratio of processors to partitions is not divided evenly
+// - detect if acquiring partitions takes too long
+// - replace Task.Delay with something more sophisticated
+
+namespace Microsoft.Health.Events.EventHubProcessor
+{
+    public class PartitionLockingProcessor
```
rogordon01, line 36: The PartitionAssignment logic here seems to be separate from the actual EventProcessor/Service. What would you think about separating these processes out completely? I'm thinking of something like:

1. Run Partition Locking Processor as a Background Service inside of our Keda External Scaler.
  1. Responsible for maintaining ownership metadata on blob based on Active Processors
1. Run Loop within each service to maintain Active Processor blobs
1. Run StaticPartitionProcessor within each service.
   1. This periodically acquires the partitions to work on from blob storage, which is managed by the Partition Locking Processor running inside of Keda   

One downside to this approach is that if we scaled our services outside of Keda then this wouldn't work. 

Interested to hear your thoughts on this.

wi-y: It is a good suggestion. I am leaning towards running the partition assignment background service as part of the application code because we need a mechanism to restart the application loop if the number of processors change. We could create a separate service for that but for simplicity it is probably easier to include it as part of the partition assignment. I also like that it allows us to still scale using Keda or scale without using Keda

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingProcessor.cs
wi-y: it is a good call out. I think the simplest approach is to still do the calculation to assign a suggested number of partitions per pod, and if it is zero for some amount of time due to an upgrade then that is fine. As long as we expect the number of pods to eventually reach the desired state (say 8), then when we periodically recalculate the number of suggested partitions per pod there should be enough to go around.

we may want to add an error condition for certain cases. like if there is one partition and we scale to 8 that could mean that no pods are assigned the one partition since we round down here. we probably shouldn't scale to 8 in that case either but the application should alert as well

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingProcessor.cs
wi-y: added logging

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingProcessor.cs
wi-y: switched to blob leases for ownership, thanks for the suggestion

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingProcessor.cs
wi-y: added CancellationTokenSource.CreateLinkedTokenSource. thanks for the suggestion Rob

src/lib/Microsoft.Health.Events/EventHubProcessor/StaticPartitionProcessor.cs
```diff
@@ -0,0 +1,215 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Azure.Messaging.EventHubs;
+using Azure.Messaging.EventHubs.Primitives;
+using Azure.Messaging.EventHubs.Processor;
+using Azure.Storage.Blobs;
+
+namespace Microsoft.Health.Events.EventHubProcessor
+{
+    public class StaticPartitionProcessor : PluggableCheckpointStoreEventProcessor<EventProcessorPartition>
```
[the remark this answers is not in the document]
wi-y: renamed it to AssignedPartitionProcessor but am open to other suggestions

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
```diff
@@ -0,0 +1,429 @@
...
+            await RegisterProcessor();
+            await ContinuallyRegisterProcessor();
+
+            while (true)
+            {
+                using var innerCts = CancellationTokenSource.CreateLinkedTokenSource(ct);
+                while (!innerCts.IsCancellationRequested)
+                {
+                    await Task.Delay(5000);
+                    _eventHubPartitions = await GetPartitionsIdsForEventHub();
```
wi-y, line 324: we should also restart if this value changes. I believe there are some tiers of event hub ([premium, dedicated](https://learn.microsoft.com/en-us/azure/event-hubs/dynamically-add-partitions)) where the number of partitions can be changed.

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
```diff
@@ -0,0 +1,429 @@
...
+
+        protected override async Task ExecuteAsync(CancellationToken ct)
+        {
+            CreateContainerClient();
+            await RegisterProcessor();
+            await ContinuallyRegisterProcessor();
+
+            while (true)
+            {
+                using var innerCts = CancellationTokenSource.CreateLinkedTokenSource(ct);
```
dustinburson, line 320: It seems like we need a task that is running to look for a change in the processors and will cancel the current loop (inner cts) and retrigger the process.

dustinburson (changes requested):

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
wi-y: Yeah I think this should be accounted for. In PartitionLockingBackgroundService we have a background task CheckProcessorCount which periodically checks the number of active processor and if the number has changed then we should cancel the current loop and restart.

src/lib/Microsoft.Health.Events/EventHubProcessor/ResumableAssignedPartitionProcessor.cs
```diff
@@ -0,0 +1,51 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Health.Events.EventHubProcessor
+{
+    public class ResumableAssignedPartitionProcessor : IResumableEventProcessor
```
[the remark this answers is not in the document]
wi-y: This service was not something I would have normally created but I think it (or something like it) may be necessary to support both PaaS and OSS codeflows. In PaaS we have a health checker that stops and restarts an IResumableEventProcessor when the health checks succeed or fail. Even though AssignedPartitionProcessor could itself be a IResumableEventProcessor we don't really want to stop and restart that in PaaS because the AssignedPartitionProcessor needs some information in order to run and it gets that information from the outer loop (PartitionLockingBackgroundService) which is responsible for registering the processor and also claiming the partitions that the AssignedPartitionProcessor should read from.

I agree that it is weird that a background service wrapping another background service. If I were just coding for OSS then we only need PartitionLockingBackgroundService and AssignedPartitionProcessor but to support the health checker logic we have in PaaS we kind of need another abstraction that implements IResumableEventProcessor which is what I did here with ResumableAssignedPartitionProcessor, or we need a bit of a refactor. I am open to refactoring more but wanted to explain the thought process.

src/lib/Microsoft.Health.Events/EventHubProcessor/AssignedPartitionProcessor.cs
```diff
@@ -0,0 +1,226 @@
...
+                if (events.Count() == 0)
+                {
+                    var maxWaitEvent = new MaximumWaitEvent(partition.PartitionId, DateTime.UtcNow);
+                    await _eventConsumerService.ConsumeEvent(maxWaitEvent, cancellationToken);
+                    return;
+                }
+
+                foreach (var currentEvent in events)
+                {
+                    var eventMessage = new EventMessage(
```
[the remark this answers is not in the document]
wi-y: Yeah pretty much so we don't have to write classes for the different event types that are passed in by the SDKs. Even on the same SDK though there are different event types passed in. Here in `AssignedPartitionProcessor : PluggableCheckpointStoreEventProcessor<EventProcessorPartition>` it uses EventData. But BaseEventProcessor which is on the same SDK uses ProcessEventArgs

src/lib/Microsoft.Health.Events/EventHubProcessor/ResumableAssignedPartitionProcessor.cs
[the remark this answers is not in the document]
wi-y: added a flowchart: https://github.com/microsoft/iomt-fhir/blob/personal/wiyochum/partition-locking-prototype/docs/PartitionLocking.md

src/lib/Microsoft.Health.Events/EventHubProcessor/AssignedPartitionProcessor.cs
```diff
@@ -0,0 +1,226 @@
...
+                Task.FromResult(
+                    _assignedPartitions.Select(partition =>
+                        new EventProcessorPartitionOwnership
+                        {
+                            FullyQualifiedNamespace = FullyQualifiedNamespace,
+                            EventHubName = EventHubName,
+                            ConsumerGroup = ConsumerGroup,
+                            PartitionId = partition.ToString(),
+                            OwnerIdentifier = Identifier,
+                            LastModifiedTime = DateTimeOffset.UtcNow,
```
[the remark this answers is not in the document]
wi-y: the DateTimeOffset doesn't really matter. this was copied from the event hub example: https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/eventhub/Azure.Messaging.EventHubs/samples/Sample08_CustomEventProcessor.md

the event hub example basically uses this method it to trick the processor into thinking it has all the partitions it needs (aka there are no additional partitions to claim or to give up) but the exact value for LastModifiedTime is not really used at all. it is kind of weird but it seems to be the recommended way based on their example and so I copied it as-is

src/lib/Microsoft.Health.Events/EventHubProcessor/AssignedPartitionProcessor.cs
```diff
@@ -0,0 +1,226 @@
...
+                }));
+
+        protected async override Task OnProcessingEventBatchAsync(
+            IEnumerable<EventData> events,
+            EventProcessorPartition partition,
+            CancellationToken cancellationToken)
+        {
+            try
+            {
+                if (events.Count() == 0)
```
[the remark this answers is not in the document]
wi-y: thanks Dustin, that makes sense. I updated the code. Instead of using instead of events.Count() the code now uses a bool which will get unset if there are actually events, and if there are not events then it is a wait event (empty event) which we can pass through to the batching service. Yes the timer triggers an empty event which is passed into this method.

src/lib/Microsoft.Health.Events/EventHubProcessor/AssignedPartitionProcessor.cs
```diff
@@ -0,0 +1,226 @@
...
+                        currentEvent.SequenceNumber,
+                        currentEvent.Offset,
+                        currentEvent.EnqueuedTime,
+                        currentEvent.Properties,
+                        currentEvent.SystemProperties);
+
+                    await _eventConsumerService.ConsumeEvent(eventMessage, cancellationToken);
+                }
+            }
+            catch (Exception ex)
```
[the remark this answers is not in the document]
wi-y: it really should never be hit since we try/catch the exceptions in our handler code and retry if appropriate. the event hub examples also squash it and have the follow comment so I think it is appropriate since we have try/catch and retry logic in the handler anyway. we could also opt to throw if think that crashing and restarting would potentially lead to a better outcome but I am not sure we know

            // It is very important that you always guard against exceptions in
            // your handler code; the processor does not have enough
            // understanding of your code to determine the correct action to take.
            // Any exceptions from your handlers go uncaught by the processor and
            // will NOT be redirected to the error handler.
            //
            // In this case, the partition processing task will fault and be restarted
            // from the last recorded checkpoint.

Console.WriteLine($"Exception while processing events: { ex }");

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
```diff
@@ -0,0 +1,429 @@
...
+        private EventHubClientOptions _eventHubClientOptions;
+
+        public PartitionLockingBackgroundService(
+            PartitionLockingBackgroundServiceOptions partitionLockingOptions,
+            EventHubClientOptions eventHubClientOptions,
+            EventBatchingService eventBatchingService,
+            StorageCheckpointClient checkpointClient,
+            ITelemetryLogger logger)
+        {
+            _processorId = Guid.NewGuid().ToString(); // consider using the pod name when running in AKS
```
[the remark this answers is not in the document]
wi-y: I was thinking that instead of generating a guid to identify the processor that we instead use the pod name as the identifier. I believe the benefit during a pod restart would be that the name would be the same (instead of generating a new guid) so there could be some continuity when claiming or reclaiming partitions. eventually with guid it will work itself out as the old pod guid will eventually go out of date in 1 minute and when we recalculate the total processors then everything would be up to date but if we reused the pod name then there would be better continuity

with guid we have:
- pod starts, guid A registered
- pod restarts, and starts guid B registered. guid A is not longer being pushed to storage
- pod with guid B temporarily thinks there are two pods, until some time elapses (we have a 1 minute offset to determine if a guid is no longer active)
- pod with guid B figures out there is only 1 pod since 1 guid is being updated

if the pod name is unchanging then we have:
- pod starts, pod name is registered
- pod restarts, pod name is the same and continued to be registered and the perceived number of pods remains the same.

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
```diff
@@ -0,0 +1,429 @@
...
+                        _logger.LogTrace($"Deleting blob {blobItem.Name}. It was last active {activeUntil}.");
+                        await containerClient.DeleteBlobAsync(blobItem.Name);
+                    }
+                }
+            }
+
+            return activePods.Count;
+        }
+
+        public async Task RunAndRenew(CancellationTokenSource cts)
```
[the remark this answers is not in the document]
wi-y: it should be private. thanks for calling that out, it has been updated to private

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
```diff
@@ -0,0 +1,429 @@
...
+            }
+        }
+
+        protected override async Task ExecuteAsync(CancellationToken ct)
+        {
+            CreateContainerClient();
+            await RegisterProcessor();
+            await ContinuallyRegisterProcessor();
+
+            while (true)
```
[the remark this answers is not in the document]
wi-y: yes, updated

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
```diff
@@ -0,0 +1,429 @@
...
+                maximumBatchSize,
+                eventHubConsumerGroup,
+                tokenCrendential,
+                eventHubName,
+                eventHubNamespaceFQDN);
+
+            try
+            {
+                await processor.StartProcessingAsync(ct);
+                await Task.Delay(Timeout.Infinite, ct);
```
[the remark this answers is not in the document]
wi-y: added a flowchart here: https://github.com/microsoft/iomt-fhir/blob/personal/wiyochum/partition-locking-prototype/docs/PartitionLocking.md but essentially when the AssignedPartitionProcessor starts we will cache off the number of known processors at the time. Every 30 seconds when we register the current processor we also check the storage account to see if the current number of total processors matches the number of processors that we have cached. If the number of processors differs from the cache then we restart.

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
```diff
@@ -0,0 +1,429 @@
...
+                    partitionIdIndex = 0;
+                    continue;
+                }
+
+                partitionIdIndex++;
+                await Task.Delay(1000);
+            }
+        }
+
+        private async Task ClaimPartitionUsingBlob(int partitionToClaim)
```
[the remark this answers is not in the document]
wi-y: agree I will look into adding this today

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
```diff
@@ -0,0 +1,429 @@
...
+            }
+
+            // divide the partitions among the active pods
+            // todo: handle cases where there is a remainder
+            // for example 8 partitions and 6 pods, which would require 4 pods to have 1 partition, and 2 pods to have 2 partitions
+            // for now, we will round down
+
+            _logger.LogTrace($"The Event Hub has {_eventHubPartitions.Length} partitions");
+
+            var suggestedPartitionsPerProcessor = Math.Floor((double)_eventHubPartitions.Length / activeProcessorCount);
```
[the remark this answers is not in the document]
wi-y: yeah I agree it would be better in scenarios where the number of partitions is not evenly divided by the number of pods. I will look into adding this today

src/console/Microsoft.Health.Fhir.Ingest.Console.Common/Extensions/FhirTransformationExtensions.cs
```diff
@@ -40,6 +39,8 @@ public static IServiceCollection AddNormalizedEventReader(this IServiceCollectio
                 // Add EventHubClientOptions (Used for EventProcessorClient - common to all applications)
                 var options = new EventHubClientOptions();
                 config.GetSection("NormalizationEventHub").Bind(options);
+                var tokenProvider = sp.GetService<IAzureCredentialProvider>();
+                options.EventHubTokenCredential = tokenProvider?.GetCredential();
```
dustinburson, line 43: Same as other comment, not sure about including tokens inside the options class.

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
```diff
@@ -0,0 +1,487 @@
...
+using EnsureThat;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Health.Events.Common;
+using Microsoft.Health.Events.EventCheckpointing;
+using Microsoft.Health.Events.EventConsumers.Service;
+using Microsoft.Health.Events.Telemetry;
+using Microsoft.Health.Events.Telemetry.Exceptions;
+using Microsoft.Health.Logging.Telemetry;
+
+// Description
```
dustinburson, line 25: Nit: Can use /* */ notations for a block of comments

src/lib/Microsoft.Health.Events/EventHubProcessor/ResumableAssignedPartitionProcessor.cs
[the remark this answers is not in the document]
dustinburson: I guess, my question then is what is the need for the PartitionLockingBackgroundService to be a background service?  Originally I was concerned we had two places where the PartitionLockingBackgroundService is started/controlled.  One in the resumable processor and one as a background service.  Upon further investigation I didn't find any place where we were invoking AddHostedService in the DI on the PartitionLockingBackgroundService which means it isn't running as a background service in the processor.  If we are just using the background service contract, I think we should create our own interface for the control flow and not reuse the background service construct since it can cause confusion.

src/lib/Microsoft.Health.Events/EventHubProcessor/ResumableAssignedPartitionProcessor.cs
```diff
@@ -0,0 +1,51 @@
...
+        private bool _isRunning = false;
+
+        public ResumableAssignedPartitionProcessor(PartitionLockingBackgroundService lockingBackgroundService)
+        {
+            _lockingBackgroundService = lockingBackgroundService;
+        }
+
+        public void Dispose()
+        {
+            // todo?
```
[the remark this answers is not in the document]
dustinburson: After further investigation, we don't appear to be registering the PartitionLockingBackgroundService as a hosted service, so we should definitely ensure it is properly stopped on dispose (though we will likely need to handle an exception for the service already being exposed in case it is disposed first).  We don't need to dispose since the DI should handle that for us.

src/lib/Microsoft.Health.Events/Telemetry/Metrics/EventMetricDefinition.cs
```diff
@@ -33,5 +33,7 @@ private EventMetricDefinition(string metricName)
         public static EventMetricDefinition DeviceIngressSizeBytes { get; } = new EventMetricDefinition(nameof(DeviceIngressSizeBytes));
 
         public static EventMetricDefinition MeasurementToFhirBytes { get; } = new EventMetricDefinition(nameof(MeasurementToFhirBytes));
+
+        public static EventMetricDefinition PartitionLastClaimedDelay { get; } = new EventMetricDefinition(nameof(PartitionLastClaimedDelay));
```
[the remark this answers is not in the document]
dustinburson: Alternatively, if in the alert configuration we can include a check for how long since the delay metrics was last emitted that could serve the purpose too.  We will need to think about how to ensure we don't get false positives for services that are deprovisioned.

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
```diff
@@ -0,0 +1,487 @@
...
+// - detect if acquiring partitions takes too long
+// - replace Task.Delay with something more sophisticated
+
+namespace Microsoft.Health.Events.EventHubProcessor
+{
+    public class PartitionLockingBackgroundService : BackgroundService
+    {
+        private string _processorId;
+
+        private List<int> _ownedPartitions;
```
dustinburson, line 46: Nit: Many of these fields can be marked readonly since they are never modified once assigned.

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
```diff
@@ -0,0 +1,487 @@
...
+
+// todos:
+// - detect if a partition has no ownership updates for some time
+// - handle cases where the ratio of processors to partitions is not divided evenly
+// - detect if acquiring partitions takes too long
+// - replace Task.Delay with something more sophisticated
+
+namespace Microsoft.Health.Events.EventHubProcessor
+{
+    public class PartitionLockingBackgroundService : BackgroundService
```
dustinburson, line 42: Question, the current logic before this maintained lease locks and partition watermarks in blob correct?  I am trying understand if there would be any net change in blob API calls.  My thinking is no since we are replacing the existing lease locks with this. Is that a correct understanding?

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
```diff
@@ -0,0 +1,487 @@
...
+                            await Task.Delay(5000, ct);
+                            partitionIdIndex = 0;
+                            continue;
+                        }
+
+                        continue;
+                    }
+                    else
+                    {
+                        throw;
```
dustinburson, line 187: Should we have a retry policy with some minor back off?  Trying to think if there are any transient exceptions that could get us into trouble here.  It looks like an exception here will kill the entire Execute block.   Have we verified this properly kills the process?

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
```diff
@@ -0,0 +1,487 @@
...
+                try
+                {
+                    await ClaimPartitionUsingBlob(partitionId);
+                }
+                catch (RequestFailedException ex)
+                {
+                    if (ex.ErrorCode == BlobErrorCode.LeaseAlreadyPresent)
+                    {
+                        _logger.LogTrace($"Failed to acquire partition {partitionId}");
+                        await Task.Delay(1000, ct);
```
dustinburson, line 171: Would prefer constants/configuration values for the wait times

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
```diff
@@ -0,0 +1,487 @@
...
+
+        public PartitionLockingBackgroundService(
+            PartitionLockingBackgroundServiceOptions partitionLockingOptions,
+            EventHubClientOptions eventHubClientOptions,
+            EventBatchingService eventBatchingService,
+            StorageCheckpointClient checkpointClient,
+            ITelemetryLogger logger)
+        {
+            _processorId = Guid.NewGuid().ToString(); // consider using the pod name when running in AKS
+            _ownedPartitions = new List<int>();
```
dustinburson, line 72: Concurrent Dictionary seems like a better structure.  There are multiple awaited tasks interacting on the collection.  The current order maybe fine but leaves us open to errors as the code changes.  Dictionary would allow you to avoid using contains on the list as well elsewhere in the code.

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
```diff
@@ -0,0 +1,487 @@
...
+        }
+
+        private async Task StartAssignedPartitionProcessor(CancellationToken ct, List<int> ownedPartitions)
+        {
+            var maximumBatchSize = 100;
+
+            string[] partitions = new string[ownedPartitions.Count];
+            for (var i = 0; i < ownedPartitions.Count; i++)
+            {
+                partitions[i] = ownedPartitions.ElementAt(i).ToString();
```
dustinburson, line 448: ElementAt enumerates the IEnumerable again. Use the indexer or can simplify with a Linq statement

``
string[] partitions = ownedPartitions.Select(i => i.ToString()).ToArray();
``

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
```diff
@@ -0,0 +1,487 @@
...
+        {
+            // get all the processors in active status
+            var activeProcessors = ListBlobs(_containerClient, "activepods");
+            var activeProcessorCount = await GetActiveProcessors(activeProcessors, _containerClient, ct);
+            _totalProcessorsRunning = activeProcessorCount;
+            _logger.LogTrace($"Active processor count: {activeProcessorCount}");
+
+            if (activeProcessorCount == 0)
+            {
+                throw new Exception("active processors is 0");
```
dustinburson, line 135: Should be a strongly typed/specific exception.

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
```diff
@@ -0,0 +1,487 @@
...
+                    }
+
+                    await Task.Delay(30000, cts.Token);
+                }
+            });
+
+            return Task.CompletedTask;
+        }
+
+        private Task CheckIfPartitionsAreUnclaimed(CancellationToken ct)
```
dustinburson, line 368: Every worker would be doing this correct?  Instead can we emit the delay when we renew?  The gap is if no one owns the partition, nothing will update the metric.  Might be worth investigating if our alerting can detect no events in a time period.

As mentioned in another comment, freshness (the time last renewed) would be better if it weren't for the metric limitations.

Alternatively, could introduce a singleton blob location for a leader and only the leader is responsible for recording the freshness while others just loop and try to claim ownership.

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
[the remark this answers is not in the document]
dustinburson: Thank you for the flowchart.  My main concern with this approach is if there is any change in ownership, everything restarts, likely other nodes as well unless the a processor can claim ownership before the others detect it.  That restart will include the entire loop evaluation as well.  It seems like we could get into a scenario where if things get slightly off, loss of ownership could cascade continuously and we never get into a state where all partitions are claimed.

Thoughts on explicitly tracking the partitions owned and assigning them their own processor and either canceling that processor when ownership is lost or adding a new one when a partition is claimed?

Also, if you have a link handy for the original event hub logic it would be useful to compare to.

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
```diff
@@ -0,0 +1,487 @@
...
+                    var activePods = ListBlobs(_containerClient, "activepods");
+                    var activeProcessors = await GetActiveProcessors(activePods, _containerClient, cts.Token);
+
+                    _logger.LogTrace($"CheckProcessorCount - active processors {activeProcessors}");
+
+                    if (activeProcessors != _totalProcessorsRunning)
+                    {
+                        _logger.LogTrace($"Detected that number of processors equals {activeProcessors} but this processor is running with {_totalProcessorsRunning} known processors");
+                        _logger.LogTrace("Restarting and recomputing");
+                        cts.Cancel();
```
dustinburson, line 358: Suggestion:  Add comment here to indicate this is where the loop is canceled.

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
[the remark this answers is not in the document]
dustinburson: Looking at the logic again, it is based solely on the number of processors, not a partition losing its lease so that seems more stable but seems likely for high churn, especially in scenarios where K8 is adding a new pod before taking one down.

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
dustinburson: It would also to be good to have a metric for restarts like this to track them easier.  The log traces are definitely helpful though.

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
```diff
@@ -0,0 +1,487 @@
...
+                    if (DateTime.Parse(activeUntil) > DateTime.UtcNow.AddMinutes(-1))
+                    {
+                        _logger.LogTrace($"Blob name: {blobItem.Name} is active");
+                        activePods.Add(blobItem.Name);
+                    }
+                    else
+                    {
+                        // todo: check the etag as a condition for delete
+                        _logger.LogTrace($"Deleting blob {blobItem.Name}. It was last active {activeUntil}.");
+                        await containerClient.DeleteBlobAsync(blobItem.Name);
```
dustinburson, line 291: We should use E-Tag to delete and handle/ignore conflicts.

dustinburson: This might be another case where we should elect a leader and only the leader can clean up old blobs.

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
dustinburson, line 42: @rogordon01 is there anyway with KEDA to pass in the total number of workers into the process?  I think if we did that model it would help eliminate a lot of complexity in the design. Then we wouldn't need to worry about determining the total number of processors.

The downside is this would likely only be applicable to our custom implementation and not apply to the generic KEDA scaler.

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
dustinburson: This applies to all locations where we Delay for X

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
```diff
@@ -0,0 +1,487 @@
...
+                        continue;
+                    }
+                    else
+                    {
+                        throw;
+                    }
+                }
+
+                // if not enough available partitions then return and wait
+                if (partitionId + 1 >= _eventHubPartitions.Length && _ownedPartitions.Count != suggestedPartitionsPerProcessor)
```
dustinburson, line 192: This means we could get stuck in this loop if we never get our suggested partitions?

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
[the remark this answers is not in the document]
dustinburson: One other side effect of having dedicated processor instances per partition is we would no longer be able to dynamically scale the buffer size up as number of partitions in the processor decreases.  We would have to look at other solutions if we make the change I suggested.

dustinburson (changes requested):
A couple of larger items for discussion and several smaller items to clean up.

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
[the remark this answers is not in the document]
dustinburson: Still a a solution for this, if there is a remainder, the remainder will go unclaimed.

src/lib/Microsoft.Health.Events/EventHubProcessor/ResumableAssignedPartitionProcessor.cs
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+        public async Task ResumeAsync(CancellationToken ct)
+        {
+            if (!_isRunning)
```
namalu, line 29: Instead of a bool you may want to use an int to represent `_isRunning`, and use `Interlocked.Exchange()` to set this value. Right now this implementation is not thread safe, and there is a slight possibility that `_isRunning` can become out of sync with the actual state.

```c
public async Task ResumeAsync(CancellationToken ct)
{
    if (Interlocked.Exchange(ref _isRunning, 1) == 0)
    {
        await _lockingBackgroundService.StartAsync(ct);
    }
}

public async Task RunAsync(CancellationToken ct)
{
    if (Interlocked.Exchange(ref _isRunning, 1) == 0)
    {
        await _lockingBackgroundService.StartAsync(ct);
    }
}

public async Task SuspendAsync(CancellationToken ct)
{
    if (Interlocked.Exchange(ref _isRunning, 0) == 1)
    {
        await _lockingBackgroundService.StopAsync(ct);
    }
}
```

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
```diff
@@ -0,0 +1,487 @@
...
+            _eventHubClientOptions = eventHubClientOptions;
+            _eventBatchingService = eventBatchingService;
+            _checkpointClient = checkpointClient;
+            _logger = logger;
+        }
+
+        private void CreateContainerClient()
+        {
+            var tokenCredential = EnsureArg.IsNotNull(_partitionLockingOptions.StorageTokenCredential, nameof(_partitionLockingOptions.StorageTokenCredential));
+            _containerClient = new BlobContainerClient(_partitionLockingOptions.BlobContainerUri, tokenCredential.GetCredential());
```
namalu, line 83: I would recommend creating an interface and wrapper class around BlobContainterClient (maybe IPartitionCoordinator and PartitionCoordinator). Currently, there is no way to Mock the BlobContainerClient which means there is no way to write unit tests for this class. Also, this class doesn't really need to know about BlobContainerClient or BlobClient it just needs a way to interact with a persistence layer in a very specific way. Abstracting away the specific storage account implementations behind a generalized interface will make testing easier and reduce the amount of code in this single class.

For example:

```c

public interface IPartitionCoordinator
{
    // Registers a processorId.
    Task ResisterProcessorIdAsync(string processorId, CancellationToken cancellationToken);

    // Returns a collection of processor ids representing active processors.
    Task<IEnumerable<string>> GetActiveProcessorIdsAsync(CancellationToken cancellationToken);

    // Checks if a partition with the given is is active (has been active within the maxInactiveTime).
    Task<bool> IsPartitionActiveAsync(string partitionId, TimeSpan maxInactiveTime, CancellationToken);

    // Claims a partition with the given partitionId.
    Task<bool> ClaimPartitionAsync(string partitionId, CancellationToken);

    ...etc...
}
```

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
```diff
@@ -0,0 +1,487 @@
...
+
+        private PartitionLockingBackgroundServiceOptions _partitionLockingOptions;
+
+        private EventHubClientOptions _eventHubClientOptions;
+
+        public PartitionLockingBackgroundService(
+            PartitionLockingBackgroundServiceOptions partitionLockingOptions,
+            EventHubClientOptions eventHubClientOptions,
+            EventBatchingService eventBatchingService,
+            StorageCheckpointClient checkpointClient,
```
namalu, line 68: You should use the interfaces here for `EventBatchingService` and `StorageCheckpointClient` rather than the implementations.

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
```diff
@@ -0,0 +1,487 @@
...
+            {
+                partitions[i] = ownedPartitions.ElementAt(i).ToString();
+            }
+
+            var eventHubName = EnsureArg.IsNotNull(_eventHubClientOptions.EventHubName, nameof(_eventHubClientOptions.EventHubName));
+            var eventHubNamespaceFQDN = EnsureArg.IsNotNull(_eventHubClientOptions.EventHubNamespaceFQDN, nameof(_eventHubClientOptions.EventHubNamespaceFQDN));
+            var eventHubConsumerGroup = EnsureArg.IsNotNull(_eventHubClientOptions.EventHubConsumerGroup, nameof(_eventHubClientOptions.EventHubConsumerGroup));
+            var tokenCrendential = EnsureArg.IsNotNull(_eventHubClientOptions.EventHubTokenCredential, nameof(_eventHubClientOptions.EventHubTokenCredential));
+
+            var processor = new AssignedPartitionProcessor(
```
namalu, line 456: I would recommend creating an interface for this class and injecting it in the constructor rather than instantiating it here. It's best to let the IoC Container handle resolving and creating classes, rather than creating them in the class.

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
```diff
@@ -0,0 +1,487 @@
...
+                    await Task.Delay(20000, ct);
+                }
+            });
+
+            return Task.CompletedTask;
+        }
+
+        private async Task StartAssignedPartitionProcessor(CancellationToken ct, List<int> ownedPartitions)
+        {
+            var maximumBatchSize = 100;
```
namalu, line 443: Shouldn't we set the maximumBatchSize to the batch size that we define in EventBatchingOptions.MaxEvents?

src/lib/Microsoft.Health.Events/EventHubProcessor/AssignedPartitionProcessor.cs
```diff
@@ -0,0 +1,231 @@
...
+            IEventConsumerService eventConsumerService,
+            StorageCheckpointClient checkpointClient,
+            ITelemetryLogger logger,
+            string[] assignedPartitions,
+            int eventBatchMaximumCount,
+            string consumerGroup,
+            TokenCredential tokenCredential,
+            string eventHubName,
+            string fullyQualifiedNamespace,
+            EventProcessorOptions clientOptions = default)
```
namalu, line 45: Should we set the `EventProcessorOptions.MaximumWaitTime` to the value we define in `EventBatchingOptions.FlushTimespan`?

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
namalu, line 456: From what I can see in the code, an instance of `AssignedPartitionProcessor` is created every 5 seconds - Is this correct?

Line 332 is in a loop that executes every 5 seconds and calls `RunAndRenew()`
Line 309 in `RunAndRenew()` calls `StartAssignedPartitionProcessor()` which creates a new `AssignedPartitionProcessor`.

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
```diff
@@ -0,0 +1,487 @@
...
+            await blobClient.UploadAsync(BinaryData.FromString(string.Empty), options, ct);
+
+            _logger.LogTrace($"Registered processor {_processorId}");
+        }
+
+        private async Task RegisterProcessorLoop(CancellationToken ct)
+        {
+            while (!ct.IsCancellationRequested)
+            {
+                await RegisterProcessor(ct);
```
namalu, line 108: Why do we need to run this loop? It looks like `_processorId` never changes, so why do we need to keep registering it every 15 seconds?

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
```diff
@@ -0,0 +1,487 @@
...
+                _logger.LogTrace("A task was cancelled");
+            }
+        }
+
+        protected override async Task ExecuteAsync(CancellationToken ct)
+        {
+            CreateContainerClient();
+            await RegisterProcessor(ct);
+            await ContinuallyRegisterProcessor(ct);
+            await CheckIfPartitionsAreUnclaimed(ct);
```
namalu, line 322: Is `CheckIfPartitionsAreUnclaimed()` just called for logging purposes?

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
```diff
@@ -0,0 +1,487 @@
...
+            while (!ct.IsCancellationRequested)
+            {
+                await RegisterProcessor(ct);
+                await Task.Delay(15000, ct);
+            }
+        }
+
+        private async Task<string[]> GetPartitionsIdsForEventHub(CancellationToken ct)
+        {
+            await using (var consumer = new EventHubConsumerClient(
```
namalu, line 115: Does a new instance of `EventHubConsumerClient` need to be created each time? This seems like a significant perf consideration. It looks like none of the values in the constructor ever change, so is there a reason we can't abstract this behind an interface and inject a singleton into this class?

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
dustinburson: I think the main limiting factor here is the processor needs the number of partitions it is working on and we don't know them at the time of DI.

namalu: I think a factory is the better pattern here if we want to get test coverage for this code.

src/console/Microsoft.Health.Fhir.Ingest.Console.Common/Extensions/EventProcessingExtensions.cs
```diff
@@ -60,6 +61,36 @@ public static IServiceCollection AddEventCheckpointing(this IServiceCollection s
...
         public static IServiceCollection AddResumableEventProcessor(this IServiceCollection services, IConfiguration config)
         {
+            // if assigned partition processor is enabled, then inject it
+            var partitionLockingOptions = new PartitionLockingBackgroundServiceOptions();
+            config.GetSection("PartitionLocking").Bind(partitionLockingOptions);
+            var partitionLockingEnabled = partitionLockingOptions.Enabled;
+
+            if (partitionLockingEnabled)
+            {
+                services.AddSingleton<ICheckpointClient, StorageCheckpointClient>();
```
rogordon01, line 71: Should we use [TryAddSingleton](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.extensions.servicecollectiondescriptorextensions.tryaddsingleton?view=dotnet-plat-ext-7.0) instead, since we add a `StorageCheckpointClient` in other [places](https://github.com/microsoft/iomt-fhir/blob/main/src/console/Microsoft.Health.Fhir.Ingest.Console.Common/Extensions/EventProcessingExtensions.cs#LL49C10-L49C12)?

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
wi-y: It is pretty much the same since we are replacing the previous locking from the event hub sdk with this implementation, and they make roughly the same amount of API calls to blob

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
wi-y: updated

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
wi-y: updated

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
wi-y: updated to throw a typed exception

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
wi-y: added a comment

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
wi-y: imo the complexity would be reduced slightly by knowing the total number of workers but I think the majority of the complexity comes from each worker needing to know which partitions it should be working on. Passing the total number of workers in as an environment variable seems like it should be possible. But passing in the assigned partitions as an environment variable requires us to switch to kubernetes [StatefulSet](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/) which allows us to pass in different values for environment variables (the current replica set setup we have today does not allow this). Before going with the approach in this PR, I was interested in pursuing StatefulSet but did not understand the implications/conflicts with our current custom resource definition and it did not seem like an easy switch

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
wi-y: yes. this is a concern I have as well and have a todo at the top of the class, but waiting seems like the best option for now. We could also force a restart after some time and try again

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
wi-y: added new interface and class, thanks for the suggestion.

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
wi-y: added a factory

wi-y: Nah the AssignedPartitionProcessor is only created once. 

We essentially go to RunProcessorAndRenewPartitionOwnership which creates StartAssignedPartitionProcessor and AssignedPartitionProcessor but we basically just await in there until we receive a cancellation token.

```
                await assignedPartitionProcessor.StartProcessingAsync(ct);
                await Task.Delay(Timeout.Infinite, ct);
```

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
wi-y: we use the blob time to determine if a processor is active or not. If the pod is still emitting the processor id to blob and/or the blob is less than 1 minute old then we consider the processor to still be active. So we have a background task that periodically uploads the processor id to blob so that we can compute how many processors are currently active

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
wi-y: yeah it is just for logging

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
wi-y: that makes sense. it is injected now as a singleton

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
wi-y: I agree conceptually but I have to leave it for now since EventBatchingService implements IEventConsumerService and there is already one being registered which is not EventBatchingService. I need to get rid of the other class that implements IEventConsumerService since it is not useful but it unfortunately touches a bit too much code to include that refactor in this PR

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
wi-y: Yeah it makes sense to set to the batch size or some multiple of the batch size. I had hardcoded it for the proof of concept here but should have updated it. it is updated now to be the batch size

src/lib/Microsoft.Health.Events/EventHubProcessor/AssignedPartitionProcessor.cs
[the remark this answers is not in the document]
wi-y: added a trace

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingBackgroundService.cs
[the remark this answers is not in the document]
wi-y: This is the link to the original event hub logic which steals partitions from other processors but does not immediately notify the other processors that the partition has been stolen, which leads to two processors having the same data in memory and writing the same data to the fhir service

https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/eventhub/Azure.Messaging.EventHubs.Shared/src/Processor/PartitionLoadBalancer.cs#L412

dustinburson (approved):
Overall looks good.  Added some comments for some improvement/clean up items

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingService.cs
```diff
@@ -0,0 +1,333 @@
...
+ * Description
+   - First, this class gets the number of total processors and total partitions
+   - Next, it will calculate how many partitions should be assigned to each processor
+   - Next, it will lock the partition(s) using blob leases
+   - Next, it will pass the partition id(s) into the AssignedPartitionProcessor which will run until the number of total processors changes
+
+ * For a flowchart of the PartitionLockingService see: \iomt-fhir\docs\PartitionLocking.md
+
+ * TODO:
+   - Handle cases where the ratio of processors to partitions is not divided evenly
```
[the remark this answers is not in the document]
wi-y: added 4 user stories under the Q1 feature

src/lib/Microsoft.Health.Events/EventHubProcessor/IPartitionCoordinator.cs
```diff
@@ -0,0 +1,40 @@
...
+        Task<IEnumerable<string>> GetActiveProcessorIdsAsync(CancellationToken cancellationToken);
+
+        // Checks if a partition with the given is is active (has been active within the maxInactiveTime).
+        Task<bool> IsPartitionActiveAsync(string partitionId, TimeSpan maxInactiveTime, CancellationToken cancellationToken);
+
+        // Claims a partition with the given partitionId.
+        Task<bool> ClaimPartitionAsync(string processorId, string partitionId, CancellationToken cancellationToken);
+
+        // Get the partitions owned
+        ConcurrentDictionary<string, DateTimeOffset> GetOwnedPartitions();
```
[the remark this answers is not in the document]
wi-y: updated to string[]

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionCoordinator.cs
```diff
@@ -0,0 +1,231 @@
...
+        }
+
+        public void ClearOwnedPartitions()
+        {
+            _ownedPartitions.Clear();
+        }
+
+        public Task<IEnumerable<string>> GetUnclaimedPartitionsAsync(IEnumerable<string> partitionsToCheck, CancellationToken cancellationToken)
+        {
+            throw new NotImplementedException();
```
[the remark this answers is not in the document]
wi-y: I thought it was needed originally but it was made redundant so I have removed it now from the class and the interface

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingService.cs
```diff
@@ -0,0 +1,333 @@
...
+   - Detect if acquiring partitions takes too long
+*/
+
+namespace Microsoft.Health.Events.EventHubProcessor
+{
+    public class PartitionLockingService
```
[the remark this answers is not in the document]
wi-y: added a user story

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionCoordinator.cs
```diff
@@ -0,0 +1,231 @@
...
+                if (blobItem.Metadata.TryGetValue("activeuntil", out var activeUntil))
+                {
+                    if (DateTime.Parse(activeUntil) > DateTime.UtcNow.AddMinutes(-1))
+                    {
+                        _logger.LogTrace($"Blob name: {blobItem.Name} is active");
+                        activePods.Add(blobItem.Name);
+                    }
+                    else
+                    {
+                        // todo: check the etag as a condition for delete
```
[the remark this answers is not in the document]
wi-y: updated the code to respect the etag

src/lib/Microsoft.Health.Events/EventHubProcessor/PartitionLockingService.cs
```diff
@@ -0,0 +1,333 @@
...
+            // find partitions that are not leased, and lease them
+            //
+            // todo: detect if this takes too long
+            // for example, if we acquire 2 partitions right away but need 2 more and cannot get them,
+            // should we renew the 2 that we have, or not? Alternatively we could quit after X minutes
+            _partitionCoordinator.ClearOwnedPartitions();
+
+            int partitionIdIndex = 0;
+
+            while (_partitionCoordinator.GetOwnedPartitions().Count < suggestedPartitionsPerProcessor && partitionIdIndex < _eventHubPartitions.Length)
```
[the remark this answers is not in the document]
wi-y: The intent was to have to partition coordinator manage all the state so I am ok with it being checked at the beginning of the while loop. Anywhere where we acquire partitions goes through the partition coordinator so I think it is ok if we query it as the source of truth

dustinburson (approved):

PR #273, opened 2023-05-02
Add additional telemetry for fhirtransformation stages

wi-y:
Adds a few additional metrics at various points of the FHIR transformation code to capture how much time is spent on each stage.

### DIFF
--- a/src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementGroupFhirImportService.cs
+++ b/src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementGroupFhirImportService.cs
@@ -86,8 +86,12 @@ namespace Microsoft.Health.Fhir.Ingest.Service
                 }
             }
 
-            await ProcessMeasurementEvents(measurements, template, log, ct);
-            await ProcessMeasurementGroupEvents(measurementGroups, template, log, ct);
+            string partitionId = events.FirstOrDefault()?.PartitionId;
+            using (ITimed processEventsDuration = log.TrackDuration(IomtMetrics.ProcessEventsDurationMs(partitionId)))
+            {
+                await ProcessMeasurementEvents(measurements, template, log, ct);
+                await ProcessMeasurementGroupEvents(measurementGroups, template, log, ct);
+            }
         }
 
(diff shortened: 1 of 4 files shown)


src/lib/Microsoft.Health.Fhir.Ingest/Service/MeasurementGroupFhirImportService.cs
```diff
@@ -139,17 +143,21 @@ private async Task ProcessMeasurementGroupEvents(IEnumerable<IEventMessage> even
                 try
                 {
                     IEnumerable<MeasurementGroup> groupedMeasurements;
+                    string partitionId;
 
                     try
                     {
-                        var partitionId = events.FirstOrDefault()?.PartitionId;
```
dustinburson, line 145: The current code works, but wanted to get your thoughts on a future improvement.  Seems like we should consider creating a context or container object that wraps the events collection and would include additional information about the collection like the partition id to reduce code redundancy in the future.

src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4FhirImportService.cs
```diff
@@ -59,19 +59,26 @@ public override async Task ProcessAsync(ILookupTemplate<IFhirTemplate> config, I
...
+                    }
+                    catch (Exception ex)
+                    {
+                        // user story 93303 to include appropriate message context for this error.
+                        throw new FhirDataMappingException(ex.Message, ex, nameof(FhirDataMappingException));
+                    }
                 }
 
-                foreach (var grp in grps)
+                using (ITimed saveObservationGroupsMs = _logger.TrackDuration(IomtMetrics.SaveObservationGroupsMs()))
```
dustinburson, line 76: I understand it is tricky in the current code but if we can figure out how to preserve the partition id so we can include it in the metric that would be a benefit.  It would also be nice if we knew the number of observation groups.

dustinburson (approved):
Looks good.  Made some suggestions but likely require a bigger refactor than what we want to take on now.

rogordon01 (approved):

wi-y:
There isn't a need to add this telemetry so going to close this PR without merging

PR #278, opened 2023-06-20
Support optional encounter linking in Lookup & Create

dustinburson:
Update the IdentityService to always link and encounter id if specified in the measurement group.

This will allow optional encounter linking in the LookupIdentityService & CreateIdentityService.

The DeviceAndPatientWithEncounterIdentityService will still require the encounter id present or a ResourceIdentityNotDefinedException will be thrown.

### DIFF
--- a/src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4DeviceAndPatientLookupIdentityService.cs
+++ b/src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4DeviceAndPatientLookupIdentityService.cs
@@ -3,6 +3,7 @@
 // Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // -------------------------------------------------------------------------------------------------
 
+using System.Collections.Generic;
 using System.Threading.Tasks;
 using EnsureThat;
 using Microsoft.Health.Extensions.Fhir;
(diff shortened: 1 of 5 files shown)


namalu (approved):

src/lib/Microsoft.Health.Fhir.R4.Ingest/Service/R4DeviceAndPatientLookupIdentityService.cs
```diff
@@ -48,5 +49,20 @@ protected async override Task<(string DeviceId, string PatientId)> LookUpDeviceA
             var device = await ResourceManagementService.GetResourceByIdentityAsync<Model.Device>(value, system).ConfigureAwait(false) ?? throw new FhirResourceNotFoundException(ResourceType.Device);
             return (device.Id, GetPatientIdFromDevice(device));
         }
+
+        protected async override Task<IDictionary<ResourceType, string>> ResolveResourceIdentitiesInternalAsync(IMeasurementGroup input)
+        {
+            EnsureArg.IsNotNull(input, nameof(input));
+
+            var identities = await base.ResolveResourceIdentitiesInternalAsync(input).ConfigureAwait(false);
```
kyclai, line 57: `LookUpDeviceAndPatientIdAsync` called by `base.ResolveResourceIdentitiesInternalAsync` can throw a `FhirResourceNotFoundException`. Should we handle this exception?

dustinburson: That is intentional, if the patient, device, or the encounter id fails to resolve on the FHIR service a FhirResourceNotFoundException is expected.

PR #280, opened 2023-06-26
Personal/rogordon/use sem var

rogordon01:
This pr updates the CI pipeline to use Semantic Version to generate versions. These versions will be used in nuget packages as well as to label PR's.

Also, this PR restricts publishing nugets to only the CI pipeline.

### DIFF
--- a/build/azure-pipelines-pr.yml
+++ b/build/azure-pipelines-pr.yml
@@ -11,6 +11,8 @@ stages:
   dependsOn: []
   jobs:
   - job: Semver
+    pool:
+      vmImage: 'ubuntu-20.04'
     variables:
       skipComponentGovernanceDetection: true
     steps:
--- a/build/azure-pipelines-pr.yml
+++ b/build/azure-pipelines-pr.yml
@@ -31,8 +31,6 @@ stages:
       vmImage: 'windows-2019'
     steps:
     - template: build.yml
-      parameters:
-        packageArtifacts: false
 
   - job: Linux
     pool:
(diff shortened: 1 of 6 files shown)


rogordon01:
The GitVersion task incorporates GitTags when determining the version. Since we have a git tag of `2.0.1` it is beginning from there, and incrementing based on every subsequent commit. This will bring the initial version to `2.0.63`

build/azure-pipelines-pr.yml
```diff
@@ -4,17 +4,40 @@
...
+    majorMinorPatch: $[stageDependencies.UpdateVersion.Semver.outputs['IomtVersion.GitVersion.MajorMinorPatch']]
+    nuGetVersion: $[stageDependencies.UpdateVersion.Semver.outputs['IomtVersion.GitVersion.SemVer']]
+  jobs:
+  - job: Windows
+    pool:
+      vmImage: 'windows-2019'
+    steps:
+    - template: build.yml
+      parameters:
+        packageArtifacts: false
```
rogordon01, line 35: It looked like we previously published nugets from the PR pipeline. I didn't see a need for this so I removed it. Let me know if this isn't desired.

pallar-ms:
This is the release that should now be updated to not publish nuget packages but only update the package version dependencies in the PaaS repos. https://microsofthealthoss.visualstudio.com/FhirServer/_release?definitionId=26&view=mine&_a=releases

rogordon01:
> This is the release that should now be updated to not publish nuget packages but only update the package version dependencies in the PaaS repos. https://microsofthealthoss.visualstudio.com/FhirServer/_release?definitionId=26&view=mine&_a=releases

Thanks @pallar-ms. Yea I ended up stumbling on it during testing. After this PR is merged I'll modify it. I want to make sure the Renovate actually picks up the new nugets before I edit it

dustinburson (approved):

PR #287, opened 2023-07-14
fix "Deploy to Azure" button

wi-y:
The URL for the site that hosted the "Deploy to Azure" button appears to no longer a have a valid SSL certificate which unfortunately means that the button does not show up when viewing the [ARM Installation page](https://github.com/microsoft/iomt-fhir/blob/main/docs/ARMInstallation.md) :(

image

The old hosting location was: https://azuredeploy.net/deploybutton.png

[These docs](https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/deploy-to-azure-button) say to use https://aka.ms/deploytoazurebutton so this PR updates to URL to the new location.

image


kyclai (approved):
Luckily we're using the recommended URL for the buttons in our Azure docs :) https://learn.microsoft.com/en-us/azure/healthcare-apis/iot/deploy-choose-method

rogordon01 (approved):

PR #288, opened 2023-07-27
Add deployment scripts and Bicep templates for OSS IoMT Service run on dotnet console apps 

oliviaw7:
Add two options (PowerShell script and a single Bicep template entry point) for deploying resources needed to run OSS IoMT Service run on dotnet console apps. 

- Main.bicep creates a resource group and calling another bicep template (InfrastructureSetup.bicep) for deploying resources within that resource group. 
- InfrastructureSetup.bicep is used for initial setup of necessary resources to run IoMT Service. 
- ContainerAppSetup.bicep pulls and runs container images of Normalization and FHIR Transformation services run as dotnet console apps on Container Apps. Each service is run on its own scalable instance of a Container App within the same Container App environment. 
- Create-IoMTContainerAppEnv.ps1 to deploy InfrastructureSetup.bicep and ContainerAppSetup.bicep. The PowerShell script also builds and pushes two container images (normalization and fhir-transformation) to the ACR created in InfrastructureSetup.bicep to be used by the Container Apps. 
- BuildContainerImages.bicep uses the deployment scripts Bicep template resource to build and push container images to ACR created in InfrastructureSetup.bicep.
- ContainerApp-SingleAzureDeploy.bicep serves as a single-entry point for deploying all necessary resources via Bicep templates. 
- BicepInstallation.md details the possible options.


deploy/templates/bicep/devicecontent.json
```diff
@@ -0,0 +1,78 @@
+{
+    "templateType": "CollectionContent",
```
dustinburson, line 2: Suggestion: Is there a way to resolve these from our https://github.com/microsoft/iomt-fhir/tree/main/sample/templates path?  Concern that having example templates here is another code location we will need to remember/maintain.

deploy/templates/bicep/ContainerAppSetup.bicep
```diff
@@ -0,0 +1,385 @@
...
+              name: 'NormalizationEventHub__EventHubConsumerGroup'
+              value: '$Default'
+            }
+            {
+              name: 'NormalizationEventHub__EventHubName'
+              value: 'normalizeddata'
+            }
+            {
+              name: 'ResourceIdentity__ResourceIdentityServiceType'
+              value: 'Create'
```
dustinburson, line 299: Ideally this should be a parameter the customer can define to switch between Create and Lookup. The tags will also need to be updated based on this parameter i.e. IomtFhirConnector: 'ResourceIdentity:Create' would IomtFhirConnector: 'ResourceIdentity:Lookup' if Lookup is selected.

deploy/templates/bicep/ContainerAppSetup.bicep
```diff
@@ -0,0 +1,385 @@
...
+
+resource fhirTransformationContainerApp 'Microsoft.App/containerApps@2022-03-01' ={
+  name: 'fhir-transformation'
+  location: location
+  tags: {
+    IomtFhirConnector: 'ResourceIdentity:Create'
+    IomtFhirVersion: 'R4'
+  }
+  identity: {
+    type: 'SystemAssigned,UserAssigned'
```
dustinburson, line 196: How is SystemAssigned and UserAssigned being used?  I would think it should only be 1.

dustinburson (changes requested):

dustinburson (commented):
Can we include a markdown file with instructions for how to execute the bicep template?

deploy/templates/bicep/devicecontent.json
oliviaw7: Yes, I can remove these. I originally had these here for testing purposes and forgot to remove them.

deploy/templates/bicep/ContainerAppSetup.bicep
oliviaw7: The UserAssigned identity is being used to pull the container image from the ACR. The same UserAssigned identity is being used to pull the image for the Normalization and FHIR transformation Container Apps. Each Container App also has its own SystemAssigned Identity to assign permissions to other resources that are used by each service. A UserAssigned identity needs to be used for pulling the image since the SystemAssigned Identity isn't created until after the Container App is deployed. I could create separate UserAssigned identities for Normalization and FHIR Transformation to encompass all the necessary permissions (AcrPull + permissions specific to Normalization/FHIR Transformation). However, this UserAssigned identity is also being used to run the script in BuildContainerImages.bicep which pushes the images to ACR so a separate UserIdentity would need to be created as well with AcrPush + Contributor permissions. I am happy to make these changes if you think the separation would make more sense.

oliviaw7:
> Can we include a markdown file with instructions for how to execute the bicep template?

Yes, I currently have two deployment options, one via a PowerShell script and one via executing a bicep template. Should I combine these deployment options into one markdown file or make one for each?

deploy/templates/bicep/ContainerAppSetup.bicep
oliviaw7: Got it, I will add that in.

deploy/templates/bicep/UploadTemplates.bicep
```diff
@@ -0,0 +1,111 @@
...
+  sku: {
+    name: 'Standard_RAGRS'
+  }
+}
+
+param blobContainerName string = 'template'
+param devicecontentFile string = 'devicecontent.json'
+param fhirmappingFile string = 'fhirmapping.json'
+param devicecontentURL string = 'https://raw.githubusercontent.com/microsoft/iomt-fhir/main/sample/templates/basic/devicecontent.json'
+param fhirmappingURL string = 'https://raw.githubusercontent.com/microsoft/iomt-fhir/main/sample/templates/basic/devicecontent.json'
```
dustinburson, line 42: Typo, should be FHIR mapping. Also double check the content.  I don't believe the device mapping has patient id and won't work with create mode.  Feel free to update that template.  Please double check with the content of the templates you deleted to make sure nothing else is missed.

dustinburson (changes requested):
Changes look good.  One issue identified, left a comment with details.

deploy/templates/bicep/ContainerAppSetup.bicep
```diff
@@ -0,0 +1,385 @@
...
+              name: 'TemplateStorage__BlobContainerName'
+              value: 'template'
+            }
+            {
+              name: 'FhirClient__UseManagedIdentity'
+              value: 'true'
+            }
+            {
+              name: 'FhirService__Url'
+              value: fhirService.properties.authenticationConfiguration.audience
```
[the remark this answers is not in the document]
oliviaw7: I wasn't able to find a URL property. Is there an alternative besides the audience URL that would be better?

deploy/templates/bicep/ContainerAppSetup.bicep
```diff
@@ -0,0 +1,385 @@
...
+              name: 'TemplateStorage__BlobStorageContainerUri'
+              value: '${storageAccount.properties.primaryEndpoints.blob}template'
+            }
+            {
+              name: 'TemplateStorage__BlobContainerName'
+              value: 'template'
+            }
+            {
+              name: 'FhirClient__UseManagedIdentity'
+              value: 'true'
```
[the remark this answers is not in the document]
oliviaw7: Thanks for pointing this out. The parameter has been added.

deploy/templates/bicep/UploadTemplates.bicep
oliviaw7: Thanks for catching this! You are right, the device mapping did not include patient id. The template has been updated to include patientid.

docs/BicepInstallation.md
```diff
@@ -0,0 +1,104 @@
...
+
+* Storage Account 
+* Event Hubs Namespace  
+* Two Event Hubs 
+* Azure Health Data Services Workspace
+* FHIR Service* 
+* Azure Container Registry 
+* Log Analytics Workspace 
+* App Insights 
+* Managed Identity
```
ms-teli, line 14: Nit, can you specify the type of the MI? For example: System Managed Identity

docs/BicepInstallation.md
```diff
@@ -0,0 +1,104 @@
...
+* Two Container Apps  
+
+An additional Storage Account and Container Instances will be created if the 'deploymentScripts' Bicep resource type is used in the chosen deployment option (Options 1 and 2). 
+
+\* NOTE: These deployment options use the FHIR service in Azure Health Data Services as opposed to the Azure API for FHIR. For more information on the differences between these services, please see this [page](https://learn.microsoft.com/en-us/azure/healthcare-apis/fhir/overview). 
+
+### Prerequisites
+To run any of these deployment options, the following items must be set up before execution:
+
+* Contributor and User Access Administrator OR Owner permissions on your Azure subscription
```
ms-teli, line 25: I feel that we don't have to mention this as prerequisite because customers may have their own way to manage and grant the permission granularity for their subscription.

docs/BicepInstallation.md
```diff
@@ -0,0 +1,104 @@
...
+
+This option deploys the Bicep file that was used to generate the ARM template in Option 1. This Bicep template serves as a single entry point for setting up all necessary Azure resources and role assignments. Sample configuration templates, [devicecontent.json](../sample/templates/basic/devicecontent.json) and [fhirmapping.json](../sample/templates/basic/fhirmapping.json) are also uploaded to the 'template' container in the Storage Account using a User-Assigned Managed Identity. 
+
+The 'deploymentScripts' resource type in Bicep is used to (1) upload the sample mapping templates and (2) build and push container container images to the ACR. An additional Storage Account is provisioned to execute these deployment scripts. A Container Instance is also created for each 'deploymentScripts' resource instance and is deleted upon successful deployment. 
+
+To view the progress of the deployment, navigate to the resource group in Azure Portal and select the 'Deployments' tab under 'Settings' in the left panel.
```
ms-teli, line 82: Nit: it would be better if we can put a screenshot here.

docs/BicepInstallation.md
```diff
@@ -0,0 +1,104 @@
...
+
+The mapping configurations for device content and converting to FHIR need to be added to the 'template' container in the deployed Azure Storage blob. Navigate to the  Storage Account and select the 'template' storage container. From there, upload the configurations to complete set up of the IoMT FHIR Connector.
+
+More information on mapping templates can be found [here](https://github.com/microsoft/iomt-fhir/blob/7794cbcc463e8d26c3097cd5e2243d770f26fe45/docs/Configuration.md). Full examples can be found in the repository under [/sample/templates](https://github.com/microsoft/iomt-fhir/tree/7794cbcc463e8d26c3097cd5e2243d770f26fe45/sample/templates)
```
ms-teli, line 97: Nit: can we point the link to the main branch or using the relative path? Otherwise, it's not linking to the latest content.

deploy/templates/bicep/main.bicep
```diff
@@ -0,0 +1,54 @@
...
+  'southeastasia'
+  'swedencentral'
+  'switzerlandnorth'
+  'westcentralus'
+  'westeurope'
+  'westus2'
+  'westus3'
+  'uksouth'
+])
+param location string
```
ms-teli, line 31: Can we add a parameter fhirVersion? It can have only one value option "R4"

deploy/templates/bicep/UploadTemplates.bicep
```diff
@@ -0,0 +1,111 @@
...
+    principalType: 'ServicePrincipal'
+  }
+}
+
+resource deploymentStorageAccount 'Microsoft.Storage/storageAccounts@2022-09-01' = {
+  name: '${baseName}deploysa'
+  location: location
+  tags: {
+    IomtFhirConnector: 'ResourceIdentity:${resourceIdentityResolutionType}'
+    IomtFhirVersion: 'R4'
```
ms-teli, line 30: And please pass the fhir version parameter to here

deploy/templates/bicep/InfrastructureSetup.bicep
```diff
@@ -0,0 +1,209 @@
+param baseName string
```
ms-teli, line 1: I'm seeing 3 different naming patterns:
 {basename}{service type abbreviation}
{service type abbreviation}-{basename}
{service type abbreviation}{basename}

can we use just one pattern? since many services don't support using "-" in the name, I think we can get rid of it.

deploy/templates/bicep/Create-IomtContainerAppEnv.ps1
```diff
@@ -0,0 +1,114 @@
...
+(
+    [Parameter(Mandatory = $true)]
+    [ValidateNotNullOrEmpty()]
+    [ValidateLength(3,16)]
+    [ValidateScript({
+        if ("$_" -cmatch "(^([a-z]|\d)+$)") {
+            return $true
+        }
+        else {
+			throw "Service name must be lowercase and numbers"
```
ms-teli, line 11: Nit pick: misaligned indention.

deploy/templates/bicep/Create-IomtContainerAppEnv.ps1
```diff
@@ -0,0 +1,114 @@
...
+# Get current account context - User/Service Principal
+$azAccountId = az account show --query user.name --output tsv
+$azAccountType = az account show --query user.type --output tsv
+if ($azAccountType -eq "user") {
+    Write-Host "Current account context is user: $($azAccountId)"
+    
+    $currentUser = az ad user show --id $azAccountId #Get-AzADUser -UserPrincipalName $azContext.Account.Id 
+
+    if ($currentUser) {
+        $currentObjectId = az ad user show --id $azAccountId --query id --output tsv
```
ms-teli, line 68: is this variable being used anywhere?

deploy/templates/bicep/Create-IomtContainerAppEnv.ps1
```diff
@@ -0,0 +1,114 @@
+param
```
ms-teli, line 1: Nit: please add period to logging messages.

deploy/templates/bicep/Create-IomtContainerAppEnv.ps1
```diff
@@ -0,0 +1,114 @@
...
+    Write-Host "Current account context is service principal: $($azAccountId)"
+    $currentObjectId = az ad sp show --id $azAccountId --query id --output tsv
+}
+else {
+    Write-Host "Current context is account of type '$($azAccountType)' with id of '$($azAccountId)"
+    throw "Running as an unsupported account type. Please use either a 'User' or 'Service Principal' to run this command"
+}
+
+# Create a resource group in provided subscription if it doesn't exist and deploy Azure resources needed to run IoMT Service.
+$setupTemplate = "Main.bicep"
```
ms-teli, line 85: the letter case in the filename doesn't match. Please make sure it's named "Main.bicep"

deploy/templates/bicep/ContainerAppSetup.bicep
```diff
@@ -0,0 +1,392 @@
...
+            {
+              name: 'Template__DeviceContent'
+              value: 'devicecontent.json'
+            }
+          ]
+        }
+      ]
+      scale: {
+        minReplicas: 1
+        maxReplicas: 10
```
ms-teli, line 185: can we make it just 2 if that meets the basic requirements?

oliviaw7: Are you referring to the number of minReplicas or are you suggesting there should always only be two replicas?

docs/BicepInstallation.md
oliviaw7: Thanks for pointing this out, I didn't consider that customers may have different permission granularities. Out of curiosity, are most OSS users familiar with Azure and the permissions needed to perform these deployments? I put this as a requirement since whoever is running it will need permissions to deploy resources and assign role assignments in their subscription, otherwise they will get an error when they go to deploy. Do you think there is a better way to call this out or do you think it's not necessary?

docs/BicepInstallation.md
```diff
@@ -0,0 +1,104 @@
...
+|**LookupWithEncounter**|Similar to the 'Lookup' setting but allows you to include an encounter identifier with the message to associate with the device/patient. The encounter is looked up during processing and any observations created are linked to the encounter. The association here is assumed to be one encounter per device.
+
+## Deployment 
+### Option 1: Single-click Deploy to Azure via ARM template generated from Bicep Template
+
+### Option 2: Deploy a single Bicep file locally 
+Deploy the [Bicep template](../deploy/templates/bicep/ContainerApp-SingleAzureDeploy.bicep) by running the following command: 
+
+```PowerShell
+az deployment sub create --location <Location> --template-file ContainerApp-SingleAzureDeploy.bicep
```
msjasteppe, line 72: You might want to show an example of how the command should look (e.g., az deployment sub create --location "westus3" --template-file ContainerApp-SingleAzureDeploy.bicep

docs/BicepInstallation.md
```diff
@@ -0,0 +1,104 @@
...
+The mapping configurations for device content and converting to FHIR need to be added to the 'template' container in the deployed Azure Storage blob. Navigate to the  Storage Account and select the 'template' storage container. From there, upload the configurations to complete set up of the IoMT FHIR Connector.
+
+More information on mapping templates can be found [here](./Configuration.md). Full examples can be found in the repository under [../sample/templates](../sample/templates/)
+
+### Additional Deployment Notes
+To view the progress of a deployment, navigate to the resource group in Azure Portal and select the 'Deployments' tab under 'Settings' in the left panel.
+
+All deployment options reference separate Bicep files that are used to provision a set of resources. To redeploy a specific supporting Bicep file, run the following command: 
+```PowerShell
+az deployment group create --resource-group <baseName> --template-file <File.bicep>
```
msjasteppe, line 103: You might want to show an example of how the command should look (e.g., az deployment group create --resource-group "testbicepdeployment" --template-file main.bicep

docs/BicepInstallation.md
```diff
@@ -0,0 +1,104 @@
+# Installation options via Bicep Templates for use with FHIR service in Azure Health Data Services 
+This article details three deployment options for provisioning and installation of the IoMT FHIR Connector for Azure and [FHIR service in Azure Health Data Services](https://learn.microsoft.com/en-us/azure/healthcare-apis/fhir/overview) using Bicep templates. 
+
+The following Azure components will be provisioned once deployment has completed:
+
+* Storage Account 
+* Two Blob Containers
+* Event Hubs Namespace  
+* Two Event Hubs 
+* Azure Health Data Services Workspace
```
msjasteppe, line 10: Workspace should be lowercase.

deploy/templates/bicep/ContainerAppSetup.bicep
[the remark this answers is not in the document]
dustinburson: I would then create a variable in the template and use it for both the audience and the FHIR URL, basically taking the logic you have here ``https://${healthWorkspace.name}-fs-${baseName}.fhir.azurehealthcareapis.com`` and putting it into a variable for you reference in multiple locations.

deploy/templates/bicep/UploadTemplates.bicep
```diff
@@ -0,0 +1,151 @@
+@minLength(3)
+@maxLength(16)
+@description('Basename that is used to name provisioned resources. Should be alphanumeric, at least 3 characters and less than 16 characters.')
+param baseName string
+
+@description('Location where the resources are deployed. For a list of Azure regions where Azure Health Data Services are available, see [Products available by regions](https://azure.microsoft.com/explore/global-infrastructure/products-by-region/?products=health-data-services)')
+@allowed([
+  'australiaeast'
```
dustinburson, line 8: Suggestion: Is there a way to share a common location list between the bicep templates?  It would simplify adding new locations.

dustinburson (approved):

deploy/templates/bicep/UploadTemplates.bicep
oliviaw7: Thanks for bringing this up, I think this is a great idea. I tried creating a separate JSON file to store a list of the deployment options and loading it into the bicep templates using loadJsonContent() in the allowed decorator for the location parameter. However, I got an error saying that the allowed values must be available at compile time. There is also a way to pass in allowed values by creating a user-defined type, but that would require enabling a preview feature. I will look into other possible options.

deploy/templates/bicep/Create-IomtContainerAppEnv.ps1
oliviaw7: $currentObjectId is being used to validate that the user has an objectId, otherwise an error will be thrown. However, I removed another line where the value of $currentObjectId is set, but never used. Thanks for pointing this out.

dustinburson (approved):

docs/BicepInstallation.md
dustinburson: I do believe it is worth calling out they need at least Contributor and User Access Admin to run the scripts.  These are standard Azure roles and their permissions are well known.

deploy/templates/bicep/ContainerAppSetup.bicep
ms-teli: oh, sorry about the confusion. I was thinking max as 2. The reason is we don't want to surprise customer. Or you can add a reminder about the max replica.

PR #290, opened 2023-08-08
Draft - Refactor DI to remove BuildServiceProvider

dustinburson:
In the current solution DI there are a few places were BuildServiceProvider is invoked to resolve instances that can be used in factories.  This is a problem because if dependencies are added out of order it can result in run time errors.

Changes:
- Moves classes in Auth Provider folder to Auth folder to align with namespace.
- Remove obsolete MeasurementCollectionToFhir console app and associated scripts & templates.
- Remove custom EnvironmentConfiguration and replace with built in HostBuilder functionality.
- Remove invocation of BuildServiceProvider from FhirTransformationExtensions and move logic inside FhirClientExensions with a factory method for TokenCredentials.


src/lib/Microsoft.Health.Extensions.Fhir.R4/FhirClientExtensions.cs
```diff
@@ -23,9 +24,27 @@ public static IServiceCollection AddFhirClient(this IServiceCollection serviceCo
             EnsureArg.IsNotNull(serviceCollection, nameof(serviceCollection));
             EnsureArg.IsNotNull(configuration, nameof(configuration));
 
-            var url = new Uri(configuration.GetValue<string>("FhirService:Url"));
+            Uri url = new (configuration.GetValue<string>("FhirService:Url"));
             bool useManagedIdentity = configuration.GetValue<bool>("FhirClient:UseManagedIdentity");
 
+            serviceCollection.TryAddSingleton<TokenCredential>(sp =>
```
namalu, line 30: Can we use a singleton TokenCredential here?

I remember going through this code before, and in several places we new up credentials in place and provide the specific instance to whatever class needs it. I believe this has to do with "internal" resources using a different credential than "external".

namalu (dismissed):

src/lib/Microsoft.Health.Extensions.Fhir.R4/FhirClientExtensions.cs
dustinburson: Thank for the feedback, I was thinking the same thing.  Looking through the OSS and PaaS code I didn't see any instances but still I don't like it.  Even if it is fine now, it doesn't seem like a resilient solution. I am trying to decide if I should move all the code in this block into it's own factory class that is registered as a singleton or if I should create an interface for the *AuthServices and that is what is registered as a singleton.  I still don't like it because at their core the auth services inherit from TokenCredential so I am leaning towards the factory approach.  Thoughts?

namalu (approved):

PR #291, opened 2023-08-09
Initialize event hub partition reading from the next event after the checkpoint

wi-y:
Previously the iot connector would begin reading from a partition based on the last checkpointed event's enqueued time. The date compassion used `EventPosition.FromEnqueuedTime(checkpoint.LastProcessed)` was inclusive, meaning that the last checkpointed event was included in the first batch on initialization. This is not desirable because it reprocesses one event on shutdown/startup or when partition ownership is lost, which is unnecessary. Reprocessing this one event also causes issues with the data freshness calculations since the reprocess of an older event makes it seem like the iot connector had fallen behind.

There are options to instead use the event offset or the event sequence id to resume processing events from a checkpoint. Since we already store the event offset and sequence id in the checkpoint the only change is to change the intialization logic to use the offset instead of a datetime. There is also an option to either include the last checkpointed event (isInclusive = true) or to exclude the last checkpointed event (isInclusive = false). We are choosing to exclude the last checkpointed event since it has already been processed.

### DIFF
--- a/src/lib/Microsoft.Health.Events/EventHubProcessor/BaseEventProcessor.cs
+++ b/src/lib/Microsoft.Health.Events/EventHubProcessor/BaseEventProcessor.cs
@@ -94,9 +94,11 @@ namespace Microsoft.Health.Events.EventHubProcessor
             {
                 EventConsumerService.NewPartitionInitialized(partitionId);
                 var checkpoint = await CheckpointClient.GetCheckpointForPartitionAsync(partitionId, initArgs.CancellationToken);
-                initArgs.DefaultStartingPosition = EventPosition.FromEnqueuedTime(checkpoint.LastProcessed);
 
-                Logger.LogTrace($"Starting to read partition {partitionId} from checkpoint {checkpoint.LastProcessed}");
+                // Get the last checkpointed event offset and begin reading from the next event by setting isInclusive = false
+                initArgs.DefaultStartingPosition = EventPosition.FromOffset(checkpoint.Offset, isInclusive: false);
+
+                Logger.LogTrace($"Starting to read partition {partitionId} from next event after checkpoint offset {checkpoint.Offset} checkpoint time {checkpoint.LastProcessed}");
                 Logger.LogMetric(EventMetrics.EventHubPartitionInitialized(partitionId), 1);
             }
             catch (TaskCanceledException ex)


rogordon01 (approved):

rogordon01 (changes requested):
Actually, how will this affect when we [flush events](https://github.com/microsoft/iomt-fhir/blob/main/src/lib/Microsoft.Health.Events/EventConsumers/Service/EventBatchingService.cs#L229)? Don't we need to update that logic to select the latest event in the batch based on offset? I think that its selected based on time now. Or will the event selected always have the correct offset to checkpoint?

wi-y:
> Actually, how will this affect when we [flush events](https://github.com/microsoft/iomt-fhir/blob/main/src/lib/Microsoft.Health.Events/EventConsumers/Service/EventBatchingService.cs#L229)? Don't we need to update that logic to select the latest event in the batch based on offset? I think that its selected based on time now. Or will the event selected always have the correct offset to checkpoint?

I don't think we need to update that logic because the Event Hub orders events already by offset and time* (there can be ties with time) and we currently read the events in order from the Event Hub, so the last event in the batch would have the highest offset.

An example batch today would have the following enqueued time and offsets, so I think it is okay to leave the checkpointing logic as-is, but if there are concerns with that please let me know

enqueued time | offset
8/9/2023 7:12:30 PM +00:00 111671159616 (first event in batch)
8/9/2023 7:12:30 PM +00:00 111671209864
8/9/2023 7:12:30 PM +00:00 111671260112
8/9/2023 7:12:30 PM +00:00 111671310360
8/9/2023 7:12:30 PM +00:00 111671360608
8/9/2023 7:12:30 PM +00:00 111671410856
8/9/2023 7:12:30 PM +00:00 111671461104
8/9/2023 7:12:30 PM +00:00 111671511352
8/9/2023 7:12:30 PM +00:00 111671561600
8/9/2023 7:12:30 PM +00:00 111671611848
8/9/2023 7:12:30 PM +00:00 111671662096
8/9/2023 7:12:30 PM +00:00 111671712344
8/9/2023 7:12:30 PM +00:00 111671762592
8/9/2023 7:12:30 PM +00:00 111671812840
8/9/2023 7:12:30 PM +00:00 111671863088
8/9/2023 7:12:30 PM +00:00 111671913336
8/9/2023 7:12:30 PM +00:00 111671963584
8/9/2023 7:12:30 PM +00:00 111672013832
8/9/2023 7:12:30 PM +00:00 111672064080
8/9/2023 7:12:30 PM +00:00 111672114328 (last event in batch, will be checkpointed)

rogordon01 (approved):

dustinburson:
Looks good, but I think this reveals another flaw with how we report the freshness metric.  I believe today we report the freshness as the time of the first event in the batch with the assumption that enqueued time always goes up.  The goal for the freshness metric is to is to ensure that is we report we are fresh for time X then there can be no more events for time X.  Based on the recent finding this appears to be an incorrect assumption and in some cases (mainly large batches) we can report a freshness of X but still have events associated with time X.  I will create a work item for us to address that problem separately.

dustinburson (approved):

rogordon01:
Why use the offset versus the sequence id? Just curious.

wi-y:
> Looks good, but I think this reveals another flaw with how we report the freshness metric. I believe today we report the freshness as the time of the first event in the batch with the assumption that enqueued time always goes up. The goal for the freshness metric is to ensure if we are fresh for time X then there can be no more events for time X. Based on the recent finding this appears to be an incorrect assumption and in some cases (mainly large batches) we can report a freshness of X but still have events associated with time X. I will create a work item for us to address that problem separately.

The freshness metric seems to be calculated based on a diff between the current time and the last event in the batch. https://github.com/microsoft/iomt-fhir/blob/ff8ffe0956dc524e795418e3f71cf08b9d5a0de1/src/lib/Microsoft.Health.Events/EventConsumers/Service/EventBatchingService.cs#L268C115-L268C142

So I believe that means if we are processing a lot of events with the same time X then each batch would diff the current time with time X. So if we are stuck processing a bunch of events with time X the diff between current time and X would increase, and should correctly report the delay (the diff between current time and time X).

wi-y:
> Why use the offset versus the sequence id? Just curious.

Either offset or sequence id should work but I went with offset since there was more documentation on the offset on this page, and not much documentation on sequence id: https://learn.microsoft.com/en-us/azure/event-hubs/event-hubs-features

dustinburson:
> > Looks good, but I think this reveals another flaw with how we report the freshness metric. I believe today we report the freshness as the time of the first event in the batch with the assumption that enqueued time always goes up. The goal for the freshness metric is to ensure if we are fresh for time X then there can be no more events for time X. Based on the recent finding this appears to be an incorrect assumption and in some cases (mainly large batches) we can report a freshness of X but still have events associated with time X. I will create a work item for us to address that problem separately.
> 
> The freshness metric seems to be calculated based on a diff between the current time and the last event in the batch. https://github.com/microsoft/iomt-fhir/blob/ff8ffe0956dc524e795418e3f71cf08b9d5a0de1/src/lib/Microsoft.Health.Events/EventConsumers/Service/EventBatchingService.cs#L268C115-L268C142
> 
> So I believe that means if we are processing a lot of events with the same time X then each batch would diff the current time with time X. So if we are stuck processing a bunch of events with time X the diff between current time and X would increase, and should correctly report the delay (the diff between current time and time X).

Yes, the freshness delay will be correct.  Sorry I was referring the last EventTimestampLastProcessedPerPartition which is in the same code block.  It looks like I was incorrect and we are basing that on the last time in the batch too.  That falls short of the original intent which was to represent the minimum guaranteed data time available in the system.  Using the last time works if we assume we have sufficient precision and each enqueued time is unique.  It looks like both of those assumptions are incorrect.  Since we can't author this metric externally today (because DateTime types aren't supported) it isn't a huge issue.  It just changes how we should view the metric, it means we have started processing some data for that time period but not necessarily finished.  I will still create a backlog item to re-evaluate how we calculate the time but it will be low priority.

PR #293, opened 2023-08-28
Update FhirClient to latest version

namalu:
This PR updates FhirClient to the latest version.

Updated FhirException class to FhirClientException class.

Updated unit test to conform with new behavior of  FhirClientException.

### DIFF
--- a/src/lib/Microsoft.Health.Extensions.Fhir.R4/Telemetry/Exceptions/FhirServiceExceptionProcessor.cs
+++ b/src/lib/Microsoft.Health.Extensions.Fhir.R4/Telemetry/Exceptions/FhirServiceExceptionProcessor.cs
@@ -41,8 +41,8 @@ namespace Microsoft.Health.Extensions.Fhir.Telemetry.Exceptions
 
             switch (exception)
             {
-                case FhirClientException _:
-                    var status = ((FhirClientException)exception).StatusCode;
+                case FhirClientException fhirClientException:
+                    var status = fhirClientException.StatusCode;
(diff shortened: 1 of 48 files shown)


rogordon01 (dismissed):

src/lib/Microsoft.Health.Extensions.Fhir.R4/Telemetry/Exceptions/FhirServiceExceptionProcessor.cs
```diff
@@ -41,8 +41,8 @@ public static (Exception customException, string errorName) CustomizeException(E
 
             switch (exception)
             {
-                case FhirException _:
-                    var status = ((FhirException)exception).StatusCode;
+                case FhirClientException _:
```
dustinburson, line 44: Suggestion:  I know this is an edit to existing code but I think you can inline have a reference to typed variable in the case statement, something like case FhirClientException fhirEx: and this would allow you to avoid the explicit cast.

dustinburson (approved):

dustinburson (dismissed):

dustinburson (approved):

src/lib/Microsoft.Health.Extensions.Fhir.R4/Telemetry/Exceptions/FhirServiceExceptionProcessor.cs
namalu: I removed the explicit casts and used the typed variable instead.

PR #295, opened 2023-09-05
Centralized package versions

pallar-ms:
* Centralized package versions
* Fix stylecop analyzer errors

### DIFF
--- a/src/console/Microsoft.Health.Fhir.Ingest.Console.Common/Extensions/CredentialProviderExtensions.cs
+++ b/src/console/Microsoft.Health.Fhir.Ingest.Console.Common/Extensions/CredentialProviderExtensions.cs
@@ -1,8 +1,7 @@
-﻿// --------------------------------------------------------------------------
-// <copyright file="CredentialProviderExtensions.cs" company="Microsoft Corporation">
+﻿// -------------------------------------------------------------------------------------------------
 // Copyright (c) Microsoft Corporation. All rights reserved.
-// </copyright>
-// --------------------------------------------------------------------------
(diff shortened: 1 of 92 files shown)


src/console/Microsoft.Health.Fhir.Ingest.Console.Common/Extensions/CredentialProviderExtensions.cs
```diff
@@ -1,8 +1,7 @@
-﻿// --------------------------------------------------------------------------
-// <copyright file="CredentialProviderExtensions.cs" company="Microsoft Corporation">
+﻿// -------------------------------------------------------------------------------------------------
 // Copyright (c) Microsoft Corporation. All rights reserved.
-// </copyright>
-// --------------------------------------------------------------------------
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
```
dustinburson, line 3: Quick look it does appear we are using the MIT header in other places in the code, these is just to correct the few places where we weren't?

dustinburson (approved):

src/console/Microsoft.Health.Fhir.Ingest.Console.Common/Extensions/CredentialProviderExtensions.cs
pallar-ms: I got these compilation errors to fix licensing (likely because either the compiler toolset or stylecop analyzer package got applied to these projects that didn't previously have it) and it matches the other files in this repo now.

PR #296, opened 2023-09-05
Add 'GenerateDocumentationFile' project build property

pallar-ms:
* Add 'GenerateDocumentationFile' project build property
* Fix errors detected with the above build property.


src/lib/Microsoft.Health.Extensions.Fhir/Search/SearchExtensions.cs
```diff
@@ -26,7 +26,7 @@ public static string CompositeOr(this IEnumerable<string> searchTokens)
         /// </summary>
         /// <param name="searchTokens">Enumerable of search tokens to combine.</param>
         /// <returns>A concatenated string of search terms joined for an And search.</returns>
-        /// <seealso cref="https://www.hl7.org/fhir/search.html#combining"/>
+        /// <seealso cref="!:https://www.hl7.org/fhir/search.html#combining"/>
```
rogordon01, line 29: What does the `!:` do? Just curious

rogordon01 (approved):

src/lib/Microsoft.Health.Extensions.Fhir/Search/SearchExtensions.cs
pallar-ms: It was to syntactically represent urls in a cref (without that the compilation fails). But I realized that it doesn't actually hyperlink it, and only renders the text. So, I changed to href.

dustinburson (approved):

PR #297, opened 2023-10-06
Return an indicator when we resume the processor

rogordon01:
The `ResumeAsync` method will restart processing if the EventProcessorClient is not already running. This PR returns an indication of when the method has started the processor.

This indication will support downstream health checks.

### DIFF
--- a/src/lib/Microsoft.Health.Events/EventHubProcessor/IResumableEventProcessor.cs
+++ b/src/lib/Microsoft.Health.Events/EventHubProcessor/IResumableEventProcessor.cs
@@ -15,7 +15,7 @@ namespace Microsoft.Health.Events.EventHubProcessor
         /// Attempts to resume the Event Processor.
         /// </summary>
         /// <param name="ct">The cancellation token</param>
-        /// <returns><true> if the Event Processor was restart. <false> if it was already running.</returns>
+        /// <returns>true if the Event Processor was restart. false if it was already running.</returns>
         Task<bool> ResumeAsync(CancellationToken ct);
 
         Task RunAsync(CancellationToken ct);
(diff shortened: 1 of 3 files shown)


namalu (dismissed):

src/lib/Microsoft.Health.Events/EventHubProcessor/IResumableEventProcessor.cs
```diff
@@ -11,7 +11,12 @@ namespace Microsoft.Health.Events.EventHubProcessor
 {
     public interface IResumableEventProcessor : IDisposable
     {
-        Task ResumeAsync(CancellationToken ct);
+        /// <summary>
+        /// Attempts to resume the Event Processor.
+        /// </summary>
+        /// <param name="ct">The cancellation token</param>
+        /// <returns><true> if the Event Processor was restart. <false> if it was already running.</returns>
+        Task<bool> ResumeAsync(CancellationToken ct);
```
wi-y, line 19: I think [ResumableAssignedPartitionProcessor.cs](https://github.com/microsoft/iomt-fhir/blob/658266857c920d8c14abd76c0eb5f16f5097975d/src/lib/Microsoft.Health.Events/EventHubProcessor/ResumableAssignedPartitionProcessor.cs#L29) also inherits from IResumableEventProcessor - do you mind adding the new bool return type for ResumeAsync? you can return false or something. I probably should have used a different interface

rogordon01: Yea I just noticed that other class due to the build failure. Updated!

wi-y (approved):

PR #300, opened 2023-10-19
Update Azure.Identity to version 1.10.3 along with downstream packages.

dustinburson:
Update Azure.Identity package to address known vulnerability.


kyclai:
Some packages have a dependency on Azure.Identity < 1.10.3, such as the following. Do they need to be updated too?
- `Microsoft.Extensions.Azure`: "Azure.Identity": "1.4.0"
- `Microsoft.Health.Client`: "Azure.Identity": "1.10.0"

kyclai:
https://github.com/microsoft/iomt-fhir/blob/main/src/console/packages.lock.json references Azure.Identity 1.9.0. Does this need to be updated too?

rogordon01 (approved):

dustinburson:
> https://github.com/microsoft/iomt-fhir/blob/main/src/console/packages.lock.json references Azure.Identity 1.9.0. Does this need to be updated too?

Thanks for the call out Carrie.  The package.lock files are automatically generated.  The syntax [1.9.0, ) means the min version is 1.9 but can go higher unbounded.

dustinburson:
> Some packages have a dependency on Azure.Identity < 1.10.3, such as the following. Do they need to be updated too?
> 
> * `Microsoft.Extensions.Azure`: "Azure.Identity": "1.4.0"
> * `Microsoft.Health.Client`: "Azure.Identity": "1.10.0"

I am running our internal compliance build to verify but I don't think it is an issue.  Our package version should be forcing the use of 1.10.3.

kenziedolish (approved):

PR #302, opened 2024-12-05
Personal/masaadi/dotnet8 migartion iomt

maryamasaaditeh:


psapozh82 (dismissed):
To me it seems like you're just changing .NET compilation level to 8 and updating packages which makes sense to me. I hope you tested the actual functional behavior of this code and that there's no regression.

I would also suggest getting another round of review from some team members who know more about this before merging.

dustinburson:
> To me it seems like you're just changing .NET compilation level to 8 and updating packages which makes sense to me. I hope you tested the actual functional behavior of this code and that there's no regression.
> 
> I would also suggest getting another round of review from some team members who know more about this before merging.

Thanks @psapozh82.  Good point and this was done.  I had a similar ask on the prior PR that was abandoned.  We have been trying to trouble shoot the PR build trigger issues.

dustinburson (approved):

